PR #5, opened 2017-04-15
Support delayed functions

### LINKED ISSUE #1: How can I use with dask.delayed functions?
Could You provide an example to map functions returning `Delayed` objects?

kszucs:
Resolves #1

### DIFF
--- a/streams/tests/test_dask.py
+++ b/streams/tests/test_dask.py
@@ -3,6 +3,7 @@ from time import time
 
 import pytest
 
+from dask import delayed
 from distributed.client import Future
 from distributed.utils_test import inc, double, gen_test, gen_cluster
 from distributed.utils import tmpfile
@@ -51,6 +52,33 @@ def test_map(c, s, a, b):
 
     results = yield c._gather(L)
     assert results == [1, 2, 3]
+
+
+@gen_cluster(client=True)
+def test_compute(c, s, a, b):
+    source = Stream()
+
+    @delayed
+    def inc1(x):
+        """Immediate delayed function"""
+        return inc(x)
+
+    def inc2(x):
+        """Function that just returns a delayed"""
+        return delayed(inc)(x)
+
+    L1 = source.to_dask().compute(inc1).sink_to_list()
+    L2 = source.to_dask().compute(inc2).sink_to_list()
+
+    for i in range(3):
+        source.emit(i)
+
+    for L in [L1, L2]:
+        assert len(L) == 3
+        assert all(isinstance(x, Future) for x in L)
+
+        results = yield c._gather(L)
+        assert results == [1, 2, 3]
 
 
 @gen_cluster(client=True)
--- a/streams/dask.py
+++ b/streams/dask.py
@@ -1,3 +1,4 @@
+from dask.delayed import Delayed
 from distributed.client import default_client
 from tornado.locks import Condition
 from tornado.queues import Queue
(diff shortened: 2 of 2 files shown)


streams/tests/test_dask.py
```diff
@@ -54,6 +55,33 @@ def test_map(c, s, a, b):
...
+    def inc1(x):
+        """Immediate delayed function"""
+        return inc(x)
+
+    def inc2(x):
+        """Function that just returns a delayed"""
+        return delayed(inc)(x)
+
+    L1 = source.to_dask().compute(inc1).sink_to_list()
+    L2 = source.to_dask().compute(inc2).sink_to_list()
```
mrocklin: What is your objective here?  What are you trying to accomplish?  Is this preferable to `.map(inc)` in some way?

kszucs: I have an already implemented function which creates the following dag from an input record:

image

Theoretically I could create the same `streaming DAG` but just computing a delayed is easier to start with.

kszucs: Do You mean placing the same logic under `map`? That was my first try, but I thought this is more explicit. `Client.submit` doesn't handle delayed objects, I need to use `Client.compute` instead.

mrocklin: Just thinking aloud, but thoughts on a function that could take delayed values as inputs and produce the equivalent streaming DAG?

    stream_inputs, stream_outputs = make_stream(delayed_inputs, delayed_outputs)

kszucs: Wouldn't it hurt the distributed scheduler? Seems reasonable and complex to implement :)

mrocklin: Hurt it how?

I agree that this is more complex and that we shouldn't do it if not useful.  I'm thinking about cases where your inputs and outputs come in and leave at different rates.  It looks like your graph has four inputs.  How do you apply this graph onto one element coming through the stream?  It is a tuple of four things maybe?

Also, just to be clear I'm not negative on your PR, just thinking about other alternatives in this space.

kszucs: Hurt the Performance of the scheduler, never mind!

kszucs:
The input is a dictionary containing 4 keys in this particular case.

```python
def pipeline(repos, account_id, project_id, connection):
    # clone repos and preprocess extracted dataframes
    frames = valmap(clone, repos)  # clone is a delayed function
    frames = valmap(preprocess, frames)  # preprocess is a delayed function

    # extract commit and patchframes
    cdfs = valmap(get(0), frames)  # toolz.curried.get
    pdfs = valmap(get(1), frames)

    # create authorframe with deduplication rules
    cdf = delayed(pd.concat)(cdfs, names=['repo_id'])
    cdf, adf = deduplicate(cdf)  # delayed 

    # hour and effort estimation
    cdf = estimate_hours(cdf)  # delayed

    ncommits = persist_commits(cdf, account_id, project_id, connection)
    nauthors = persist_authors(adf, account_id, project_id, connection)
    npatches = [persist_patches(pdf, account_id, project_id, repo_id,
                                connection)
                for repo_id, pdf in pdfs.items()]
    npatches = delayed(sum)(npatches)

    return ncommits, npatches, nauthors
```

CJ-Wright:
@kszucs Does the current setup with https://github.com/mrocklin/streamz/blob/master/streamz/dask.py support what you are asking for? If so can you please close?

PR #8, opened 2017-04-20
Stream wrapper

ordirules:
Here's a suggestion for multiple arguments to streams: wrappers. Wrappers are functions that process the inputs before they're `map`ed or `accumulate`d. I think this is also analogous to network packets, where at each layer in the stack, you have the processing function unwrap the header of the data and pass through the data.

I include the ability to use wrappers in a way that should be backwards compatible with the streams idea (passes all tests included).

Wrappers are typically used with data that has a recongizable signature type. An example wrapper is included in here which included two main pieces:
1. StreamDoc.parse_streamdoc : the decorator to intercept the function
2. StreamDoc.StreamDoc : a StreamDoc object

The combination of the two serve as:
- An argument multiplexer :  to go around circumvent the one(two)
  argument(s) requirement of the mapping(accumulating) functions.
- An attibute container : to allow the passing of attributes to
  functions
- Statistics bookkeeping : to allow the passing of attributes that are
  specific to the instance of the processed stream but not specific to
  the data contained in the streams. For example, run start and run stop
  times would be useful to include here.

This is sort of what we use for our pipeline, except we don't use a streaming fashion (rather, we're looping over delayed objects). I like the streaming API in that it neatly abstracts the data from the pipeline in a better way.

I'd be interested to hear your thoughts, if this may be a potential extension to the library. I envisage if yes, it would require significant discussion/work as it's quite a big PR.

Thanks for the support!

### DIFF
--- a/streams/StreamDoc.py
+++ b/streams/StreamDoc.py
@@ -1,0 +1,249 @@
+'''
+    This code is where most of the StreamDoc processing in the application
+    layer should reside. Conversions from other interfaces to StreamDoc are
+    found in corresponding interface folders.
+'''
+from functools import wraps
+
+_MAX_STR_LEN = 72  # Maximum length for string representations of objects
+
+
+class StreamDoc(dict):
+    def __init__(self, streamdoc=None, args=(), kwargs={}, attributes={}):
+        ''' A generalized document meant to be parsed by Streams.
+
+            Components:
+                attributes : the metadata
+                outputs : a dictionary of outputs from stream
+                args : a list of args
+                kwargs : a list of kwargs
+                statistics : some statistics of the stream that generated this
+                    It can be anything, like run_start, run_stop etc
+        '''
+        # TODO : make attributes a separate class with its own get/set and
+        # parsing
+        # TODO : make option whether not attributes should be inherited?
+        # (global attributes). Such an implementation requires a
+        # 'global_attributes' name list
+        # TODO : functions to select, remove replace etc
+        # TODO : join method? join(override=True) for ex where override says
+        # what to do with conflicting attributes
+
+        # initialize
+        self['attributes'] = dict()
+        self['kwargs'] = dict()
+        self['args'] = list()
+        self['statistics'] = dict()
+        # needed to distinguish that it is a StreamDoc by stream methods
+        self['_StreamDoc'] = 'StreamDoc v1.0'
+
+        # update
+        if streamdoc is not None:
+            self.updatedoc(streamdoc)
+
+        # override with args
+        self.add(args=args, kwargs=kwargs, attributes=attributes)
+
+    def updatedoc(self, streamdoc):
+        self.add(args=streamdoc['args'], kwargs=streamdoc['kwargs'],
+                 attributes=streamdoc['attributes'],
+                 statistics=streamdoc['statistics'])
+
+    def add(self, args=[], kwargs={}, attributes={}, statistics={}):
+        ''' add args and kwargs'''
+        if not isinstance(args, list) and not isinstance(args, tuple):
+            args = (args, )
+
+        self['args'].extend(args)
+        # Note : will overwrite previous kwarg data without checking
+        self['kwargs'].update(kwargs)
+        self['attributes'].update(attributes)
+        self['statistics'].update(statistics)
+
+    @property
+    def args(self):
+        return self['args']
+
+    @property
+    def kwargs(self):
+        return self['kwargs']
+
+    @property
+    def attributes(self):
+        return self['attributes']
+
+    @property
+    def statistics(self):
+        return self['statistics']
+
+    def get_return(self, elem=None):
+        ''' get what the function would have normally returned.
+
+            returns raw data
+            Parameters
+            ----------
+            elem : optional
+                if an integer: get that nth argumen
+                if a string : get that kwarg
+        '''
+        if isinstance(elem, int):
+            res = self['args'][elem]
+        elif isinstance(elem, str):
+            res = self['kwargs'][elem]
+        elif elem is None:
+            # return general expected function output
+            if len(self['args']) == 0 and len(self['kwargs']) > 0:
+                res = dict(self['kwargs'])
+            elif len(self['args']) > 0 and len(self['kwargs']) == 0:
+                res = self['args']
+                if len(res) == 1:
+                    # a function with one arg normally returns this way
+                    res = res[0]
+            else:
+                # if it's more complex, then it wasn't a function output
+                res = self
+
+        return res
+
+    # TODO : allow partial remapping both for args and kwargs
+    def select(self, mapping):
+        ''' remap args and kwargs
+            combinations can be any one of the following:
+
+
+            Some examples:
+
+            (1,)        : map 1st arg to next available arg
+            (1, None)   : map 1st arg to next available arg
+            'a',        : map 'a' to 'a'
+            'a', None   : map 'a' to next available arg
+            'a','b'     : map 'a' to 'b'
+            1, 'a'      : map 1st arg to 'a'
+
+            The following is NOT accepted:
+            (1,2) : this would map arg 1 to arg 2. Use proper ordering instead
+            ('a',1) : this would map 'a' to arg 1. Use proper ordering instead
+
+            Notes
+            -----
+            These *must* be tuples, and the list a list kwarg elems must be
+                strs and arg elems must be ints to accomplish this instead
+        '''
+        if not isinstance(mapping, list):
+            mapping = [mapping]
+        streamdoc = StreamDoc(self)
+        newargs = list()
+        newkwargs = dict()
+        totargs = dict(args=newargs, kwargs=newkwargs)
+
+        for mapelem in mapping:
+            # duck-type to figure out if mapelem is a str key or int index
+            if isinstance(mapelem, str):
+                mapelem = mapelem, mapelem
+            elif isinstance(mapelem, int):
+                mapelem = mapelem, None
+
+            # only for strings
+            if len(mapelem) == 1 and isinstance(mapelem[0], str):
+                mapelem = mapelem[0], mapelem[0]
+            elif len(mapelem) == 1 and isinstance(mapelem[0], int):
+                mapelem = mapelem[0], None
+
+            oldkey = mapelem[0]
+            newkey = mapelem[1]
+
+            if isinstance(oldkey, int):
+                oldparentkey = 'args'
+            elif isinstance(oldkey, str):
+                oldparentkey = 'kwargs'
+
+            if newkey is None:
+                newparentkey = 'args'
+            elif isinstance(newkey, str):
+                newparentkey = 'kwargs'
+            elif isinstance(newkey, int):
+                raise ValueError("Integer tuple pairs not accepted")
+
+            if newparentkey == 'args':
+                totargs[newparentkey].append(streamdoc[oldparentkey][oldkey])
+            else:
+                totargs[newparentkey][newkey] = streamdoc[oldparentkey][oldkey]
+
+        streamdoc['args'] = totargs['args']
+        streamdoc['kwargs'] = totargs['kwargs']
+
+        return streamdoc
+
+
+def _is_streamdoc(doc):
+    if isinstance(doc, dict) and '_StreamDoc' in doc:
+        return True
+    else:
+        return False
+
+
+def parse_streamdoc(name):
+    ''' Decorator to parse StreamDocs from functions
+
+    This is a decorator meant to wrap functions that process streams.
+        It must make the following two assumptions:
+            functions on streams process either one or two arguments:
+                - if processing two arguments, it is assumed that the operation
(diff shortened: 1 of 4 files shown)


ordirules:
Some context: We're trying to solve the issue of bookkeeping calculations on xray images acquired at an x-ray beamline. 

Typical pipelines involve grabbing images, looking at slices, averaging regions, running correlation functions on images, running correlation functions on time series.
Along with processing we also need to:
 - Save the results either in a database, raw file, xml file (sinks). 
 - Keep track of metadata. Along with this data is tons of metadata passed, such as sample name, measurement type, some motor positions etc. These metadata are not necessarily essential to computing results but essential for future review.

Since acquisition times are rather slow, we're not so much interested in high throughputs as we are in acquiring a consistent set of results, which are computed using methods easily understood by busy scientists. This explains why we're willing to sacrifice throughput with the overhead introduced in this PR. The chunks of data being passed in these functions are also usually much larger than the overhead, so in our case there may not even be much overhead.

The bookkeeping should help for the distributed regime as well as reviewing results.

We have methods in python, but are hoping to develop something easier to understand and scalable for an incoming array of processing techniques we have planned to incorporate (in collaboration with the computing dept). The idea of streams that keep tracks of their parents and children would be very beneficial to us.

I hope this helps better explain the reasoning behind this PR. Thanks!

CJ-Wright:
A small nit pick, I would argue that saving results in a database should be a pervasive operation not just a sink.

ordirules:
I agree on that one and it's a very good point. I think I'll wait for @mrocklin before discussing further, I feel this post has been long enough. Thanks for bringing that up @CJ-Wright. Also, whether or not this is compatible with streams, I think we could gain a lot from each others' insight @CJ-Wright and would like to keep discussing in the long run. (I like that idea of `tee`ing to make an infinite stream, for instance)

mrocklin:
OK, some quick feedback.  

As I see it this PR proposes the following:

1.  A class to contain data alongside arbitrary metadata as well as an adapter such that this metadata passes through existing streams operations
2.  A general wrapping system that enables the first point, but might also enable other similar systems 
3.  A policy that when two streamdoc objects meet that their metadata should be merged

My initial gut response after reading is as follows:

-  I agree with the value of passing around arbitrary metadata
-  I found that bringing up the questions of how to handle multi-input streams an interesting one.  Here it looks like you hard-coded in what should happen for map and scan, but presumably this gets more complex when we try to handle things generally (like zip, combine_latest, sliding_window, etc.) and when metadata becomes more interesting (like event time, etc. where the merging approach to metadata may no longer be correct).  I think that this will be a challenging problem to get right in a general way.  I expect that we will probably have to think about this a bit before coming to a good solution.
-  If possible it might be useful to pass around dicts rather than custom StreamDoc objects, though this desire for using common data structures probably conflicts a bit with the point above about how to handle different kinds of update rules.
-  The wrapping system is interesting.  To flesh out this system I think that we would want to try it out with a few possible applications.  The code in dask.py might be an interesting case study.  For example is there a way to achieve the `.to_dask()` method without creating another `DaskStream` class by using wrappers?  What other possible applications are there?  If we can't find any then we may want to avoid the general solution.

Thoughts?  Feedback?

ordirules:
Thanks for the thorough reply. I'll try to answer quickly as well. I may not have time to be more thorough until tomorrow evening.
The merging policy is not discussed but I agree it needs also to be added if this is pursued.

The way I see this PR currently, is it's a patchy method to get streams to work in a way that I see is convenient for our data (which we use something similar to this). I'm hoping that thinking out a solution may allow us to find generalizations that we can all benefit from.

I agree with the problems of multi-input streams. I agree that it's complex enough of an issue that it will likely take a lot of planning/trial and error/testing to adopt a method if possible. This is one reason for my motivation of proposing a wrapper option as opposed to proposing a new object defined in streams. My intention here is that `StreamDoc` is just an example of how one would use wrappers (and so implementations are up to the user). But I am not arguing against ignoring an attempt to generalize it overall; I would like to see a nice solution emerge one day and hope perhaps discussion here may help towards this goal.
It would be possible to rewrite `StreamDoc` as a `dict` and move the member functions outside, and probably is a better approach.

I think the suggestion to get this to potentially replace `DaskStream` by this is a good step towards seeing if the `wrapper` option could be of general use. I think  the `test_delayed` test is close to that goal.
When writing it, for instance (and `test_apply` as well, which is also closer to what we want to do), I realized that we need two functions, `map` and `apply`, where `map` works on the unwrapped data whereas `apply` works on the wrapped data. I think this might suggest that we may have to think of two classes of operators, one that acts on wrapped and unwrapped data. In my opinion, adding a new method to supply an option begins to complicate matters and should definitely be given more careful thought. However, wrapping seems like such a common operation in streaming data that I'd be tempted to keep persisting on pushing it, to see if it could be useful.

I can give `DaskStream` a test tomorrow evening and see if this uncovers any other subtle details and see if I can think of any more ideas.

Thanks for reading this (heavy) PR and insight. I think this is a very promising library. (and hope to see more discussion about it)

ordirules:
Okay, I read your code in more detail. I have to be honest I still have yet to fully understand some of it, mainly because I have not used it (but of course I get the idea). So if I am wrong in anything please correct me.

Here are some more organized comments:
1. Multiple intputs/outputs : I did not notice that `zip` handled those. I think it handles them nicely. There is just the issue of `kwargs`, but perhaps that should be deferred for now? I think if `zip` is used then what is missing is a multiplexer (splitting a returned result into multiple streams). Something like (in `Stream` class):
```python
    def multiplex(self, nelems):
        ''' Multiplex stream into a series of nelems streams.

            Assumes that the emitted value of parent is a tuple
                of at least length nelems.
        '''
        def get_elem(args, elem=None):
            if elem is None:
                raise ValueError("Must supply an elem number")
            return args[elem]

        return [self.map(get_elem, elem=i) for i in range(nelems)]
```
@CJ-Wright  may have more experience with this as well.

This allows things like:
```python
def times2(*args):
    newargs = list()
    for arg in newargs:
        newargs.append(args*2)
    return newargs

s1 = Stream()
s2 = Stream()
s3 = s1.zip(s2)
s4 = s3.apply(myfunc)
sa, sb = s4.multiplex(2)
sa.sink(print)
sb.sink(print)
s1.emit(1)
s2.emit(2)
```
Also, I would suggest maybe `def zip(self, other):` should be changed to `def zip(self, *other):` to allow for zipping in multiple arguments (I know it can be done with `streams` but doing it from an instance of `Stream` itself could also be useful, maybe a PR? You may have had a specific reason for not doing this though).
In all, I think for now, it might be worth pushing the limits of `zip` in most applications to see if any other method is needed.

2. I find the wrapping system very useful for delayed objects. I have been testing it and it seems to work fine. Here's a simplified example that doesn't book-keep the attributes:
```python
def test_delayed():
    ''' try making a delayed wrapper.'''
    from dask import delayed
    def myadder(argtuple):
        a,b = argtuple
        return a + b

    from dask import delayed, compute
    from distributed import client
    client = Client()


    # future streams inherit wrapper, wrapper is a function that takes a name as parameter
    # and returns decorator. Name is a way to allow giving the wrapper some identifier, may
    # not be the best implementation
    s1 = Stream(wrapper=lambda name : delayed(pure=True))
    s2 = Stream(wrapper=lambda name : delayed(pure=True))

    # map maps the unwrapped function, apply maps the wrapped function
    s3 = s1.zip(s2)
    s4 = s3.map(myadder)
    s5 = s4.map(print_args)
    s5.apply(compute)
    # or for distributed client
    s5.apply(client.submit)

    s1.emit(1)
    s2.emit(2)
```
I think if wrapping is used, there is no choice but to allow some `apply` function. Basically, something that `map`s to a function that has been wrapped. Either that, or let the `sink` received the wrapped data. What do you think?

Overall, I like the syntax of this library. It flows much nicer (to me) than by simply using dask. On top of that, it nicely abstracts away the computations from the data. This really helps when the complexity increases; streams can be saved in bits and pieces and connected gradually. I would like to start implementing it if you don't mind, and see how far it goes. If it is not supported, I'm sure we can use some bits and pieces (and of course, you will be credited in any piece of code we use).

So I think I have three main thoughts:
1. Add a `multiplex` option
2. Change `zip(self, other)` to `zip(self, *other)` (minor, and not necessary)
3. Add a `wrap` option. It seems to help take care of `delayed` and allow passing of attributes. However, supporting attributes, at this stage (when very little is known of the requirements/details) is probably not a good idea.

What do you think? I apologize this is long. Please, feel free to ignore and focus on the bits and pieces you think are relevant. If I find something is worth mentioning again, I'll write it again.

thanks!

mrocklin:
Things like multiplex make sense.  It might be worth looking at grouped streams in other streaming systems (like flink) to see if there is a way to unify APIs.  Terms like `toolz.pluck` and `unzip` might also inform that API.

Supporting `zip(self, *other)` makes sense to me.

I'm still not confident that we have enough information to implement a clean wrapping system, but I think that implementing something is probably a good way of learning more.  As long as you're ok with a high probability of having to rewrite things I say go for it.

> Overall, I like the syntax of this library. It flows much nicer (to me) than by simply using dask. On top of that, it nicely abstracts away the computations from the data. This really helps when the complexity increases; streams can be saved in bits and pieces and connected gradually. I would like to start implementing it if you don't mind, and see how far it goes. 

I'm glad you find it useful.  I agree that using this in some application will probably reveal issues and help to identify opportunities for improvement.

> If it is not supported, I'm sure we can use some bits and pieces (and of course, you will be credited in any piece of code we use).

I've just pushed a license to the repository (thank you for the reminder).  It's BSD-3.  I don't need any credit other than what is listed there.  Feel free to do as you like.

ordirules:
Thanks, a little busy but will take a look later. For the record, I'm starting here:
https://ci.apache.org/projects/flink/flink-docs-release-1.3/dev/datastream_api.html

Our current implementation misses many things that streams are useful for, such as more complex control flow like windowing, grouping, etc. I'd say the only more complex control flow we currently can use is `zip`, which waits until all inputs are filled before piping out the results (and `accumulate`). Currently, we use it to better define a typical pipeline, which is not as simple as iterating over a list of commands, but rather a DAG of commands. The same implementation could be carried out in dask, but it looks much cleaner using streams. This leads to much better code. One more complex example other than branching I can think of is `zip`. A user would be much less hesitant to add such control flow in their pipeline using streams as opposed to a brute force loop approach.

So I would like to spend some time when I can to explore the more complex features of streams, and see if we can take advantage of that. I'll keep you updated. Thanks for all the support.

mrocklin:
Just to be clear, the streams library does have a few other things than zip.  Operations like combine_latest, timed_window, unique, etc. handle some (very incomplete) fraction of what things like flink does.  Obviously this isn't close to a full implementation, I just wanted to make sure you were seeing things other than zip.

ordirules:
yep, I had definitely seen and looked over them. By 'our current implementation' I meant our pipeline. I was trying to emphasize just that, that we're only utilizing a subset of the features that you have in `streams` (and that is also in flink). But I feel we can gain from and would like to (and so hence we can then have a more fair evaluation on whether `wrapper` is useful as a core parameter or not). I just need a little bit of time to sort of absorb the rest. thanks!

mrocklin:
> By 'our current implementation' I meant our pipeline

Ah, my mistake.  That makes sense then.

> I just need a little bit of time to sort of absorb the rest. thanks!

Cool.  I appreciate you looking into this.  Working with an application in hand always produces nicer results.

ordirules:
No problem and thanks for the support. I have been busy with other things. I may go silent for maybe a week (but I'll keep reading any updates in this library).

ordirules:
This seems interesting here:
https://ci.apache.org/projects/flink/flink-docs-release-1.2/dev/windows.html

In our code, the type of windowing we would make the most use of (but not exclusive use) are tumbling windows (see flink link above), which are not quite sliding windows and not quite timed windows implemented (correct me if I'm wrong). Seems to me that a nice constructive next step is to implement a tumbling window. I'm not sure how quick I'll be on that (maybe I'll start in the next couple days), so it could be posted as a general issue for anyone else interested? Just thoughts, using this as a log, trying to keep it alive.

mrocklin:
How do tumbling windows differ from the current `Stream.timed_window`?

ordirules:
my mistake, sorry, please ignore that comment :-(. Yes, they're the same. I was thinking of a `partition`...
I think it would be useful to have a hybrid of `partition` with a maximum collection time. Basically, aggregate elements together for either a certain maximum number of elements, or a certain maximum time.

Here is the practical reason. The `partition` is useful in that we can aggregate a series of 2D images to send to some machine learning algorithm (GPU computation) for inference all at once (there is a lot of overhead so sending one image at a time is a waste of resources). However, if we're sending data at a very slow rate (or say, we stopped at one last image and lead to an incomplete buffer and won't run again for a few days), we want the computations to run anyway.

The algorithm won't care how many 2D images we aggregate together so ensuring a certain length is not a big deal. Just thoughts. If this is already possible or is not but worth adding, let me know.

mrocklin:
Yes, I think I've seen similar kinds of windowing policies in other projects as well.

ordirules:
discussion continued to #11

PR #9, opened 2017-04-27
allow bound method zip to merge with more than one stream

ordirules:
from discussion in #8 . Allow multiple elements to be zipped.
for the record, the alternative is also:
```python
import streams as s
d = s.zip(a,b,c)
```

what is now allowed with this PR is also this:
```python
a = Stream()
b = Stream()
c = Stream()
# previous method
d = a.zip(b)
d = d.zip(c)

a = Stream()
b = Stream()
c = Stream()
# now this is allowed
d = a.zip(a,b,c)
```

### DIFF
--- a/streams/tests/test_core.py
+++ b/streams/tests/test_core.py
@@ -251,7 +251,6 @@ def test_zip():
     a.emit(1)
     b.emit(2)
     d.emit(3)
-    print(L2)
     assert L2 == [(1, 2, 3)]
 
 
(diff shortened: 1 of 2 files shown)


streams/tests/test_core.py
```diff
@@ -242,6 +242,17 @@ def test_zip():
...
+    d = Stream()
+    # test zip from the object itself
+    # zip 3 streams together
+    e = a.zip(b, d)
+    L2 = e.sink_to_list()
+
+    a.emit(1)
+    b.emit(2)
+    d.emit(3)
+    print(L2)
```
ordirules: oops! done. thanks!

PR #11, opened 2017-05-09
Add __stream_map__, __stream_reduce__, and __stream_merge__ protocols

mrocklin:
This allows objects within a stream to take control of map, accumulate, and
combination operations (like zip, combine_latest).  I *hope* that this helps to
resolve problems around dask, metadata tracking, batching, etc..

cc @ordirules @danielballan

### DIFF
--- a/streams/core.py
+++ b/streams/core.py
@@ -1,11 +1,10 @@
 from collections import deque
-import math
 from time import time
 
 import toolz
 from tornado import gen
 from tornado.locks import Condition
-from tornado.ioloop import IOLoop, PeriodicCallback
+from tornado.ioloop import IOLoop
 from tornado.queues import Queue
 
 
--- a/streams/core.py
+++ b/streams/core.py
@@ -203,6 +203,14 @@ class Stream(object):
         """ Add a time delay to results """
         return delay(interval, self, loop=None)
 
+    def combine_latest(self, *others):
+        """ Combine multiple streams together to a stream of tuples
+
+        This will emit a new tuple of all of the most recent elements seen from
+        any stream.
+        """
+        return combine_latest(self, *others)
+
     def concat(self):
         """ Flatten streams of lists or iterables into a stream of elements
 
--- a/streams/core.py
+++ b/streams/core.py
@@ -224,6 +224,20 @@ class Stream(object):
 
     flatten = concat
 
+    def union(self, *others):
+        """ Combine multiple streams into one
+
+        Every element from any of the children streams will immediately flow
+        into the output stream.  They will not be combined with elements from
+        other streams.
+
+        See also
+        --------
+        Stream.zip
+        Stream.combine_latest
+        """
+        return union(children=(self,) + others)
+
     def unique(self, history=None, key=identity):
         """ Avoid sending through repeated elements
 
@@ -540,3 +554,8 @@ class unique(Stream):
         if y not in self.seen:
             self.seen[y] = 1
             return self.emit(x)
+
+
+class union(Stream):
+    def update(self, x, who=None):
+        return self.emit(x)
--- a/streams/core.py
+++ b/streams/core.py
@@ -316,14 +316,15 @@ class Sink(Stream):
 
 
 class map(Stream):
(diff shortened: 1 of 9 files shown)


streams/core.py
```diff
@@ -474,6 +483,8 @@ def update(self, x, who=None):
         if len(L) == 1 and all(self.buffers):
             tup = tuple(buf.popleft() for buf in self.buffers)
             self.condition.notify_all()
+            if tup and hasattr(tup[0], '__stream_merge__'):
+                tup = tup[0].__stream_merge__(*tup[1:])
```
mrocklin: In the future this would need some sort of dispatching mechanism to combine different kinds of objects.

ordirules:
I just found it will take a bit of work to get to function with `Delayed`. The reason is that I now need to subclass `Delayed` (adding the `__stream_map__` etc methods). My previous wrapper method allowed me to avoid this by just supplying the `delayed` routine as a wrapper.

Basically, as the data comes in, I need to transform it with `delayed(wrapper(func))(data)` where `data` is encoded/decoded by `wrapper`. Adding this as a wrapper to streams was enough previously to get this working, and running on cluster.

Any tips on getting that implemented? Would it be simple enough to subclass `Delayed` for this? I currently can't see a simple way, but I'm not too familiar with the library. Thanks!

mrocklin:
I'm somewhat curious why you're passing delayed objects through the stream.  

We might consider using something like `functools.singledispatch` instead of a `__stream_map__` attribute.

mrocklin:
I started making a batched solution with the protocol and found that I also needed to change some things around.  I suspect that we'll need to iterate on this a few times before it's stable.

ordirules:
I'm passing delayed objects through the stream to construct a task graph of delayed computations. I can then choose to submit pieces of the stream to the cluster if needed and save the futures to deques, to cache various intermediate steps.
If there is a cleaner solution, I would be interested to hear. Currently this has worked in the past.

I'm not sure singledispatch would work. For example, what if these wrappers were nested? How would you distinguish them wrapper1(wrapper2(f)) from wrapper1(wrapper3(f))? I'm thinking of the scenario where a user may have their own wrapper that they `delay` (which again may not be useful). However, this may not be a common usage case.

mrocklin:
I've also been running into the nesting problem.  (futures of batches).

mrocklin:
I plan to work on this a bit today.  My plan is the following:

1.  Implement a Batch type that maps over many elements.  This will require me to change some of the protocols like `__stream_reduce__` and `__stream_merge__`
2.  Try to handle Futures of Batches.  I think that this will force me to think about nesting.
3.  Remove the monkey-patching of Futures, which I think will require the use of a dispatching solution

I think that this will cover the cases that you bring up.  Regardless it would be useful to have your input if I'm able to put something together today.

ordirules:
That would be great. I have been a little busy on my end, but I will definitely try to read and offer input and test things out.

The nesting piece will be very tricky I think. I think it seems like there are two choices: either you give the stream information on how to handle the data, or else you force the data to know about it. You mentioned you don't want to do the former to keep the API clean, if I understand. I agree. The fact this library is lightweight is a nice plus.

I think for my specific case, subclassing a Delayed object to evaluate certain `hasattr` checks could work, so that:
```python
hasattr(x, '__stream_map__')
```
would evaluate true for a delayed(StreamDoc(something)), and so the object would know to do the mapping. It might be hackish though. But I think delayed is a special type of class and so it probably needs to be handled differently. I haven't tried this since I don't know so much about delayed to see if it's possible. If you had input, I'd be happy to hear.

Also, forcing user to write wrappers that are aware that they're nesting other wrappers is probably not a bad constraint (so the object with the `__stream_wrapper__` method must check for the next returned doc's `__stream_wrapper__`). If you're nesting, you should be aware of what you're doing.

Please correct me if I misunderstood anything. I hope this gives enough user input. By the way, I have reverted to using the wrapper before temporarily (since I can't delay objects so easily). It was pretty easy to switch to the new method, so I'm confident that I can update our library to any changes to test them. This allows me to test more features while we wring out these details. Thanks again for all the support!

PR #12, opened 2017-05-09
ENH: Add collect.

danielballan:
A nice method to have. See docstring for usage example.

### DIFF
--- a/streams/core.py
+++ b/streams/core.py
@@ -72,7 +72,7 @@ class Stream(object):
                 result.extend(r)
             else:
                 result.append(r)
-        return [x for x in result if x is not None]
+        return [element for element in result if element is not None]
 
     @property
     def child(self):
--- a/streams/core.py
+++ b/streams/core.py
@@ -72,7 +72,7 @@ class Stream(object):
                 result.extend(r)
             else:
                 result.append(r)
-        return [r for r in result if r is not None]
+        return [x for x in result if x is not None]
 
     @property
     def child(self):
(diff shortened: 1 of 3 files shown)


streams/core.py
```diff
@@ -519,3 +538,18 @@ def update(self, x, who=None):
...
+		if cache is None:
+			self.cache = deque()
+        self.cache = cache
+
+        Stream.__init__(self, child)
+
+    def update(self, x, who=None):
+        self.cache.append(x)
+
+    def flush(self, _):
```
mrocklin: We should probably make `_` a keyword argument to support `s.flush()`

streams/core.py
```diff
@@ -519,3 +538,18 @@ def update(self, x, who=None):
...
+        self.emit(self.cache)
```
mrocklin: Do we want to clear out the cache?  Perhaps the following?

```python
out = tuple(self.cache())
self.cache.clear()
```

streams/core.py
```diff
@@ -241,6 +241,25 @@ def unique(self, history=None):
...
+		>>> collector = collect(source1)
+		>>> L = collector.sink(print)
+		>>> source2.sink(collector.flush)
+		>>> source1.emit(1)
+		>>> source1.emit(2)
+		>>> source2.emit('anything')  # flushes collector
+		...
+		[1, 2]
+		"""
+		return collect(self, cache=cache)
```
mrocklin: There appear to be some indentation issues here.

danielballan:
Rebased, added tests, force-pushed.

mrocklin:
I'll fix up flake8 issues (these were my fault) then rebase and merge.

mrocklin:
Thanks @danielballan !

PR #13, opened 2017-05-12
WIP - Batches, dispatch, and nesting

mrocklin:
OK, this is a work-in-progress PR that shows some ability to nest different kinds of elements in a stream.  This PR does a few things to enable this:

1.  We make a `Batch` type that is a tuple of many elements as a single batch in a stream
2.  We create a singledispatch system to replace the monkeypatching that we used to do on Futures
3.  We do a bit of magic so that the functions we pass down to elements again checks whatever it runs on and possible nests further.

As a result, we can do the following example, which maps a function over remote batches:

```python
In [1]: from dask.distributed import Client

In [2]: client = Client()

In [3]: from streams import Stream, Batch

In [4]: import streams.dask

In [5]: source = Stream()

In [6]: batches = source.partition(3).map(Batch)

In [7]: futures = streams.dask.scatter(batches)

In [8]: L = futures.map(lambda x: x + 1).sink_to_list()

In [9]: for i in range(9):
   ...:     source.emit(i)
   ...:     

In [10]: L
Out[10]: 
[<Future: status: finished, type: Batch, key: _stream_map-06488d54c44f3a3db136e6c34468eb13>,
 <Future: status: finished, type: Batch, key: _stream_map-cad3989e05761099bf5b22db5992eed3>,
 <Future: status: finished, type: Batch, key: _stream_map-ecd902c4b2cd3a96cb3c73694d64cfe6>]

In [11]: client.gather(L)
Out[11]: [(1, 2, 3), (4, 5, 6), (7, 8, 9)]
```

Note that things like accumulate (which will have to be changed) and merge (which will have to be split) don't work yet.  I wanted to get feedback from @ordirules and @danielballan .

### DIFF
--- a/streams/core.py
+++ b/streams/core.py
@@ -395,8 +395,8 @@ class scan(Stream):
         Stream.__init__(self, child)
 
     def update(self, x, who=None):
-        if hasattr(x, '__stream_reduce__'):
-            self.state, result = x.__stream_reduce__(self.func, self.state)
+        if hasattr(x, '__stream_accumulate__'):
+            self.state, result = x.__stream_accumulate__(self.func, self.state)
         else:
             self.state, result = stream_accumulate(x, self.func, self.state)
 
--- a/streams/core.py
+++ b/streams/core.py
@@ -1,4 +1,5 @@
 from collections import deque
+from functools import singledispatch
 from time import time
 
 import toolz
@@ -371,7 +372,7 @@ class map(Stream):
         if not self.raw and hasattr(x, '__stream_map__'):
             result = x.__stream_map__(self.func, **self.kwargs)
         else:
-            result = self.func(x, **self.kwargs)
+            result = stream_map(x, self.func, **self.kwargs)
 
         return self.emit(result)
 
@@ -396,17 +397,13 @@ class scan(Stream):
     def update(self, x, who=None):
         if hasattr(x, '__stream_reduce__'):
             self.state, result = x.__stream_reduce__(self.func, self.state)
-            if result is not no_default:
-                return self.emit(result)
-            else:
-                return []
-
-        if self.state is no_default:
-            self.state = x
         else:
-            result = self.func(self.state, x)
-            self.state = result
-            return self.emit(self.state)
+            self.state, result = stream_accumulate(x, self.func, self.state)
+
+        if result is not no_default:
+            return self.emit(result)
+        else:
+            return []
 
 
 class partition(Stream):
@@ -606,3 +603,17 @@ class collect(Stream):
         out = tuple(self.cache)
         self.emit(out)
         self.cache.clear()
+
+
+@singledispatch
+def stream_map(obj, func, **kwargs):
+    return func(obj, **kwargs)
+
+
+@singledispatch
+def stream_accumulate(obj, func, accumulator):
+    if accumulator is no_default:
+        return obj, no_default
+    else:
+        result = func(accumulator, obj)
+        return result, result
--- a/streams/core.py
+++ b/streams/core.py
@@ -394,13 +394,17 @@ class scan(Stream):
         Stream.__init__(self, child)
 
     def update(self, x, who=None):
+        if hasattr(x, '__stream_reduce__'):
+            self.state, result = x.__stream_reduce__(self.func, self.state)
+            if result is not no_default:
+                return self.emit(result)
+            else:
+                return []
+
         if self.state is no_default:
             self.state = x
         else:
-            if hasattr(x, '__stream_reduce__'):
-                result = x.__stream_reduce__(self.func, self.state)
-            else:
-                result = self.func(self.state, x)
+            result = self.func(self.state, x)
             self.state = result
             return self.emit(self.state)
 
--- a/streams/core.py
+++ b/streams/core.py
@@ -1,5 +1,5 @@
 from collections import deque
-from functools import singledispatch
+from functools import singledispatch, partial
 from time import time
 
 import toolz
@@ -369,12 +369,10 @@ class map(Stream):
         Stream.__init__(self, child)
 
     def update(self, x, who=None):
-        if not self.raw and hasattr(x, '__stream_map__'):
-            result = x.__stream_map__(self.func, **self.kwargs)
-        else:
-            result = stream_map(x, self.func, **self.kwargs)
+        if self.raw:
+            return self.emit(self.func(x, **self.kwargs))
 
-        return self.emit(result)
+        return self.emit(_stream_map(x, self.func, **self.kwargs))
 
 
 class filter(Stream):
@@ -605,9 +603,24 @@ class collect(Stream):
         self.cache.clear()
 
 
+def _stream_map(obj, func=None, **kwargs):
+    if hasattr(obj, '__stream_map__'):
+        return obj.__stream_map__(partial(_stream_map, func=func), **kwargs)
+    else:
+        sm = stream_map.dispatch(type(obj))
+        if sm is base_stream_map:
+            return func(obj, **kwargs)
+        else:
+            return sm(obj, partial(_stream_map, func=func), **kwargs)
+
+
+
 @singledispatch
 def stream_map(obj, func, **kwargs):
     return func(obj, **kwargs)
+
+
+base_stream_map = stream_map.dispatch(object)
 
 
 @singledispatch
--- a/streams/__init__.py
+++ b/streams/__init__.py
@@ -1,2 +1,3 @@
 from .core import *
 from .sources import *
+from .batch import Batch
--- a/streams/batch.py
+++ b/streams/batch.py
@@ -1,0 +1,17 @@
+try:
+    from cytoolz import accumulate
+except ImportError:
+    from toolz import accumulate
+
+class Batch(tuple):
+    def __stream_map__(self, func):
+        return Batch(map(func, self))
+
+    def __stream_reduce__(self, func, accumulator):
+        if isinstance(accumulator, Batch):
+            return Batch(accumulate(func, self, accumulator[-1]))
+        else:
+            return Batch(accumulate(func, self, accumulator))
+
+    def __stream_merge__(self, *others):
+        return Batch(zip(self, *others))
--- a/streams/batch.py
+++ b/streams/batch.py
@@ -3,15 +3,21 @@ try:
 except ImportError:
     from toolz import accumulate
 
+from .core import no_default
+
 class Batch(tuple):
     def __stream_map__(self, func):
         return Batch(map(func, self))
 
     def __stream_reduce__(self, func, accumulator):
-        if isinstance(accumulator, Batch):
-            return Batch(accumulate(func, self, accumulator[-1]))
+        if accumulator is not no_default:
+            seq = accumulate(func, self, accumulator)
         else:
-            return Batch(accumulate(func, self, accumulator))
+            seq = accumulate(func, self)
+        next(seq)  # burn first element, this is the old accumulator
+        seq = Batch(seq)
+        acc = seq[-1] if seq else accumulator
+        return acc, seq
 
     def __stream_merge__(self, *others):
         return Batch(zip(self, *others))
--- a/streams/batch.py
+++ b/streams/batch.py
@@ -3,10 +3,13 @@ try:
 except ImportError:
     from toolz import accumulate
 
+import functools
 from .core import no_default
 
 class Batch(tuple):
-    def __stream_map__(self, func):
+    def __stream_map__(self, func, **kwargs):
+        if kwargs:
+            func = functools.partial(func, **kwargs)
         return Batch(map(func, self))
 
     def __stream_reduce__(self, func, accumulator):
--- a/streams/batch.py
+++ b/streams/batch.py
@@ -12,7 +12,7 @@ class Batch(tuple):
             func = functools.partial(func, **kwargs)
         return Batch(map(func, self))
 
-    def __stream_reduce__(self, func, accumulator):
+    def __stream_accumulate__(self, func, accumulator):
         if accumulator is not no_default:
             seq = accumulate(func, self, accumulator)
         else:
--- a/streams/dask.py
+++ b/streams/dask.py
@@ -3,23 +3,21 @@ from tornado.locks import Condition
 from tornado.queues import Queue
 from tornado import gen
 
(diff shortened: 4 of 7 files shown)


ordirules:
thanks for this. extremely busy. i may be delayed for this until next week (hopefully i can look sooner). but i am very interested in this.

streams/core.py
```diff
@@ -602,3 +601,32 @@ def flush(self, _=None):
         out = tuple(self.cache)
         self.emit(out)
         self.cache.clear()
+
+
+def _stream_map(obj, func=None, **kwargs):
+    if hasattr(obj, '__stream_map__'):
```
ordirules: If I understand correctly, can we interchange using `__stream_map__` or the `singledispatch` implementation? Do you use `singledispatch` just to define a mapping without having to modify the object itself?

ordirules:
I like this implementation, it completes the `__stream_wrap__` idea by allowing nesting. Nice implementation for point 1 and 2. However, I have a comment about point 3 you mention. The following discusses just this.

I don't think it resolves the usage I had in mind. However, maybe my idea of usage is not a good idea, and I am open to alternatives. I'll try to think of a better way. I think I'm set on the idea of wanting to compute a task graph with `dask`, without necessarily using `distributed`.

I would like to use the stream to construct a task graph of delayed objects, and then only specify `compute` in certain sections of the stream. This would give me more control over when functions compute. It would also allow me to transition from non-distributed to distributed regime with more ease. Also, caching is easy by sinking futures to a list.

Previously, what I did is just had the streams wrap the function with a decorator that was basically `delayed(unwrap_layer1( ...) )`. I externalized the knowledge of the unwrapping not from the data itself but to the stream (through an inherited `wrap` argument).

This however, doesn't work when we do it the other way around, move the knowledge of the unwrapping from the stream to the data. Since computations are delayed objects, which don't know what the resultant object is until the computation is performed, we can't rely on asking them if their resultant computation will be a function with a `stream_map` method. They will just say that it's a delayed object with no such attribute. We must then tell the stream itself to perform operations assuming a certain data type, even though it is unknown. I feel that the only way to resolve this is add a wrapper, which makes the streams messy.

However, there may be a better way and I am not seeing it correctly. I thought of following your suggestion of using `singledispatch` here but it seems that we'd run into an infinite recursive loop; the `Delayed` object will always return a `Delayed` object. I think this makes intuitive sense; it doesn't seem logically possible to devise a scenario to ask an object that doesn't know anything about itself (yet) to decode itself.

As far as I understand, you are assuming that each element in the stream will `emit` only when data has been processed. Data should then not be emitted if the task it is set to do is not completed. I am breaking this by allowing streams to pass `delayed` objects. So perhaps an initial point of discussion could be whether this is wise thing to try with streams, or whether this notion should be dropped.

I apologize again for the length. Please let me know if I misunderstood something. This also doesn't discuss the behaviour of `accumulate` yet as well.
Thanks for all the support. For what it's doing, this is very clean :-)


Oh, and for nested data, otherwise it works great, I tried this for example, two layer nesting:

```python
    # unwrap classes of type
    # dict(header=stuff, data=(dict(header=stuff, data=singleton)))
    class InnerLayer(dict):
        ''' The inner layer of the packet.
            Inherit dict to be simple.
        '''
        # ignore kwargs to be simple
        def __stream_map__(self, func):
            result = func(self['data_inner'])
            # re-wrap
            # I write "data_outer" instead of just "data" to distinguish both
            # layers for debugging
            # InnerLayer instantiation (__init__) is the same as dict
            return InnerLayer(header=self['header'], data_inner=result)
    
    class OuterLayer(dict):
        ''' The inner layer of the packet.
            Inherit dict to be simple.
        '''
        # ignore kwargs to be simple
        def __stream_map__(self, func):
            result = func(self['data_outer'])
            # re-wrap
            return OuterLayer(header=self['header'], data_outer=result)
    
    from dask.delayed import Delayed, delayed
    # is it's register and not the base function, this will check
    # nests
    from streams.core import stream_map
    
    
    
    # some data packet
    data = OuterLayer(header="outer header",
                data_outer=InnerLayer(header="inner header", data_inner=10))
    #data =InnerLayer(header="inner header", data_inner=10)
    from streams.core import Stream
    s = Stream()
    def addone(x):
        return x + 1 
     
    s2 = s.map(addone) 
    s2.map(print, raw=True)
    # first try
    s.emit(data)
```

ordirules:
I have gone through your logic more carefully. It should work with `delayed`. I got it to work. Thanks
here is an example:
```python
@stream_map.register(Delayed)
def stream_map(obj, func, **kwargs):
    return delayed(func)(obj)
# now add compute

data = 1
s = Stream()
s2 = s.map(addone)
s2.map(compute, raw=True).map(print)
s.emit(delayed(data))
```

Previously, I was defining my function to be:
```python
@stream_map.register(Delayed)
def stream_map(obj, func, **kwargs):
    return func(obj)
```
This is incorrect as we want the computation to be delayed, which in this case, means we don't want to run the function until the input argument is defined.


Sorry for the confusion and thanks for this PR. I will go through the rest a little later when I have time.

streams/core.py
ordirules: I think I understand better. `singledispatch` serves as a way to add a `__stream_map__` to an object already defined. Useful when the object itself generates new instances of itself. Correct?

ordirules:
I think this is perfect. `map` and `accumulate` i think are both pretty much resolved.

However, `merge` may take some time to wring out. It might be worth considering it as a separate PR. I see there potentially being 4 functions (from the functions you define) to discuss there when considering merge. Let's take the pandas DataFrame as an example and a dict of numpy arrays as a second example just to differentiate each:

1. `merge` : what to do with data that is merged together. 

ex: merging two pandas dataframes could result in a join of columns (or it could be defined as something else). For joins, certain assumptions would have to be made regarding collisions.

ex2: merging two dicts of numpy arrays would aggregated all keys in one bigger dict. Note collisions would need to be handled.

2. `partition` : how to bunch data (should it have knowledge of the stream document or should it still create a tuple)

ex: `partition(3)` could take every 3 pandas dataframe and append the rows together, creating extra columns if necessary. note that here that no collisions should ever occur.

ex2 : `partition(3)` would take the numpy arrays in the dictionaries and join them as  one big array with higher dimension. So 1D numpy arrays would be packed together as a big 2D array.
for ex : `partition(2)` of `{'array1' : [1,2]}`  and `{'array1' : [3,4]}` would yield `{'array1' : [[1,2],[3,4]]|}`

3. `concat` : how to break up tuples into one long list of stuff. should it also work on the stream document. Maybe we could call this `flatten` not `concat`.

ex: `concat` run on pandas dataframes could make each row in a dataframe become one unit in the stream. Perhaps `concat` could also take a number, for example `concat(2)` would give pairwise dataframes.

ex2 : `concat` for the numpy dict would break up a 2D array into a list of 1D arrays.
for ex : `{'array1' : [[1,2],[3,4]]}` would yield `{'array1' : [1,2]}` `{'array1' : [3,4]}`

4. `union` : like merge, but takes two streams and replays elements coming in streams one after the other.
For both pandas and numpy dicts (and anything else), I think this should yield the same result.
ex2 : `{'array` : [1,2]}` from stream `s1` and  `{'array` : [3,4]}` from stream `s2` would be emitted one after the other in `s1.union(s2)`.

and these functions could default to working as they normally would when `raw=True` (except union, whose behaviour doesn't change).

Anyway, that's my take on it from the perspective of the experimentalist. What do you think? It could be this could be simplified, or the base operations modified. @danielballan as well, any thoughts?

mrocklin:
Slowly getting back to this now.  (coming back from conferences with some backlog of work). 

In general I agree with your analysis.  I suspect that this approach leads us towards dispatching on everything.  I'm not sure how I feel about this.  The `streams` project today is pretty simple and easy to understand.  The changes presented here that allow for nesting of different behaviors are powerful, but really complicate things.  I'm going to let this linger for a bit to see if I can't find something better.

ordirules:
@mrocklin no worries. I agree with you. Perhaps a better solution would be to externalize these assumptions I mention. Let `partition`, `concat` etc combine/split tuples of entities together (regardless of their contents) and have the user define their own functions to work on them.
For example, for `merge`, it could be:
`s1.zip(s2).map(joincolumns, raw=True)`
etc

Anyway, no worries about the delay, I hope in time we can converge to something nice.

Also, [spark streaming](https://spark.apache.org/docs/latest/streaming-programming-guide.html) may be another interesting library to follow. Perhaps you could also update documentation with that link.

ordirules:
(I know you're busy @mrocklin but I'll leave this here in case we go back down this route, as it's something that needs to be addressed. We can proceed our way without addressing this fully just yet)

There seems to be an issue with `_stream_map`. If `_stream_map` eventually maps into the base function with just kwargs (no args) then there is an error. I didn't catch this earlier because I always supplied an arg. 

The problem is that the new code recursively calls `_stream_map` until it reaches the base function, but requires `obj` as the first argument during each call. The mapping functions can change the supplied arguments to the function, and that needs to be handled. I think this comes down to having to decided how to do the mapping. I think requiring a dispatch on the first argument only is likely sufficient? (but adds complexity)

Here is a patchy fix that I tested to highlight the issue:
```python
# dispatch on first arg
def _stream_map(*args, func=None, **kwargs):
    if len(args) > 0:
        obj = args[0]
        if hasattr(obj, '__stream_map__'):
            # assume map is done on first arg only
            return obj.__stream_map__(wraps(func)(partial(_stream_map, *args[1:],func=func, **kwargs)))
        else:
            # assume map is done on first arg only
            sm = stream_map.dispatch(type(obj))
            if sm is base_stream_map:
                # run on args instead
                return func(*args, **kwargs)
            else:
                # assume map is done on first arg only
                return sm(obj, wraps(func)(partial(_stream_map, *args[1:], func=func, **kwargs)))
    else:
        return func(*args, **kwargs)
```

ordirules:
I would say that this PR is no longer necessary. It seems like things like `scatter(stream)` is better.

For subclassing, @CJ-Wright and I have been using `mylib.map(func, stream)` for example for mapping which returns a subclass of `Stream` where `mylib` is a library that implements classes that subclass streams (or they could be functions).

What do you think? This addressed adding `__stream_map__` etc which I think we all agree at this point is not necessary. (Please correct me if wrong)

mrocklin:
Sure

limx0:
https://github.com/mrocklin/streamz/pull/13#issuecomment-332880425 @ordirules can you provide an example of this?

CJ-Wright:
@limx0 if you meant an example with subclassing https://github.com/xpdAcq/SHED/blob/master/shed/event_streams.py

ordirules:
@CJ-Wright  maybe also include the use of the subclassed stream in a working pipeline, here's the link (this link is CJ's work, I'm just passing it on :-)):
https://github.com/xpdAcq/xpdAn/blob/master/xpdan/pipelines/main.py

CJ-Wright:
Thank you (I'm on my phone and have limited linking capacity)

limx0:
Thanks guys, this must have been fixed in one of the many recent commits. Previously (I believe) I was unable to so the following:

```
from streamz import Stream

class FilterEven(Stream):
    def __init__(self, child):
        Stream.__init__(self, child)
        if x % 2 == 0:
            self.emit(x)

class MyStream(Stream):
    
    def __init__(self, child=None, children=None):
        super().__init__(child=child, children=children)
        
    def filter_even(self):
        return FilterEven(self)

s = MyStream()

# This would raise an error, 
# MyStream has no attribute unique() - not being inherited by Stream
s.filter_even().unique().sink(print) 
```

CJ-Wright:
My suggestion would to use the class directly rather than relying on the method call from the base class, it's just syntactical sugar. I don't think we found a good way to append methods onto the base class.

ordirules:
to elaborate on @CJ-Wright, what would resolve this is basically:
```python
s = MyStream()
sout = FilterEven(s)
# you can use the shortcut syntax here since sout inherits from Stream
sout = s.unique()
sout.map(print)
```
the syntactical sugar is nice but it seems to lead to too many complications (like re-implementing every function as you said).

There likely is a way to resolve this but we've moved on to exploring use cases for now as far as I understand. I suggested a way here in #25 where the class is dynamically created at the time the method is called. This dynamic allocation creates a new class which subclasses from the class it's in (so `map` etc is not locked into a particular class). It's old and I've also left it aside since I think we need more use cases before we move this step. I'm still interested to discuss that (if it is useful).

ordirules:
umm sorry, what you wrote would work...
this wouldn't however:
```python
s = MyStream()
sout = FilterEven(s)
# you can use the shortcut syntax here since sout inherits from Stream
sout = sout.unique()
# unique is subclass of Stream only, you're no longer class MyStream, so you can't do this
sout = sout.filter_even()
sout.map(print)
```

limx0:
So really I want to be able to move between my custom transforms and the inbuilt ones. So you're saying the following wouldn't work?

`s = MyStream().filter_even().unique().filter_even().accumulate().sink(print)`

@CJ-Wright it is syntactical sugar, but it also makes your pipeline definition readable. Otherwise using the classes, correct me if I am wrong, it would be
```
s = MyStream()
x = FilterEven(Unique(FilterEven(Accumulate(Sink(print, s)))))
```

ordirules:
exactly, unless if you copied the whole stream class and added that one function, and imported that class rather than the stream class. That's another solution.

I think something near #25 would resolve this, but it may not be the cleanest implementation. We have to have stream methods dynamically create classes rather than import the old ones

ordirules:
I also recommend using `sink_to_list`. `Sink(print)` won't output anything, just print to screen. And you have it backwards (inner calls happen first)
```python
s = MyStream()
s2 = FilterEven(Unique(FilterEven(Accumulate(s))))
L = s2.sink_to_list()

```

limx0:
Sure thing, that was just for demonstration purposes. I suppose the classes aren't _terrible_. I'll take a look at #25, but good to know that it can be done! 

Thanks gents

mrocklin:
@limx0 short answer is that you can't easily subclass streams, nor is there a dispatch mechanism to add API.  

I don't currently see a clean way to do subclassing, to resolve this I'm resorting to encapsulation, making classes that contain a stream instead of are a stream.  A good example of this is the [pandas-streaming](https://github.com/dask/pandas-streaming) prototype.

mrocklin:
Providing a dispatch mechanism in the future is probably possible.

PR #16, opened 2017-06-01
ENH: add starmap

CJ-Wright:
Sometimes you just want unpacked things


CJ-Wright:
Closed as @ordirules mentioned, it is easier to just wrap the function in a decorator which performs this.

PR #18, opened 2017-06-01
added wraps to function

ordirules:
small suggestion for ongoing PR, but may as well send as a note before I forget
(still would need to do accumulate)


mrocklin:
This seems to be unnecessary since we removed protocols

PR #20, opened 2017-06-01
Split

CJ-Wright:
Split the stream class into a base class (which has emit, parents, child, etc.) and the Stream class which has everything else

### DIFF
--- a/streams/core.py
+++ b/streams/core.py
@@ -348,11 +348,11 @@ class Stream(StreamBase):
         return self.scan(update_frequencies, start={})
 
 
-class Sink(Stream):
+class Sink(StreamBase):
     def __init__(self, func, child):
         self.func = func
 
-        Stream.__init__(self, child)
+        StreamBase.__init__(self, child)
 
     def update(self, x, who=None):
         result = self.func(x)
@@ -362,13 +362,13 @@ class Sink(Stream):
             return []
 
 
-class map(Stream):
+class map(StreamBase):
     def __init__(self, func, child, raw=False, **kwargs):
         self.func = func
         self.kwargs = kwargs
         self.raw = raw
 
-        Stream.__init__(self, child)
+        StreamBase.__init__(self, child)
 
     def update(self, x, who=None):
         if not self.raw and hasattr(x, '__stream_map__'):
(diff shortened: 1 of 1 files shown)


mrocklin:
FYI I may not have the bandwidth to keep up with you guys (also including @ordirules ).  Feel free to fork.

CJ-Wright:
Do you want an org for this? (Potentially with multiple people with merge rights?)

mrocklin:
No strong thoughts.  You should organize however makes sense for you all.

ordirules:
I definitely agree we need to  subclass streams. However, I think there may be a cleaner approach, which I believe @mrocklin was suggesting (please correct me if wrong).

I think we could try setting up some global function registry by registering and unregistering functions. With functions that exist, it's easy with `singledispatch`. For example, to redefine map, just add the `@singledispatch` decorator to the map class and then you can do the following in some test script:
```python
from streams import Stream, map
from functools import singledispatch
# register as a singledispatch
@map.register(object)
class newmapclass(Stream):
    def __init__(self, func, child, raw=False, **kwargs):
        print("new stream")
        self.func = func
        self.kwargs = kwargs
        self.raw = raw

        Stream.__init__(self, child)

        return self.emit(self.func(x, **self.kwargs))


# or a new function name added as the following
# the singledispatch wrapper is just to allow this function also eventually be
# renamed
class NewStream(Stream):
    def newmap(self, func, **kwargs):
        return newmapclass(func, self, **kwargs)


# TODO : need to develop a function registry for streams
s = Stream()
s2 = s.map(lambda x : x+1).filter(lambda x : x > 0).map(lambda x : x+1)
s2.newmap(lambda x : x + 1)#.filter(lambda x : x > 0).map(lambda x : x + 1)
s.emit(1)
```

However, for functions that don't exist, I'm not sure what the best, most pythonic approach is. @mrocklin if you have any pointers, I'd appreciate it. I do understand you're busy so no worries.


Also, @CJ-Wright @danielballan and I have discussed and decided we will move discussion here:
https://github.com/xpdAcq/redsky/pull/13/files
(and any subsequent PR's regarding this).

CJ-Wright:
I'm closing this PR since I think the discussion has moved to how to override/dispatch existing methods rather than copy pasting the class

PR #21, opened 2017-06-08
ENH: Add Stream classes that expect DataFrames.

danielballan:
This is a modest start at integrating DataFrames with streams, as suggested in the blog post. Example:

```python
In [54]: source = Stream()

In [55]: df = pd.DataFrame({'a': [1,2,3], 'b': [2,3,4]})

In [56]: import streams.dataframe as sd

In [57]: sd.max(source).map(print)
Out[57]: <streams.core.map at 0x111f20940>

In [58]: source.emit(df)
a    3.0
b    4.0
dtype: float64
Out[58]: []

In [59]: source.emit(df)
a    3.0
b    4.0
dtype: float64
Out[59]: []

In [60]: source.emit(df * 2)
a    6.0
b    8.0
dtype: float64
Out[60]: []
```

### DIFF
--- a/streams/dataframe.py
+++ b/streams/dataframe.py
@@ -1,0 +1,44 @@
+import pandas as pd
+from .core import Stream
+
+
+class sum(Stream):
+    def __init__(self, child):
+        self.cumsum = pd.Series()
+        Stream.__init__(self, child)
+
+    def update(self, x, who=None):
+        self.cumsum = self.cumsum.add(x.sum(), fill_value=0)
+        self.emit(self.cumsum)
+
+
+class mean(Stream):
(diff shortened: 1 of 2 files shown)


mrocklin:
This may be an alternative: https://github.com/mrocklin/pandas-streaming

danielballan:
Interesting. What you've done with the `example` attribute is a good idea.

mrocklin:
Stolen from dask.dataframe, which does the same thing.

jrmlhermitte:
Great start :-)
I think this has been superseded with #73 .

danielballan:
Indeed!

PR #25, opened 2017-06-16
added subclassing to map

ordirules:
this is an initial thought. i think it ties strongly to #21 which is subclassing, and is a continuation of the discussion in #20. The stream class is dynamically created. The new class inherits all subclasses of the `child`. It also inherits the base class of the `child` only if the `child` does not have an `update` method (if it does, then it's not a base stream class).

I think for methods like `zip` we need to double check that all children share the same base class, then do the same creation. If we assume that each function contains a `child` or `children` kwarg, then this could perhaps be done with some decorator or metaclass. I have not gone this far yet.

Note that the `StreamBase` separation was unnecessary for this implementation, but I think it looks cleaner.

For example:
```python
# what plusone inherits is irrelevant
class plusone(Stream):
    def __init__(self, child, raw=False, **kwargs):
        # Need to take child or children as kwarg
        if child is None:
            raise ValueError("Child cannot be None")
        self.kwargs = kwargs
        self.raw = raw 

        Stream.__init__(self, child)

        # just emit 1
        result = x + 1 

        return self.emit(result)

class newStream(Stream):
    def plusone(self, **kwargs):
        return self.makestream(self, **kwargs, stream_method=plusone)

L = list()
s = newStream()
s2 = s.plusone()
s3 = s2.map(lambda x : x + 1)
s4 = s3.plusone()
s4.map(L.append)

s.emit(1)
s.emit(1)
print(L)
assert (L == [4,4])
```

### DIFF
--- a/streams/core.py
+++ b/streams/core.py
@@ -96,7 +96,7 @@ class StreamBase(object):
         self._loop = IOLoop.current()
         return self._loop
 
-    def _makestream(self, *args, **kwargs):
+    def makestream(self, *args, **kwargs):
         '''
             This is a wrapper to allow the dynamic creation of stream
             subclasses. The point of this is to allow dynamic subclassing.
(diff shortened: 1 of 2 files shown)


ordirules:
#74 resolves this very nicely

PR #28, opened 2017-07-05
Remove protocols

mrocklin:
This removes protocols.  This is messy.  I think it's possible that the right solution here just isn't to try to do this at all, and instead leave this burden on the user.  

Is this feasible in practice?


CJ-Wright:
I think that that is reasonable (at least for my own work). I have found that many times when I want to override the behavior for these stream functions I want to play with a lot more of the internals than the protocol allowed.

mrocklin:
This was my experience as well.  I found myself faced with the options of "make everything a protocol" or "make nothing a protocol".  I may take a crack at working with a streaming dataframe implementation and am pretty much planning to just encapsulate a vanilla Stream object, which seems to be somewhat clean.

mrocklin:
@ordirules how do you feel about this change?  It would stop most attempts to make streams polymorphic.

ordirules:
I agree with this. I think the current strategy we've taken is to fork or subclass streams. It makes more sense to fork this protocol-free version. This protocol strategy has its own merit though, and I'd suggest keeping it as a separate branch we can update (I will myself somewhere).

For example, @CJ-Wright , I don't think you need this line at all (and other simliar lines): https://github.com/xpdAcq/redsky/pull/13/files#diff-1893b7ca95af11be75f00bf53d93de96R72

Thanks @mrocklin for all the efforts in working with us. It's much appreciated. Streams is a quite nice way to formalize things that need to be book-keeped and distributed.

CJ-Wright:
@ordirules raised a good point (via gchat) how does this change impact the dask methods/classes?

mrocklin:
I'm somewhat inclined to drop the protocol approach to Dask as well.  I think it might make sense for Dask to separately implement an API compatible implementation, perhaps by just full code-duplication.  Thoughts?

CJ-Wright:
Full code-duplication may be a way to go for now (I've done it a bunch). There is nothing stopping us from making the code more compact if we find a clever way to implement it.

ordirules:
I agree it would be more practical for now. Anyway, I agree with this PR as is, thanks for mentioning and showing this to me @mrocklin.

As an aside, I think this brings up the question of subclassing, how can it be done cleanly? I think we need the base stream class to construct the stream class with a class constructor function.  I suggested something [here](https://github.com/mrocklin/streams/pull/25/files) (see `test_subclass`). But there may be a better way. Do you agree streams could be better subclassed or do you think it's cleaner to keep making new copies each time? I don't have a preference but for our applications allowing subclassing doesn't seem to add too many hindrances.

mrocklin:
Do you all depend on the current dask stream implementation?  Would you object to it being removed?  We can also revert back to the previous implementation before protocols.

CJ-Wright:
I don't currently use the dask implementation so no objection here, although I wanted to start testing it out soon (eg would it run with my weird subclass stuff?)

elementace:
I echo CJ's point in wanting to start testing it. After playing with streams for a while I just rolled my own subclasses to build fairly complex dags (multiple in/out flows between multiple stream objects)

CJ-Wright:
Should we make a separate issue for discussion of the Dask API/setup? I'm not certain I grokked it while it existed.

ordirules:
I haven't used it either yet. I have been turning certain components into dask delayed instances and then computing later on (so distributing or not can be easily interchanged by defining the client). Not the best way of course (since things are no longer computed as it passes through stream).
So I'm okay with it as well. thanks!

mrocklin:
OK, merging this in.

mrocklin:
> Should we make a separate issue for discussion of the Dask API/setup? I'm not certain I grokked it while it existed.

Sure?  If I had to do this immediately my plan would be to encapsulate a stream object rather than subclass.

```python
class DaskStream(object):
    def __init__(...):
        self.stream = stream
```

This is the approach that I've taken in [this pandas-streaming](github.com/mrocklin/pandas-streaming) micro-project and it seems to be going well so far.

ordirules:
Thanks for sharing that library!

ordirules:
PS: I think `.stream` is needed (here)[https://github.com/mrocklin/pandas-streaming/blob/master/pandas_streaming/tests/test_core.py#L97]. nice implementation

PR #29, opened 2017-07-06
Scan state and map-args

mrocklin:
This allows scan/accumulate functions to return the state as well as an output.

This also allows `*args` in `map`


mrocklin:
These were both necessary for https://github.com/mrocklin/pandas-streaming

ordirules:
This is great! This is something I also needed in my code (except it wasn't backwards compatible as you did here with `return_state`).

Issue #22, opened 2017-06-09
Combine latest on some


CJ-Wright:
In conversation with @ordirules and @mrocklin 

The idea is that you may only want combine latest to trigger on some of the updates.

ordirules:
I think for simplicity, maybe `combine_latest` should just assume that you only emit on the first child, which means changing this line in `combine_latest`:
```python
        if not self.missing and who == self.children[0]:
```

thus this will emit once:
```python
s1 = Stream()
s2 = Stream()
sout = s1.combine_latest(s2)
sout.map(print)

s2.emit(1)
s1.emit(1)
```

In the long run, I think a trigger should be added. Perhaps add a `trigger` flag to the `buffer` stream element?

What do you think? I think it's simpler than anything else I can think of. Also, we should fork streams itself (separate from event model). It could be on your or my branch, up to you.

mrocklin:
There are definitely cases where people want to emit on updates from
multiple children.  I think that one would take in a list of booleans in a
trigger= or emit_on= keyword for which children should activate an emit.

On Mon, Jun 12, 2017 at 10:52 AM, Julien Lhermitte <notifications@github.com

CJ-Wright:
I like the idea of a kwarg for this.

I guess it would be like:
```python
emiting_children = [child for child, t in zip(self.children, trigger) if t is True]
if not self.missing and who in allowed_emiting:
```

ordirules:
Sorry yes, that was too restrictive. I like that too.
Maybe have the keyword be the list (or one element) of numerically ordered children as a compromise?
(so `s.combine_latest(emit_on=0)` would be fine)
I suggest a PR somewhere? Maybe @CJ-Wright self-PR from your feature branch to your `master` and we can work from there? (Unless @mrocklin  you want to do it)

CJ-Wright:
Closed by #32

PR #32, opened 2017-07-07
emit on combine latest

CJ-Wright:
I'm not certain how I forgot to PR this

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: python
 
 matrix:
   include:
-    - python: 2.7
+#    - python: 2.7
     - python: 3.6
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: python
 
 matrix:
   include:
-#    - python: 2.7
+    - python: 2.7
     - python: 3.6
 
 install:
--- a/streams/core.py
+++ b/streams/core.py
@@ -202,13 +202,20 @@ class Stream(object):
         """ Add a time delay to results """
         return delay(interval, self, loop=None)
 
-    def combine_latest(self, *others, **kwargs):
+    def combine_latest(self, *others, emit_on=None):
         """ Combine multiple streams together to a stream of tuples
 
         This will emit a new tuple of all of the most recent elements seen from
         any stream.
+
+        Parameters
+        ---------------
+        emit_on : stream or list of streams or None
+            only emit upon update of the streams listed.
+            If None, emit on update from any stream
+
         """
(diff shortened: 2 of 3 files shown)


CJ-Wright:
closes #22

.travis.yml
```diff
@@ -4,7 +4,7 @@ language: python
 
 matrix:
   include:
-    - python: 2.7
+#    - python: 2.7
```
mrocklin: Was this intentional?

streams/core.py
```diff
@@ -524,17 +530,23 @@ def update(self, x, who=None):
 
 
 class combine_latest(Stream):
-    def __init__(self, *children):
+    def __init__(self, *children, emit_on=None):
```
mrocklin: I recall that earlier you used numbers here to refer to the children by index location.  I can imagine that also being useful.  Perhaps we can accept these as well and normalize to the streams?

```python
emit_on = tuple(children[x] if isinstance(x, int) else x for x in emit_on)
```

streams/core.py
```diff
@@ -524,17 +530,23 @@ def update(self, x, who=None):
...
         self.last = [None for _ in children]
         self.missing = set(children)
+        if emit_on is not None:
+            if not hasattr(emit_on, '__iter__'):
+                emit_on = (emit_on, )
```
mrocklin:
```python
if not isinstance(emit_on, Iterable):
```

.travis.yml
CJ-Wright: I guess it was since travis is complaining. It seems that [this](https://github.com/mrocklin/streams/pull/32/files#diff-14d53a4d1028cb6ff92ab388f7274169R533) is not valid syntax.

streams/core.py
```diff
@@ -524,17 +530,25 @@ def update(self, x, who=None):
...
 class combine_latest(Stream):
-    def __init__(self, *children):
+    def __init__(self, *children, emit_on=None):
         self.last = [None for _ in children]
         self.missing = set(children)
+        if emit_on is not None:
+            if not isinstance(emit_on, Iterable):
+                emit_on = (emit_on, )
+            emit_on = tuple(
+                children[x] if isinstance(x, int) else x for x in emit_on)
```
mrocklin: This behavior could use a test

streams/core.py
mrocklin: Right, you might have to do `**kwargs` and then `emit_on = kwargs.pop('emit_on', None)`

mrocklin:
Thanks @CJ-Wright !

PR #34, opened 2017-07-13
ENH: add str method to streams

CJ-Wright:


streams/core.py
```diff
@@ -59,6 +59,13 @@ def __init__(self, child=None, children=None, **kwargs):
             if child:
                 child.parents.append(self)
 
+    def __str__(self):
+        s = self.__class__.__name__
+        for m in ['func', 'predicate']:
+            if hasattr(self, m):
+                s += ', {}'.format(getattr(self, m).__name__)
+        return s
```
mrocklin: It might be nice to bound this with `<...>`, which might help disambiguate things if, for example, there are many of them in a list.  Similarly I suggest using colon instead of a comma.

mrocklin: We might also consider adding `func=` or `predicate=` and then extending this list to other common-and-easy-to-render attributes like `n` and `interval`

CJ-Wright: Both these are good ideas, will implement. As a side note, this is how I made the pipeline graph for the abstract that got sent around (with some extra stuff that is currently in https://github.com/xpdAcq/redsky/pull/15, which I may try to PR into this as a utility if there is any interest).

mrocklin:
Thanks @CJ-Wright !

PR #37, opened 2017-07-19
Add DaskStream prototype

mrocklin:
CC @CJ-Wright @ordirules 

### TODO

1.  Support more operations (if desired)
2.  Add `asynchronous=True` keywords to scatter/gather to ensure asynchronous behavior (need to add this to dask first)
3. Add synchronous tests


mrocklin:
https://github.com/dask/distributed/pull/1272

CJ-Wright:
*Disclaimer: I understand that my subclassing of these things is rather janky and thus may not be of interest/applicable so feel free to ignore this.*

What is the best way to distribute variables from a stream node among the dask processes?

Eg I have a stream which is going to add 1 to every element of the stream. At the start of the stream I make a UID, which I store as an attribute in the stream class, which needs to be distributed to all the subsequent data points. I'd still like to run this via the dask support. Is it possible to inject a variable somewhere so that all the jobs from a given stream node can get at it? How could that be done? Will tasks wait for that variable to be created if it waiting on a previous node to create a UID?

mrocklin:
Perhaps you could scatter your value as a future and then include that in a map call?

```python
future = client.scatter(my_uid)

source = Stream().scatter()
a = source.map(my_func, uid=future)
```

streams/dask.py
```diff
@@ -0,0 +1,90 @@
...
+        self.returns_state = returns_state
+        DaskStream.__init__(self, child)
+
+    def update(self, x, who=None):
+        if self.state is core.no_default:
+            self.state = x
+            return self.emit(self.state)
+        else:
+            client = default_client()
+            result = client.submit(self.func, self.state, x)
```
CJ-Wright: How does this work? 

Lets say that our data is `range(10)`, we scatter it to the cluster and then start performing an `add` so we get a cumulative sum. The cumulative sum is in `self.state`, right? How do the other tasks access that state?

mrocklin: In this case `self.state` is a Future pointing to data tracked on the cluster.  As more elements stream through this `self.state` attribute will point to different futures.  These futures may exist on different machines.  This is actually fairly likely if the state is much smaller than the elements of the stream.

CJ-Wright: Ah ok, I think I get it. The scatter operation puts the data onto the cluster, when then the various `DaskStream` objects operate upon that data by manipulating Futures and/or submitting functions against those Futures.

CJ-Wright: And just to finish, the functions are handed to the scheduler which handles the complexity of the actual execution order (eg properly fanning out the summation into a tree).

mrocklin: The scheduler doesn't contain parallel algorithms like this, it just executes what we told it to do.  In a case of `scan(add)` we're giving it a sequential algorithm, which it will execute faithfully on parallel hardware.  Most of the benefit of parallelism will probably come from maps.

CJ-Wright: Hmm ok

ordirules:
this looks great, i will test a little later, @mrocklin if you could rebase that would be great (just to test the other small additions)

streams/tests/test_dask.py
```diff
@@ -0,0 +1,85 @@
...
+
+@gen_cluster(client=True)
+def test_map(c, s, a, b):
+    source = Stream()
+    futures = scatter(source).map(inc)
+    futures_L = futures.sink_to_list()
+    L = futures.gather().sink_to_list()
+
+    for i in range(5):
+        yield source.emit(i)
```
ordirules: is it possible to not have to yield? It would be nice not to have to worry about that. I am not very familiar with asynchronous environments yet (if you have recommended documentation I'd be happy to read)

streams/tests/test_dask.py
```diff
@@ -0,0 +1,85 @@
...
+
+    yield a.emit(1)
+    yield b.emit('a')
+    yield a.emit(2)
+    yield b.emit('b')
+
+    assert L == [(1, 'a'), (2, 'b')]
+
+
+def test_sync(loop):
```
ordirules: This is a technical question. I use nosetests and not pytest. This fails in nosetests (but not pytest), as it does not understand where to obtain the `loop` argument. Where does loop come from?
Running:
```python
def test_sync():
    loop = IOLoop()
```

resolved the issue. thanks!

CJ-Wright: `loop` comes from an import at the top
`from distributed.utils_test import gen_cluster, inc, cluster, loop`

mrocklin: Dask testing generally depends depends on pytest for testing, not nose.  `loop` is a pytest fixture.  It should be easy to swap out your use of `nosetests` for `py.test` without difficulty.

streams/tests/test_dask.py
mrocklin: Not currently.  This currently only works in asynchronous environments.

ordirules:
Could we add support for arbitrary objects? I think it'd be useful. Basically add the following test (it's already supported):
```python
@gen_cluster(client=True)
def test_map_pickle(c, s, a, b):
    ''' test pickling an arbitrary object.'''
    class myObj:
        pass

    def make_new_obj(x):
        res = myObj()
        res.x = x
        return res

    source = Stream()
    futures = scatter(source).map(make_new_obj)
    futures_L = futures.sink_to_list()
    L = futures.gather().sink_to_list()

    yield source.emit(0)

    assert L[0].x == 0
```

ordirules:
I think this looks good. Running in an `IOLoop` is a drawback but the user could just run `.map(lambda x: client.submit(f, x)).map(lambda x: client.submit(f2, x)).map(client.gather)` themselves to go around it I guess. What I've been doing is to just physically add in wrappers for now, for ex: `.map(dkf(func)` etc. where `dkf` would be some wrapper. Not as clean as it could be, but I feel more comfortable in the sense that I can just use your code as is and contribute if possible (which is much cleaner!).

I'm wondering about `filter`. Is it possible and worth trying? I imagine for each `filter`, you need to spawn a new client process running asynchronously each time. Some `max_clients` should then be observed.
Does that sound right? Would it be worth an effort? The alternative is of course to `gather` before each `filter`.

mrocklin:
This now support synchronous (normal) syntax.  It requires changes in dask/distributed master

mrocklin:
> I'm wondering about filter. Is it possible and worth trying? I imagine for each filter, you need to spawn a new client process running asynchronously each time. Some max_clients should then be observed.

You would reuse the same client.  I suspect that a filter would be implemented by submiting a predicate on each future, waiting for the result to come back and emitting the original future if the predicate came back true.  This would have to be built out with async syntax to work nicely.  Alternatively we could also use dask.distributed queues.  

If you haven't seen it already you might enjoy reading through this doc, especially towards the middle and end: http://dask.pydata.org/en/latest/futures.html

ordirules:
Very interesting indeed! I must have missed that on the first pass, but then again I had much less knowledge of `distributed` back then.
One question about the seceding:

>However, each running task takes up a single thread, and so if you launch many tasks that launch other tasks then it is possible to deadlock the system if you are not careful. You can call the secede function from within a task to have it remove itself from the dedicated thread pool into an administrative thread that does not take up a slot within the Dask worker:

What does it mean to move to an administrative thread? Is there no load balancing on the slots in the dask worker (i.e. they're all given the same priority or something similar)? And the administrative thread does have priority?

mrocklin:
Dask workers have a thread pool executor that is used to run tasks.  This thread pool is limited.  Seceding moves the thread out of the thread pool.

If possible, can I ask that we move general Q&A out to Stack Overflow under the #dask tag?

ordirules:
Thanks, and thanks for suggesting that, it wasn't very considerate of me! (I'd like to give back if I can in any way for what you've provided)

I asked a slight variant here with the concrete example, I'm still not 100% sure if I should be using `secede` in this case:
https://stackoverflow.com/questions/45357277/seceding-tasks-from-thread-pools-in-dask-distributed

PR #39, opened 2017-07-19
Add more string goodness

CJ-Wright:
Sorry I don't know why I didn't have this in the previous PR. The idea is that streams can have names for more descriptive power.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 import sys
 import pytest
-import os
 
 if __name__ == '__main__':
     # show output results from every test function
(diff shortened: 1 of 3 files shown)


CJ-Wright:
@mrocklin @ordirules ready for review.

ordirules:
I don't have much opinions. I'm actually not familiar with the use cases of `str(foo)`. All I've used it for is for logging, in which case I'd agree. So I'm fine either way.

mrocklin:
Same.  Merging.

Also, FWIW, I've been squash-merging on PRs with many commits that boil down to one objective.  Hope you're ok with this.

CJ-Wright:
That's fine with me

CJ-Wright:
There purpose of this is for the plotting if the graph which will use the string data in the node names

PR #40, opened 2017-07-19
Graph3

CJ-Wright:
Use this one not #38


mrocklin:
Is it possible to test this code?

CJ-Wright:
one sec

CJ-Wright:
Note that this doesn't do a good job with cyclic graphs since sink doesn't know what it is sinking to. Eg if we use sink to take data from one stream and emit it into another. This may be possible if we use map?

CJ-Wright:
I don't want to try to test the plotting itself, although I have left an easy enough to run example in the examples folder. (mostly because I don't want to try to install graphviz/pygraphviz in travis). (making a conda package for that is on my todo list)

mrocklin:
Just as long as we can test the majority of the code leading up to the plotting I'm happy.  Another thought is that we might choose to conform to the `visualize` API within Dask if we expect to share a userbase.  This might reduce surprise.  https://github.com/dask/distributed/pull/1270

mrocklin:
Whoops, that was the wrong link entirely: http://dask.pydata.org/en/latest/delayed-overview.html?highlight=visualize

CJ-Wright:
Ok I can add a method onto streams which will plot the graph. I'll look at the dask code to try to match the app.

CJ-Wright:
Do you mind if I borrow the docs from Dask (and modify them as needed?)

mrocklin:
Go for it

CJ-Wright:
Ready for review.
I've revised things so that stream has a `visualize` method, although I didn't do all the dot ipython visualization, it still writes to disk the image though.

streams/graph.py
```diff
@@ -0,0 +1,64 @@
...
+        A node in the task graph
+    filename : str, optional
+        The name (without an extension) of the file to write to disk.
+    format : {'png', 'pdf', 'dot', 'svg', 'jpeg', 'jpg'}, optional
+        Format in which to write output file.  Default is 'png'.
+    """
+
+    g = nx.DiGraph()
+    create_graph(node, g)
+    create_graph(node, g)
```
mrocklin: Why twice?

CJ-Wright: sorry, that was from a refactor gone wrong. In the past I traversed the graph in two steps, thus the two calls.

CJ-Wright:
@mrocklin Do you want the new "filename has both name and format in it" visualization API or do you want to keep with dask's?

mrocklin:
I think in practice everyone uses dask's visualize function with a full filename.  I don't think I've seen anyone use the `format=` keyword in the wild.  I wouldn't mind deprecating it in the future.

CJ-Wright:
Ok I'll change the API, any idea why the conda graphviz install isn't taking?

mrocklin:
`graphviz` is the name of the system library, you might also want `python-graphviz` or `graphviz-python` or something like that.

CJ-Wright:
@mrocklin @ordirules ready for review/merge.

mrocklin:
This looks good to me.  I've added a small change to pass through keyword
arguments to graphviz.

Tried this out in a notebook and the results looks stellar!  thanks!

CJ-Wright:
Awesome, would there be any interest in putting an example image in the `README.md`?

ordirules:
looks awesome! i don't see any issues. good job!

streams/core.py
```diff
@@ -7,6 +7,7 @@
 from tornado.ioloop import IOLoop
 from tornado.queues import Queue
 from collections import Iterable
+from streams.graph import visualize
```
ordirules: one after the fact comment about this. I'd much rather not have the core import visualize, and also would be tempted to remove the member function `visualize`. I feel like it's too much overhead.
```python
from streams.graph import visualize
visualize(s1)
```
seems to me sufficient as opposed to `s1.visualize(....)`.

What do you think?

streams/graph.py
```diff
@@ -0,0 +1,136 @@
+"""Graphing utilities for EventStreams"""
+import networkx as nx
+import graphviz
```
ordirules: This seems to be a required dependency for streams now. I think it might be wiser to have it be a soft dependency.
Streams can likely be things run on clusters etc which don't really need rendering of graphics.

CJ-Wright: That's fair

streams/core.py
CJ-Wright: This comes from the dask api which had visualize as a method. But I don't have too much feeling either way. Although thinking about it now we may want the graph creation exposed as a method in case users are interested.

mrocklin:
Dask includes visualize as a method on dask objects as well as a top-level
dask.visualize function that takes in a list of objects to visualize all at
once.

PR #41, opened 2017-07-19
API : Changed default behaviour for scan method.

ordirules:
I would prefer the accumulator emitting a result on the first iteration. I know it's an API change, but the library is still young. Was there a motivation to choose otherwise? (note this would affect dask scan as well)

### DIFF
--- a/streams/core.py
+++ b/streams/core.py
@@ -419,6 +419,7 @@ class scan(Stream):
     def update(self, x, who=None):
         if self.state is no_default:
             self.state = x
+            return self.emit(x)
         else:
             result = self.func(self.state, x)
             if self.returns_state:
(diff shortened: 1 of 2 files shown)


mrocklin:
Sure, we'll see what happens.  Worst case scenario we can always revert it back to the old behavior.

ordirules:
Thanks, this avoids having to worry about an initialization case for us.
I meant to ping @CJ-Wright as well. He assumes more or less the same thing as well [here](https://github.com/xpdAcq/redsky/pull/13/files#diff-1893b7ca95af11be75f00bf53d93de96R613). (correct me if I'm wrong CJ)

CJ-Wright:
Yes, PySHED emits on all elements for the accumulate stream function

Issue #44, opened 2017-07-27
graph dependencies


ordirules:
Right now, graph

```python
from streams import Stream
s1 = Stream()
s2 = Stream()

sout1 = s1.map(lambda x : x +1)
sout1.map(print)

# connect stream 2 to stream 1
s2.map(s1.emit)

from streams.graph import visualize 
visualize(s2, 'stream2.png')
```

Result is this:
http://imgur.com/a/0Z0JA

What would solve this is that rather than using `emit`, we could define a new function function, say `s2.inject(s1)` which would add `s2` as a child to `s1`. The only issue I see here is we'll need an `update` defined for the base stream object, which will just call `emit`. What do you think?
@mrocklin @CJ-Wright 
I'm interested to hear of a better way, thanks!

ordirules:
I'm also happy to do the grunt work on a PR but just want consensus first.

CJ-Wright:
Closed by #45

PR #45, opened 2017-07-27
Connect streams

ordirules:
Suggested resolution to #44 . It makes a big difference when visualizing graphs and I'm using it now, so I thought I'd already go ahead and send it. Please let me know if there are better ways.

From issue, changing from `s2.map(s1.emit)` to `s2.connect(s1)` leads to difference of the two figures:

`s2.map(s1.emit)` : http://imgur.com/a/0Z0JA

`s2.connect(s1)` : http://imgur.com/a/rcrX0

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -108,13 +108,15 @@ class Stream(object):
         self.emit(x)
 
     def connect(self, parent):
-        ''' Connect another child to stream.
-            Note that parents go downstream and children go upstream.
+        ''' Connect this stream to a downstream element.
+
+            Parameters
+            ----------
+            parent: Stream
+                the parent stream (downstream element) to connect to
         '''
-        if self.parents == [None]:
-            self.parents = [parent]
-        else:
-            self.parents.append(parent)
+        # Note : parents go downstream and children go upstream.
+        self.parents.append(parent)
 
         if parent.children == [None]:
             parent.children = [self]
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -104,6 +104,9 @@ class Stream(object):
                 result.append(r)
         return [element for element in result if element is not None]
 
+    def update(self, x, who=None):
+        self.emit(x)
+
     @property
     def child(self):
         if len(self.children) != 1:
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -107,6 +107,20 @@ class Stream(object):
     def update(self, x, who=None):
         self.emit(x)
 
+    def connect(self, parent):
+        ''' Connect another child to stream.
+            Note that parents go downstream and children go upstream.
+        '''
+        if self.parents == [None]:
+            self.parents = [parent]
+        else:
+            self.parents.append(parent)
+
+        if parent.children == [None]:
+            parent.children = [self]
+        else:
+            parent.children.append(self)
+
     @property
     def child(self):
         if len(self.children) != 1:
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -567,15 +567,21 @@ def test_piping():
 
 
 def test_connect():
-    source = Stream()
-    source2 = Stream()
+    source_downstream = Stream()
+    # connect assumes this default behaviour
+    # of stream initialization
+    assert source_downstream.parents == []
+    assert source_downstream.children == [None]
 
-    sout = source.map(lambda x : x + 1)
+    # initialize the second stream to connect to
+    source_upstream = Stream()
+
+    sout = source_downstream.map(lambda x : x + 1)
     L = list()
     sout = sout.map(L.append)
-    source2.connect(source)
+    source_upstream.connect(source_downstream)
 
-    source2.emit(2)
-    source2.emit(4)
+    source_upstream.emit(2)
+    source_upstream.emit(4)
 
     assert L == [3, 5]
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -551,21 +551,6 @@ def test_triple_zip_latest():
     assert L_simple == [(1, 'III', 'a'), (2, 'III', 'a'), (3, 'III', 'b')]
 
 
-def test_piping():
-    source = Stream()
-    source2 = Stream()
-    # put source2 upstream of source
-    source2.parents = [source]
-
-    L = list()
-    source.map(L.append)
-
-    source2.emit(1)
-    source2.emit(3)
-
-    assert L == [1,3]
-
-
 def test_connect():
     source_downstream = Stream()
     # connect assumes this default behaviour
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -549,3 +549,19 @@ def test_triple_zip_latest():
     s3.emit('b')
     s1.emit(3)
     assert L_simple == [(1, 'III', 'a'), (2, 'III', 'a'), (3, 'III', 'b')]
+
+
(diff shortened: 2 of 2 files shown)


streams/core.py
```diff
@@ -105,6 +105,23 @@ def emit(self, x):
                 result.append(r)
         return [element for element in result if element is not None]
 
+    def update(self, x, who=None):
+        self.emit(x)
+
+    def connect(self, parent):
+        ''' Connect another child to stream.
+            Note that parents go downstream and children go upstream.
```
CJ-Wright: Can we put this somewhere special? I think this may trip potential new devs up and thus should be placed somewhere they'd find it.

mrocklin:
Can anyone think of a more specific name than connect?  Arguably all operations connect two streams.

ordirules:
`branch`?

mrocklin:
It might be useful to look at other projects to see if they have such a
construct and then see what names they use.

On Fri, Jul 28, 2017 at 12:30 PM, Julien Lhermitte <notifications@github.com

CJ-Wright:
`sink_to_stream`?

ordirules:
( @CJ-Wright agreed about examples I'll add examples later on this evening or worst case late weekend, occupied here).

ordirules:
I checked [spark-streaming](https://spark.apache.org/docs/latest/) and [apache flink](https://ci.apache.org/projects/flink/flink-docs-release-1.3/dev/datastream_api.html#datastream-transformations) but don't see something obvious as of now. Will check again later, I do like `sink_to_stream` though.

Another option is to have `map` and `sink` take a stream as an argument. (But I would vote against and rather be more explicit)

ordirules:
I have checked and as far as I can see, I don't see an obvious analogue to this. I am a bit surprised. I must have missed something.
The libraries I have gone through are:
* [spark-streaming](https://spark.apache.org/docs/latest/streaming-programming-guide.html)
* [apache flink](https://ci.apache.org/projects/flink/flink-docs-release-1.3/dev/datastream_api.html#datastream-transformations)
* [apache Beam](https://beam.apache.org/documentation/programming-guide/#transforms)
* [RxPy](http://reactivex.io/documentation/operators.html)

To be honest, I haven't been able to spend that much time, but I don't see anything.
I'd vote for `sink_to_list`. It could eventually be superseded by a briefer term.

streams/core.py
CJ-Wright: I meant more with the module level docs

elementace:
I'm not sure if you guys are interested; but for the case of multiple outputs from one stream object and other stream objects wanting to take those, I was outputting a dictionary in my emit statements and taking a secondary argument in my sink to stream which was the key.

elementace:
Sorry, on phone, will put this in my own pull request when I get to laptop

ordirules:
yeah PR would help, could you please mention us? (just in case)
Also, maybe syntax highlight the code you have using github's [markdown](https://guides.github.com/features/mastering-markdown/) syntax.
Please include an example or unit test in PR as well, it will save a lot of time to review/understand the code.

CJ-Wright:
Was there a spec for putting out multiple streams?

streams/tests/test_core.py
```diff
@@ -494,3 +494,34 @@ def test_partition_str():
 def test_stream_name_str():
     source = Stream(name='this is not a stream')
     assert str(source) == '<this is not a stream; Stream>'
+
+
+def test_piping():
+    source = Stream()
+    source2 = Stream()
+    # put source2 upstream of source
+    #source.children = [source2]
```
CJ-Wright: remove debug code?

CJ-Wright:
Aside from some cleanup LGTM

ordirules:
sounds good. personally, i like `subscribe` better as @elementace suggested.
but `connect` is also fine.

ordirules:
(and we need @mrocklin 's opinion on this before making any merge decision of course)

CJ-Wright:
One of the nice features I think that comes from this is that you can make a library(s) of streams, putting different chunks of a pipeline into modules, load them into a script, and use this PR's mechanism to plumb them all together.

streams/tests/test_core.py
```diff
@@ -494,3 +494,33 @@ def test_partition_str():
...
+    # put source2 upstream of source
+    source2.parents = [source]
+
+    L = list()
+    source.map(L.append)
+
+    source2.emit(1)
+    source2.emit(3)
+
+    assert L == [1,3]
```
mrocklin: This test doesn't seem to do much related to this PR

streams/core.py
```diff
@@ -105,6 +105,23 @@ def emit(self, x):
...
+        '''
+        if self.parents == [None]:
```
mrocklin: When does this happen?  Do we every introduce `None` into self.parents?

streams/core.py
mrocklin: The numpydoc standard recommends having a single-line description, followed by a blank line, followed by a more extended description if necessary.  https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections

http://dask.pydata.org/en/latest/develop.html#docstrings

streams/core.py
ordirules: looks like it doesn't. I think I mirrored `if parent.children == [None]` below, which does happen. I will remove it.
Might be an issue later to address? Why does a stream with no parents contain `parents=[]` whereas a stream with no children contain `[None]`? (should it be `children = []`?)

streams/tests/test_core.py
ordirules: you're right, I'll remove it (it's what `connect` is effectively doing behind the scenes).

streams/core.py
ordirules: thanks will change

streams/tests/test_core.py
```diff
@@ -494,3 +494,24 @@ def test_partition_str():
 def test_stream_name_str():
     source = Stream(name='this is not a stream')
     assert str(source) == '<this is not a stream; Stream>'
+
+
+def test_connect():
+    source_downstream = Stream()
+    # connect assumes this default behaviour
+    # of stream initialization
+    assert source_downstream.parents == []
```
ordirules: added these in test, this expected behaviour is essential to `connect`

ordirules:
FYI : it seems that `subscribe` is basically the same as `connect` in ReactiveX: 
http://reactivex.io/documentation/observable.html

It may make more sense to rename to this? @mrocklin, @CJ-Wright ? Any preference?

CJ-Wright:
I am not sure if we have a connect in the same way as RX does but I still don't have a preference for the name.

CJ-Wright:
@ordirules I am on board with `connect` or `subscribe` can you please rebase and go with one or the other? I'll try to merge tonight.

ordirules:
finished a run then dinner. will get back in 2 hrs

CJ-Wright:
Sorry I didn't mean to put you under pressure. Take your time.

ordirules:
no worries. just trying to emphasize im def motivated to have this one in!

ordirules:
pushed but didn't have time to check the coverage

CJ-Wright:
Codecov seems to be having issues as a check, here is the link
https://codecov.io/gh/mrocklin/streamz/tree/e4102be4d8ab0ccafd8360f339b4dad2c0524ee6/streamz
Coverage seems fine though, so merging

PR #46, opened 2017-07-31
added a test on pickling and caching

jrmlhermitte:
I think this general use would be useful for our library and it would be nice to support. Basically ensure the pickling of an arbitrary object and that the cluster caches previously computed results appropriately.

I create an arbitrary object with members `a` and `b` and `tokenize` it into a hash that only depends on `a` as a test.

What do you think? We already use this for the repetition of the calculation of large coordinate arrays.

thanks!

EDIT :  PS: I noticed I had to add a `time.sleep(.1)` or else I would get `CancellationError`s. I'm not sure how it's related (or if it is even).

### DIFF
--- a/streamz/tests/test_dask.py
+++ b/streamz/tests/test_dask.py
@@ -178,8 +178,6 @@ def test_obj(c, s, a, b):
                 a, b = args[0], args[1]
                 self.a = a
                 self.b = b
-            else:
-                raise ValueError("Num args must be 1 or 2")
 
     @normalize_token.register(myObj)
     def tokenize_myObj(obj):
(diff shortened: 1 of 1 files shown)


ordirules:
interesting. the caching doesn't seem to happen in py2.7

ordirules:
how is coverage being decreased? should codecov ignore the test files as well?

CJ-Wright:
The test files are included in the coverage to make certain that the tests run as expected.

ordirules:
good point. actually i cant see how im decreasing the coverage, ill need to check in more detail later, seems odd.

ordirules:
hmm
```
value = ValueError('Could not find dependent type-25247ac17b0664b534d8e17c13477ada.  Check worker logs',)
```
this happened before, I'm not sure where it comes from, I'll have to do some debugging, and eventually post to `distributed` if it's an issue....

ordirules:
I would say this is more of a dask distributed thing than a `streamz` thing. I will do some reading on `distributed` later to see if this exists. Last I remember, it's a supported feature but not documented just yet.

PR #47, opened 2017-08-03
ENH: Graph now contains the actual stream nodes.

CJ-Wright:
In case you wanted to play with the graph itself.

Potential use case:
Use a function to configure the task graph, then want to add things on.
We need to return a full representation of all the nodes, so you can
add to any one of them. The actual graph seems like a good way to do
that. Now if you want to add a sink to a node you just go to the graph,
get the stream_node for the associated node and run with that.

Thoughts?
(I'm worried about how to configure pipelines on the fly)

### DIFF
--- a/streams/graph.py
+++ b/streams/graph.py
@@ -17,7 +17,7 @@ def create_graph(node, graph, prior_node=None, pc=None):
     if node is None:
         return
     t = hash(node)
-    graph.add_node(t, str=str(node))
+    graph.add_node(t, str=str(node), stream_node=node)
     if prior_node:
         tt = hash(prior_node)
         if graph.has_edge(t, tt):


ordirules:
I don't have many thoughts. I think the addition is useful and doesn't hurt, but the graph data structure is already there in the stream itself. The use case I see for this is if you want to have an interactive GUI added on. In that case maybe something like [model-view](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) may be a good idea to follow.

CJ-Wright:
I personally don't want to build a GUI for this since that seems like a pain, but it could be interesting if anyone wants to build it. I thought that accessing the graph structure from the stream itself may be a bit difficult for the user. Eg If I only import the source and want to add some visualization onto a pipeline I'd have to access the other nodes, which would essentially require the user to build the graph itself (which this function does).

CJ-Wright:
For GUI ideas one may look at [VisTrails](https://www.vistrails.org/index.php/Main_Page)

jrmlhermitte:
I don't see harm in adding reference to node. However, maybe it should be a `weakref` so as to not disturb the stream functionality? (Also add a test)
Can you supply a working example of a use case using this?

CJ-Wright:
The `weakref` is fair.

Sure I can put something together. The main motivation is having access to all the nodes in a single object without having to walk the tree yourself. For example:
I configured a pipeline using a function and it only returned to me the source. Now I want to access other nodes to destroy them. How do I access those nodes, by looking a the graph.

PR #50, opened 2017-08-10
Zip latest

### LINKED ISSUE #48: combine_latest emit_on behavior
Should we buffer the `combine_latest` `emit_on` stream?

## Context
Consider the following situation, we have two streams `a` and `b`. We are going to use `combine_latest` to combine them, with an emit only on `a`. 5 entries come down from `a` then one comes down from `b` then one from `a`. Due to the lack of data from `b` none of the 5 initial `a` entries have been emitted since the `b` was missing. Now that we have `b` data do we expect it to emit 6 times (which would require the buffering of the `emit_on` stream or only once (which may violate the idea of the `emit_on` stream always emitting)?

@mrocklin @ordirules @danielballan

CJ-Wright:
Closes #48 
@limx0 care to look this over?

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -654,3 +654,44 @@ class collect(Stream):
         out = tuple(self.cache)
         self.emit(out)
         self.cache.clear()
+
+
+class zip_product(Stream):
+    """Combine multiple streams together to a stream of tuples
+
+    This will emit a new tuple of the elements from the lossless stream paired
+    with the latest elements from the other streams.
+    """
+
+    def __init__(self, lossless, *children):
+        """Initialize the node
+
+        Parameters
+        ----------
+        lossless : Stream instance
+            The stream who's documents will always be emitted
+        children: Stream instance
+            The streams to combine
+        """
+        children = (lossless,) + children
+        self.last = [None for _ in children]
+        self.missing = set(children)
+        self.lossless = lossless
+        self.lossless_buffer = deque()
+        Stream.__init__(self, children=children)
+
+    def update(self, x, who=None):
+        idx = self.children.index(who)
+        if who is self.lossless:
+            self.lossless_buffer.append(x)
+
+        self.last[idx] = x
+        if self.missing and who in self.missing:
+            self.missing.remove(who)
+
+        if not self.missing:
+            L = []
+            while self.lossless_buffer:
+                self.last[0] = self.lossless_buffer.popleft()
+                L.append(self.emit(tuple(self.last)))
+            return L
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -330,13 +330,13 @@ class Stream(object):
         """ Combine two streams together into a stream of tuples """
         return zip(self, *other)
 
-    def zip_product(self, *others):
+    def zip_latest(self, *others):
         """Combine multiple streams together to a stream of tuples
 
         This will emit a new tuple of the elements from the lossless stream paired
         with the latest elements from the other streams.
         """
-        return zip_product(self, *others)
+        return zip_latest(self, *others)
 
     def sink(self, func):
         """ Apply a function on every element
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@ordirules @danielballan

streams/tests/test_core.py
```diff
@@ -494,3 +494,39 @@ def test_partition_str():
...
+    L2 = d.sink_to_list()
+
+    a.emit(1)
+    a.emit(2)
+    b.emit('a')
+    b.emit('b')
+    a.emit(3)
+
+    assert L == [(1, 'a'), (2, 'a'), (3, 'b')]
+    assert L2 != L
```
ordirules: please be more explicit in the assert, write out what is expected, will help save mental energy in the long run
I assume it's `L2 == [(2, 'a'), (3, 'b')]`

CJ-Wright:
Although I am more than happy to write whatever the consensus is, I'd personally rather have a whole bunch of nodes which do things that may overlap a little rather than have nodes which have many flags. This way the user only has to remember more nodes rather than having to remember more flags.
I'd like to get this in today if possible.

ordirules:
We discussed about it off PR. Merging `combine_latest` with `zip_latest` is not a good idea, since it would force adding a buffer to `combine_latest`. I agree with the current PR and see its value.

However, we might consider discussing more complex scenarios for this stream before discussing merging. Let's say we have 3 streams:
```
Stream 1 : ----1----------2---------------------3-
Stream 2 : --------I--II-----III-------------------
Stream 3 : ---------------------------a----b------
```

What should be the output of this? Should it be:
```
(2, III, a), (2, III, b), (3 III, b)
```

or should it be recursive, like:
```
(2, I, a), (2, II, b), (2, III, a), (2, III, b), (3, III, b)
```

I would be tempted to say the former (which as far as I can tell is what the code is currently doing), since the latter could be done with (I think...):
```python
s4 = s1.zip_latest(s2)
s4.zip_latest(s3)
```

So my vote is we need to write two tests that output the two scenarios I mentioned, then I'm okay with it.
So something like:
```python
s1  = Stream()
s2 = Stream()
s3 = Stream()
s_simple = s1.zip_latest(s2, s3)
s_recursive = s3.zip_latest(s1.zip_latest(s2)) 

L_simple = s_simple.sink_to_list()

L_recursive = s_recursive.sink_to_list()

s1.emit(1)
s2.emit('I')
s2.emit("II")
s1.emit(2)
s2.emit("III")
s3.emit('a')
s3.emit('b')
s1.emit(3)
```
 Any other thoughts? Note, I haven't tested this, and output may not be correct but this is the idea.

CJ-Wright:
I'm confused, which is the lossless stream in your example?

ordirules:
I updated it to clarify. The output was incorrect. Also shortened. Please read above.

streamz/core.py
```diff
@@ -330,6 +330,14 @@ def zip(self, *other):
         """ Combine two streams together into a stream of tuples """
         return zip(self, *other)
 
+    def zip_latest(self, *others):
+        """Combine multiple streams together to a stream of tuples
+
+        This will emit a new tuple of the elements from the lossless stream paired
```
ordirules: define which is lossless

streamz/tests/test_core.py
```diff
@@ -494,3 +494,39 @@ def test_partition_str():
 def test_stream_name_str():
     source = Stream(name='this is not a stream')
     assert str(source) == '<this is not a stream; Stream>'
+
+
+def test_zip_latest():
```
ordirules: add test for 3 streams

ordirules:
After discussion with @CJ-Wright, to clarify, this routine makes an assumption. Only one stream is considered to be the lossless stream, and thus buffered. In `s1.zip_latest(s2, s3)`, the lossless stream is always `s1`. This is like `combine_latest` with `emit_on=[s1]` and an additional lossless buffer for `s1`. Once all streams have emitted at least once, the behaviour is exactly the same as `s1.combine_latest(s2, s3, emit_on=[s1])`.

The case of multiple lossless buffers is not considered, but @CJ-Wright will post an issue. It will require recursion through the buffers.

CJ-Wright:
I'm going to merge this, everything passes and such. If we find the name super distasteful we can always change it again.

Issue #35, opened 2017-07-13
Release?


CJ-Wright:
Would it be possible to send out an alpha release?
I don't mind conda-forge packaging.

mrocklin:
This also probably depends on names

CJ-Wright:
Fair

CJ-Wright:
It seems (at least from #36) that `dask-streams` is the working title for this work. Would it be possible to do all the name changing in the near future? I'm happy to PR this in, but I don't want to break everyone's PRs. I'm hoping that we can package/ship an alpha/beta version of this by Sept 5th?

ordirules:
I'm fine with breaking all my PR's for name change. We can probably vote between us since it's only @mrocklin @danielballan you and I.

danielballan:
👍

CJ-Wright:
I think that is buy-in from all the major contributors, so `dask-streams` it is! @mrocklin are there any major changes (aside from the name change) associated with attaching `dask` to the name (eg licence change)?

mrocklin:
Sorry for my absence recently.  I'll try to merge/release/push docs
sometime today.

CJ-Wright:
Are there other barriers to release? (aside from tagging it)

mrocklin:
I can do this any time.  We might want basic API docs first?  This doesn't
really block though.

CJ-Wright:
I'd prefer sooner rather than later (I'm going to use this in a release of software next week). When we have a tag let me know I'll put it up on conda-forge. Are we still tied to distributed master?

mrocklin:
I've added docs, screwed around with the README, released, and pushed to PyPI here: https://pypi.python.org/pypi/streamz/0.0.1

mrocklin:
If you send a PyPI username I'll add you to the list

CJ-Wright:
Ok, will do also I will put up a conda-forge recipie

CJ-Wright:
(Will close when conda-forge is up)

PR #52, opened 2017-08-15
REL: rename the library

### LINKED ISSUE #3: Rename library
The name "streams" is overly generic.  It is also taken on PyPI.

We should find a better name.  One option would be to use the dask namespace and call this `dask-streams` although this makes less sense if the vanilla stream implementation is to be a core feature.

CJ-Wright:
The one PR to break them all.
[Cue foreboding music](https://youtu.be/l5pnoSgIuVo?t=15s)
@ordirules 
In prep for #35 
Closes #3


mrocklin:
Yeah, I was thinking about this last night and was having second thoughts about `dask_streams`, mostly because no one seems to be using the dask bits yet.  Best I came up with was `streamz`.

CJ-Wright:
I'm ok with that too.
@ordirules @danielballan

ordirules:
great!
`streamz` might make more sense. I'm fine with either.

streamz/core.py
```diff
@@ -9,7 +9,7 @@
 from tornado.ioloop import IOLoop
 from tornado.queues import Queue
 from collections import Iterable
-from streams.graph import visualize
+from streamz.graph import visualize
```
mrocklin: Can we make this a relative import?

    from .graph import visualize

streamz/tests/test_core.py
```diff
@@ -9,11 +9,11 @@
 from tornado.queues import Queue
 from tornado.ioloop import IOLoop
 
-import streams as s
+import streamz as s
```
mrocklin: Perhaps start the two-letter-import-convention of `sz`?

danielballan:
Hahahaha. I love streamz.

CJ-Wright:
@mrocklin any more comments?

mrocklin:
I've done a few more name changes.

CJ-Wright:
Cool, can you merge when you are ready?

CJ-Wright:
Awesome, thank you!

mrocklin:
Docs are up: http://streamz.readthedocs.io/en/latest

mrocklin:
To add to docs you just need to add to the docs/source directory.  readthedocs.org will build and publish on every push to master.  Looking at projects like dask or toolz might give good examples of docs with rst if you're not already familiar.

CJ-Wright:
I like the docs, their like the library, minimal :laughing:

ordirules:
great :-)

PR #53, opened 2017-08-15
Defend against missing dask library

mrocklin:


ordirules:
this is much appreciated, thanks!
(we have a server that we probably don't care about installing `graphviz`)

PR #55, opened 2017-08-17
MTN: name -> stream_name

CJ-Wright:
`name` is a fairly common name, so rename to `stream_name` so no conflicts with map

### DIFF
--- a/examples/plot_graph.py
+++ b/examples/plot_graph.py
@@ -1,9 +1,9 @@
 from streamz import Stream
 from operator import add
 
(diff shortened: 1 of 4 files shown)


ordirules:
Yes, it would be great if this change could be made, since at @CJ-Wright mentions, `s.map(func, name="something")` is a disallowed `kwarg` to pass to `func`.

ordirules:
this looks good, what's blocking it?

CJ-Wright:
Sorry! Pulling in now.

Issue #59, opened 2017-08-30
ENH: colorful graphs


CJ-Wright:
Is it possible to color each node in graphvis? If so we may want to come up with a color scheme such that every node type has a dedicated color for ease of viewing (and making nice graphs for presentations/publications).

mrocklin:
It is.  I've often found this page to be a helpful reference for the DOT
language: http://www.graphviz.org/doc/info/attrs.html

CJ-Wright:
Do we just stuff color information in the graphviz nodes somewhere and expect graphviz to know what to do?

mrocklin:
Yes.  It looks like we're using the graphviz library.  A quick search
yields this documentation:
https://graphviz.readthedocs.io/en/stable/manual.html#attributes

mrocklin:
You might also find constructing some DOT graphs by hand to be educational
if you haven't done it already.  It's straightforward and a useful tool for
publication.

ordirules:
sounds good. tried it and seems simple (use `style=filled` with `color=clr`). Looks like we can use much more, such as shape.

Did you have any coloring in mind?
I think differentiating stream methods (`map`, `filter` etc) as well as stream subclasses (for example from `Stream` to `DaskStream`) could be nice. Maybe it could be something simple such as color for the methods and shape for the subclasses. Since 'method' might be a little ambiguous, here's an example:
```python
from streamz import Stream
s = Stream()
s2 = s.map(lambda x :x + 1)
from streamz import DaskStream
# now objects are Futures
s3 = DaskStream.scatter(s2)
# still a Future
s4 = s3.map(lambda x : x**2)
s5 = DaskStream.gather(s4)
```
Shape would be extremely useful. For example, running `filter ` on a `DaskStream` would yield a `Stream` (which is unlikely to be intended use). This could be useful for debugging.

What do you think?

CJ-Wright:
I am good with all of those. I'd like to make these (private) attributes of the classes, this way classes which inherent from them/override them would be easy to use. (eg we'd just need to change the shape once for dask streams if they all inherited)
I don't have any particular coloring in mind, we may consider drawing from an MPL color cycle since those are more likely to be colorblind safe.

PR #64, opened 2017-09-25
Stream name

jrmlhermitte:
we're not passing the names properly. this is just a placeholder. Probably we should define a list of stream only strings and pass these through. One issue is anyone subclassing `__init__` has to be aware of this.
If someone has a better idea let me know. I can look into this later.
This is will affect #59 .

ex:
```python
s = Stream()
s2 = s.map(lambda x : x + 1, stream_name="Foo")
# etc...
```

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -68,7 +68,7 @@ class Stream(object):
             self.children = children
         else:
             self.children = [child]
-        if self.loop is not None:
+        if loop is not None:
             self._loop = loop
         for child in self.children:
             if child:
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -195,9 +195,9 @@ class Stream(object):
             child.parents.remove(self)
             self.children.remove(child)
 
-    def scatter(self):
+    def scatter(self, **kwargs):
         from .dask import scatter
-        return scatter(self)
+        return scatter(self, **kwargs)
 
     def remove(self, predicate):
         """ Only pass through elements for which the predicate returns False """
@@ -211,7 +211,7 @@ class Stream(object):
     def concat(self):
         return self.flatten
 
-    def sink_to_list(self):
+    def sink_to_list(self, **kwargs):
         """ Append all elements of a stream to a list as they come in
 
         Examples
@@ -227,12 +227,12 @@ class Stream(object):
         self.sink(L.append)
         return L
 
-    def frequencies(self):
+    def frequencies(self, **kwargs):
         """ Count occurrences of elements """
         def update_frequencies(last, x):
             return toolz.assoc(last, x, last.get(x, 0) + 1)
 
-        return self.scan(update_frequencies, start={})
+        return self.scan(update_frequencies, start={}, **kwargs)
 
     def visualize(self, filename='mystream.png', **kwargs):
         """Render the computation of this object's task graph using graphviz.
@@ -336,10 +336,10 @@ class sink(Stream):
     """
     _graphviz_shape = 'trapezium'
 
-    def __init__(self, child, func):
+    def __init__(self, child, func, **kwargs):
         self.func = func
 
-        Stream.__init__(self, child)
+        Stream.__init__(self, child, **kwargs)
         _global_sinks.add(self)
 
     def update(self, x, who=None):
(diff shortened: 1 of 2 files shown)


CJ-Wright:
Would it be possible to add a line or two to each test to check the naming went though?

streamz/core.py
```diff
@@ -457,22 +459,27 @@ def update(self, x, who=None):
...
 
 class filter(Stream):
-    def __init__(self, predicate, child):
+    def __init__(self, predicate, child, **kwargs):
         self.predicate = predicate
+        # this is one of a few stream specific kwargs
+        stream_name = kwargs.pop('stream_name', None)
 
-        Stream.__init__(self, child)
+        Stream.__init__(self, child, stream_name=stream_name)
```
mrocklin: We should be able to just pass through kwargs directly to `Stream` without unpacking the `stream_name` keyword

mrocklin: This will also help us to err properly when someone passes an unexpected keyword

mrocklin: (which would be a good case to test for)

CJ-Wright:
I think this will be added to the visualization/string repr automatically (since we defined `__str__` in the base class to take the `self.name`)

ordirules:
Just to make sure we're on the same page @mrocklin , this is meant to resolve the issue where we want to add an extra name to the visualized streams, like here:
https://imgur.com/a/O5KGS

Note there was a mistake in my code, I will try to elaborate in the comment you mentioned.

streamz/core.py
ordirules: I didn't understand exactly. First, I made a mistake. Here we should have:
```
    def filter(self, predicate, **kwargs):
        """ Only pass through elements that satisfy the predicate """
        return filter(predicate, self, **kwargs)
```
(the method should pass kwargs, and every other method as well. I will change this soon)

With that correction, did you mean that kwargs should be passed directly to `Stream` for every method or some methods? Methods like `map` can receive arbitrary kwargs. I don't think we can do that. Please let me know if I misunderstood. thnaks!

mrocklin: I agree that with `map` or `scan` we will need to handle this differently.  In most cases though we can pass through kwargs explicitly.  We want to avoid leaving kwargs dangling.  The following should hold.

```python
def test_bad_kwarg_raises():
    source = Stream()
    with pytest.raises(TypeError):
        source.filter(lambda x: x > 1, bad_kwarg=True)
```

jrmlhermitte:
Added tests and also removed `**kwargs` in the `Stream.__init__()`. This still needs to be added to the subclasses (`dask` etc). I'm not sure how I feel about this in general. Maybe sit on it and see if we can think of a cleaner extensible way?

streamz/tests/test_core.py
```diff
@@ -65,7 +71,10 @@ def f(acc, x, y=None):
 
 def test_filter():
     source = Stream()
-    L = source.filter(lambda x: x % 2 == 0).sink_to_list()
+    sout = source.filter(lambda x: x % 2 == 0, stream_name="filtering")
+    L = sout.sink_to_list()
+
+    assert sout.name == "filtering"
```
mrocklin: It might be nice to avoid having these checks in all of the normal tests.  I understand the objective to test that all of the different Stream subclasses pass through kwargs.  Maybe this can be tested in a single isolated test rather than in all of the tests?

jrmlhermitte: I agree, I think it is more organized, and less painful to go back and read. Changed it. thanks!

mrocklin:
Merging.  Thanks @jrmlhermitte !

streamz/core.py
```diff
@@ -647,7 +652,8 @@ def __init__(self, *children, **kwargs):
 
         children2 = [child for child in children if isinstance(child, Stream)]
 
-        Stream.__init__(self, children=children2)
+        Stream.__init__(self, children=children2, stream_name=stream_name,
+                        **kwargs)
```
jrmlhermitte: Sorry this is a mistake (I must have added the wrong thing when rebasing and resolving conflicts)
It should be:
```
Stream.__init__(self, children=children2, **kwargs)
```
and `stream_name = kwargs.pop('stream_name', None)` should be removed. It's on my branch [https://github.com/jrmlhermitte/streamz/tree/quickfix-streamname](quickfix-streamname) if you want to take a look.

jrmlhermitte:
Thanks! I made a small typo. See line note. Doesn't affect much but the suggested change will be less forgiving if a bad kwarg is supplied (which is what we want).

jrmlhermitte:
(finally see #92, also added a test)

Issue #57, opened 2017-08-21
ENH: inverse of zip (split?)


CJ-Wright:
We may want the inverse of zip, which takes in a stream of tuples and then either: a) breaks them up  into separate streams or b) selects one "column" of the tuples to report. I think a) might be more general.

mrocklin:
Do you actually need this thing or can you get by with pluck?

The experience from pytoolz was that people like to add things even if
those things are unlikely to see active use.  Minimizing the number of
things out there is helpful to reduce maintenance costs.

CJ-Wright:
I am not familiar with pluck, what is it?

ordirules:
part of `toolz` lib:
http://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.pluck

you can map 
```python
s = Stream()
s_first = s.map(partial(pluck, [0]))
s_second = s.map(partial(pluck, [1]))
s.emit([1,2])
# ...
```
(I knew of pluck but not use case, wondering if this is the line of thinking)

CJ-Wright:
Cool. I think this would work.

mrocklin:
Sorry, I assumed that pluck was already part of the API.  It seems like a decently common use case and so possibly merits being added as syntactic sugar.

avolkov1:
I don't understand the API. I don't know ahead of time how many streams might have been zipped. I need to take these zipped streams and zip more streams, but zipping zipped streams doesn't work for what I'm trying to do. I want something like:
```
streams_uzip = streamz.unzip(streams_zip)
streams_zip_out = streamz.zip(*streams_uzip, another_stream)
return streams_zip_out
```

CJ-Wright:
Why not provide the constituent streams to create `streamz_zip_out`? At some point they were all independent. You could also write a function that unpacks the zipped stream so that all the elements are at the same level in the hierarchy. If that was done right after the zipping you could be confident in how many levels you need to go down.

avolkov1:
I don't have the constituent streams. It's a dynamic framework. I know that some zipped streams are passed into a function. I don't know how many have been zipped.

Yes, if I need to delve into the API to understand how to unpack the zipped stream. That's what an unzip would be.  Can you give an example of this function?

```
def unzip(zipped_stream):
    if not is_zipped_stream(zipped_stream):
        raise Exception('Not a zipped stream')
    . 
    .
    .
   return streams_zip_out
```

[...]

Issue #61, opened 2017-09-05
Merging Hygiene


mrocklin:
How do we feel about git hygiene when merging?  I personally tend to squash-and-merge PRs, except if git history is particularly clean (this happens very rarely for me).  Any thoughts?

When people maintain larger projects it's usually nice to be able to assume a few things about git history

1.  It is entirely linear, without branching
2.  All tests pass on all commits

These conditions are very rarely held during normal active development of any feature branch, but are typically held when a branch is merged.  We can use github's squash-and-merge feature to automatically squash all commits in a PR to one.  This loses some history though, so there is a balance here if the git history of that branch is particularly valuable.

Most pydata projects that I know of use squash-and-merge by default.

CJ-Wright:
squash-and-merge works for me. I personally am often guilty of committing breaking code usually as a way to keep track of where I'm at in the middle of a refactor or so I can have other people look at the code, so I'm usually less strict about it, but I could try to be cleaner about that.  I guess with squash-and-merge we are guaranteed that all tests pass so long as travis passes on the last commit so I am ok with that.

mrocklin:
Committing breaking or WIP code makes lots of sense.  I do it all the time. Squash-and-merge allows this kind of workflow during a PR without having it bleed into the final history.

CJ-Wright:
Do we want a .github/contributing.md to go along with this? (Can we just lift one off of another pydata project?)

mrocklin:
I see that as separate.  That is information that a newcomer should be
aware of like write tests, check style, document functions, etc..  This
issue is about how people with commit access should manage things.

CJ-Wright:
That's fair (maybe github should have an advanced contributing guide). Do you want to enshrine this information somewhere? If so, where?

mrocklin:
In dask I tend to just informally tell individuals when they get commit
rights.  Happy with whatever

ordirules:
I am fine with whichever. Personally, I do like micro commits as they could be useful for code forensics. But the PR's can also serve this (assuming that we can still see comments on outdated diffs after the history has been modified?).

ordirules:
Perhaps we should have a suggestion message on the commits, with rebase instructions, such as:

To squash your commits, please type:
`git rebase -i HEAD~N`
[...]

PR #66, opened 2017-09-28
ENH: Added pluck syntactic sugar

ordirules:
* Pluck allows picking certain elements from the stream.
      It assumes the stream is a list of elements

Getting ball rolling for #57. Feel free to hijack if I don't respond quick enough.

Was this intended use?
@mrocklin @CJ-Wright

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -319,9 +319,16 @@ class Stream(object):
         >>> source.pluck([0, 3]).sink(print)
         >>> for x in [[1, 2, 3, 4], [4, 5, 6, 7], [8, 9, 10, 11]]:
         ...     source.emit(x)
-        (1,4)
-        (4,7)
-        (8,11)
+        (1, 4)
+        (4, 7)
+        (8, 11)
+
+        >>> source = Stream()
+        >>> source.pluck('name').sink(print)
+        >>> for x in [{'name': 'Alice', 'x': 123}, {'name': 'Bob', 'x': 456}]:
+        ...     source.emit(x)
+        'Alice'
+        'Bob'
         """
         return pluck(self, pick)
 
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -311,6 +311,7 @@ class Stream(object):
         ----------
         pluck : object, list
             The element(s) to pick from the incoming element in the stream
+            If an instance of list, will pick multiple elements.
 
         Examples
         --------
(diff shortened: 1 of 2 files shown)


streamz/core.py
```diff
@@ -675,6 +690,20 @@ def update(self, x, who=None):
...
+        self.pick = pick
+        super(pluck, self).__init__(child)
+
+    def update(self, x, who=None):
+        L = []
+        # Note will raise IndexError if x is too small
+        for elem in self.pick:
+            L.append(x[elem])
+
+        return self.emit(L)
```
mrocklin: I think that consistency with toolz would be something like the following:

```python
def update(self, x, who=none):
    if isinstance(self.pick, list):
        return self.emit(tuple([x[ind] for ind in self.pick]))
    else:
        return self.emit(x[self.pick])
```

ordirules: ok will add it (ignore the commit i just sent, was done before reading this)

ordirules:
This is a good opportunity to work out #61. I squashed my commits.

streamz/core.py
```diff
@@ -304,6 +304,27 @@ def union(self, *others):
         """
         return union(children=(self,) + others)
 
+    def pluck(self, pick):
+        """ Select elements from a tuple in the stream.
```
mrocklin: I recommend replacing "from a tuple" with "from elements" .  This will work with any indexable, including dicts.

streamz/core.py
```diff
@@ -304,6 +304,27 @@ def union(self, *others):
...
+
+        Parameters
+        ----------
+        pluck : list
+            The elements to pick from the incoming element in the stream
+            (note: must be a list)
```
mrocklin: This also works with non-list elements.  Perhaps `list` -> `object, list` and "The elements" -> "The element(s)"

streamz/core.py
```diff
@@ -304,6 +304,27 @@ def union(self, *others):
...
+
+        Examples
+        --------
+        >>> source = Stream()
+        >>> source.pluck([0,3]).sink(print)
```
mrocklin: Style nit, space after comma.

streamz/core.py
ordirules: sounds good. I changed the "note" to:

`If an instance of list, will pick multiple elements.`

ordirules:
thanks! (I wish github had a nicer way of showing the final commit without having to look at the actual code history)

PR #70, opened 2017-09-29
Add disconnect operation

mrocklin:


streamz/core.py
```diff
@@ -110,10 +110,10 @@ def update(self, x, who=None):
     def connect(self, parent):
         ''' Connect this stream to a downstream element.
 
-            Parameters
-            ----------
-            parent: Stream
-                the parent stream (downstream element) to connect to
+        Parameters
```
ordirules: thanks for the docstring fix!

ordirules:
Just tested with visualize just in case... Seems to work fine
```python
In [1]: from streamz import Stream

In [2]: s = Stream()

In [3]: sout = s.map(lambda x : x+1)

In [4]: s.map(print)
Out[4]: <streamz.core.map at 0x7f12823259e8>

In [5]: sout.map(print)
Out[5]: <streamz.core.map at 0x7f1282323dd8>

In [6]: s.disconnect(sout)

In [7]: s.visualize("sin.png")
Out[7]: <IPython.core.display.Image object>

In [8]: sout.visualize("sout.png")
Out[8]: <IPython.core.display.Image object>
```
(sin.png and sout.png top and bottom here: https://imgur.com/a/OHgsz)

PR #72, opened 2017-09-29
Garbage collect streams

### LINKED ISSUE #71: Garbage collect streams
Currently streams contain reference cycles between parent and child.  As a result, they never get cleaned up.  Generally this is ok, they're pretty small in memory.  It does make it difficult to stop operating on a stream though.

    source.map(print)  # this stream will print forever

This behavior can be both convenient and limiting.  

We might instead choose to require that a reference to a stream be kept explicitly in order to keep it around.

    printer = source.map(print)  # this stream will print forever
 
Then if we delete all references to the printer, the stream will be cleaned up and printing will stop

    del printer

This would be a change in semantics.  Do we want this change?

mrocklin:
We replace parents with a weakref.WeakSet so that streams only stick
around if there is an active reference to them.

The sink method is more durable though.  Sinks place themselves into a
global set so that their common use for side effects remains valid.

Fixes #71

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 from collections import deque
 from time import time
+import weakref
 
 import toolz
 from tornado import gen
@@ -11,6 +12,8 @@ from tornado.queues import Queue
 from collections import Iterable
 
 no_default = '--no-default--'
+
+_global_sinks = set()
 
 
 def identity(x):
@@ -52,7 +55,7 @@ class Stream(object):
     str_list = ['func', 'predicate', 'n', 'interval']
 
     def __init__(self, child=None, children=None, stream_name=None, **kwargs):
-        self.parents = []
+        self.parents = weakref.WeakSet()
         if children is not None:
             self.children = children
         else:
@@ -61,7 +64,7 @@ class Stream(object):
             self._loop = kwargs.get('loop')
         for child in self.children:
             if child:
-                child.parents.append(self)
+                child.parents.add(self)
         self.name = stream_name
 
     def __str__(self):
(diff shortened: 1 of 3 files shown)


ordirules:
I'm definitely noticing some issues after this commit. Some streams don't propagate. I am guessing it has something to do with `connect` or `disconnect`. I will look into it more later when I have the time, hopefully this evening or tomorrow. @CJ-Wright  may also have an idea.

CJ-Wright:
I haven't really started playing with this yet, so I don't have ideas on this yet.

ordirules:
Got it.
```python
from streamz import Stream
s = Stream()
s2 = s.map(lambda x : "Got here")
s.map(lambda x : "Got here 2")
s.emit(1)
```
will only output "Got Here"

However, if you run the same thing in an ipython terminal, it will output both "Got Here" and "Got Here2".
The reason is that ipython saves a reference to the returns of variables....

So the problem here is that if the returned reference is not kept, then the stream is dropped. I would disagree with this behaviour. Especially since it can lead to inconsistent results when running or not in ipython. Should i move this to an issue? What do you think @mrocklin ? Thanks!

mrocklin:
Yes, sorry there wasn't a longer time to comment on this.  This was exactly the intention of this PR.  

Your code will probably work fine if you use `sink` rather than `map`.  Sink does exactly the same thing, but also registers the output sink with a global set so that it isn't cleaned up.  Sink is intended for use with side effects.

ordirules:
Thanks. I'll do that from now on, that actually is a good compromise. I think this should definitely be something added to the documentation. I posted as issue #77 if you don't mind. I would like to help contribute to documentation. Feel free to close issue if you find it too out of scope. Thanks!

PR #74, opened 2017-10-02
Register methods with Stream.register_api decorator

mrocklin:
This moves methods off of the main stream class and offers an opportunity for
extension without subclassing.

People want to build their own operations for streams. The obvious way to do
this, subclassing, is hard due to how we have laid things out.  As an
alternative we consider function dispatch.  This also helps to reduce
boiler-plate, although it also adds some complexity and indirection.

Currently we define operations like `map` as a separate class and then add a
special method:

```python
class Stream(object):
    ...
    def map(...):
        return map(...

class map(Stream):
    ...
```

This is bad for two reasons:

1.  The map method is somewhat redundant
2.  We have hard coded all valid operations.  It's hard to add new ones.

As an alternative we dynamically add methods to stream with a decorator

```python
class Stream(object):
    @classmethod
    def register_api(cls, func):
        ...

@Stream.register_api
class map(Stream):
    ...
```

cc @CJ-Wright, @ordirules and @limx0 for feedback.

cc @eriknw for help with wrapping effectively if he has the time.  Currently we lose function names and docstrings (see failing `test_docstrings` test).

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
 from collections import deque
+import functools
 from time import time
 import weakref
 
@@ -71,6 +72,14 @@ class Stream(object):
             if child:
                 child.parents.add(self)
         self.name = stream_name
+
+    @classmethod
+    def register_api(cls, func):
+        @functools.wraps(func)
+        def _(*args, **kwargs):
+            return func(*args, **kwargs)
+        setattr(cls, func.__name__, _)
+        return func
 
     def __str__(self):
         s_list = []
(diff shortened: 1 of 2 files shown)


mrocklin:
To be clear I'm not sure that we should merge this.  The cost of complexity of this change is currently non-trivial.  When something goes wrong it will be somewhat difficult for a user or developer to trace back to the source of the problem (no docstrings, partialed/wrapped functions, etc..)  I think that the challenge here is to see how cleanly we can make `my_stream.foo` map back to the appropriate `foo` class.

CJ-Wright:
To be fair our tracebacks are really difficult to read anyway, since the traceback goes through all the emits and updates in the graph.

ordirules:
I really like it, but one thing bothers me, the following:
```python
from streamz import Stream

class NewStream(Stream):
    pass

@NewStream.register_api
class foo(Stream):
    """ Apply a function to every element in the stream """
    def __init__(self, child, *args, **kwargs):
        self.kwargs = kwargs
        self.args = args

        Stream.__init__(self, child)

        print("foo")

        return self.emit(x)


sf = NewStream()
sout = sf.foo()


s = Stream()
s2 = s.foo()
```

After registering `foo` to the `NewStream`, it's also now available for `Stream`. Is there a way around this?
We should be able to cleanly subclass. (Correct me if I did something wrong) But yes, this is definitely something that would be useful :-).

mrocklin:
Yeah, that sounds like a valid concern.  Looks like this is the problem:

```python
In [4]: NewStream._api is Stream._api
Out[4]: True
```

limx0:
Yep I like this, it suits my requirements for streamz.

mrocklin:
So you could resolve this by redefining `_api`

```python
class NewStream(object):
    _api = {}
```

mrocklin:
@limx0 are you actively using `streamz`?  If so, how?  If not, why not?

limx0:
I am playing around with it in a development branch of a trading system and looking to merge it sometime soon.

ordirules:
It sounds good to me. I just have one personal preference that would disagree with this implementation. I see the streaming nodes as self contained nodes in a tree. The functions it may call in my opinion should be more a local state than a global. So my preference is that the methods the stream knows it can access or not should be created at instance creation time, not class creation time.

What I mean is this, I would prefer this to be possible (or something similar):
```python
s = Stream()
s2 = NewStream(s)
# s3 is a NewStream but map is made inaccessible
s3 = NewStream(s, remove=[map])
# and then maybe some ease of inspection of the instance
print(s2.methods())
print(s3.methods()) 
```
or something of the sort. In the current scheme, I'd have to hard code a `NewStream` and `NewStream2` (where latter acts on `s2` and doesn't contain map) to get this done.

I'm definitely fine with this implementation though. But I'm trying to play devil's advocate as far as possible to see how much we like this.

mrocklin:
In general methods are usually handled at the class level and not the instance level.  You can always attach any attribute you like to a Python object, but I don't think that this is something that streamz should support on its own.

I'm now just going forward with vanilla dynamic functions.   I'm also using the namespace of the classes.  What's here seems simple and effective to me.  I plan to merge on passed tests if there are no major concerns.

streamz/core.py
```diff
@@ -72,6 +73,14 @@ def __init__(self, child=None, children=None, stream_name=None, **kwargs):
                 child.parents.add(self)
         self.name = stream_name
 
+    @classmethod
+    def register_api(cls, func):
+        @functools.wraps(func)
+        def _(*args, **kwargs):
+            return func(*args, **kwargs)
+        setattr(cls, func.__name__, _)
```
ordirules: Just out of mere curiosity (I can't see the subtle reason as of now), why this step and not just use `func`?

ordirules:
it looks good to me (if tests pass). thanks again, this will be very useful

streamz/core.py
mrocklin: Python didn't like it.  I guess it didn't identify it as a method and so didn't pass it `self`.  This gets into levels of Python that I never cared to learn about.

ordirules: thanks!

streamz/core.py
```diff
@@ -546,8 +447,10 @@ def update(self, x, who=None):
             return []
 
 
+@Stream.register_api
 class map(Stream):
-    def __init__(self, func, child, args=(), **kwargs):
+    """ Apply a function to every element in the stream """
+    def __init__(self, child, func, *args, **kwargs):
```
ordirules: This is an API break. Should we tag it as such somewhere or mention in commit? (same for `filter` etc)

mrocklin: If the project was further along, yes.  However at this early stage I plan to move quickly and break things.  I apologize if this causes you all some stress as early adopters.  You may be better served by relying on a previously released version for a while.

I apologize for my lack of rigor here.  I am fairly time constrained.

ordirules: For me it's no problem (worst case I can keep a fork). I just thought I would point it out. I wasn't sure which stage of development `streamz` is in at this point since it seems to work fairly well (and we're using it and have deployed an initial prototype).

ordirules: (and thanks for this great library!)

CJ-Wright: We may want to release the library for a `0.2.0` backdated before this merge? That way we have something before the break.

mrocklin:
No strong thoughts.  I don't object to this.  I'm unlikely to do it myself.

PR #75, opened 2017-10-02
Graph and string styling

mrocklin:
Modify visual styling for graph and string repr
    
We add graphviz attributes to the Stream classes directly (I'm not sure how I feel about this generally, but for now it seems productive).

```python   
class Foo(Stream):
    _graphviz_shape = 'box'
```

We remove the name from string reprs
    
    <map; func=inc>  # before
    <map: inc>       # after

### Example

```python
from streamz import Stream

def inc(x):
    return x + 1

def dec(x):
    return x - 1

import operator

source = Stream.from_textfile('foo.txt')
a = source.map(inc)
b = source.map(dec)
c = a.zip(b).scan(operator.add).sink(print)
c

c.visualize(rankdir='LR')
```

image

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -491,8 +491,8 @@ class Stream(object):
         from .sequence import StreamingSequence
         return StreamingSequence(stream=self, **kwargs)
 
-    @gen.coroutine
-    def from_textfile(self, f, poll_interval=None):
+    @staticmethod
+    def from_textfile(f, poll_interval=None):
         """ Read data from file into stream
 
         Parameters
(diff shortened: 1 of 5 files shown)


streamz/core.py
```diff
@@ -708,6 +713,9 @@ def cb(self):
 
 
 class zip(Stream):
+    _graphviz_orientation = 270
```
CJ-Wright: Does this mean that things will always be left to right?

mrocklin: Left to right is controlled by the keyword `rankdir='LR'` rather than `rankdir='TD'`

This affects the orientation of the triangle shape in zip to be pointing to the right.

CJ-Wright: Right, I assume that the triangle is pointing to the right because the pointy part of the triangle is where the stream comes "out". But that preferences one graph orientation over others.

mrocklin:
Agreed.  I think that it looks decent in either layout.

CJ-Wright:
We could write a little thing in the graph visualizer which looks for `rankdir` and if not `LR` changes the zip orientation.

mrocklin:
This sounds somewhat complex.  I'd prefer to avoid adding special rules for
a while and wait until things settle.

CJ-Wright:
Fair enough.

Issue #80, opened 2017-10-06
visualize() issue


ordirules:
Logging the details here for now. This now raises an error:

```python
In [1]: from streamz import Stream

In [2]: s = Stream()

In [3]: s2 = s.map(lambda x : x + 1)

In [4]: s.visualize()
Error: Unknown HTML element <lambda> on line 1 
in label of node map; <lambda>
---------------------------------------------------------------------------
CalledProcessError                        Traceback (most recent call last)
```

but this works:
```python
In [1]: from streamz import Stream

In [2]: s =Stream()

In [3]: def inc(x):
   ...:     return x+1
   ...: 

In [4]: s2 =s.map(inc)

In [5]: s.visualize()
Out[5]: <IPython.core.display.Image object>

```
I did not have issues with it before.

I will try to get back to it later and add more details if I cannot figure it out. If anyone has a quick idea to save time that would be great. thanks!

ordirules:
while we're here, `visualize` doctstring has a typo (`dsk` should be `node`)

CJ-Wright:
I think this is related to #76.

ordirules:
oops should have read issues. closing and posting there.

Issue #76, opened 2017-10-04
Error in graph with lambda


CJ-Wright:
It seems that lambda functions now make the graph plotting rather unhappy.
```python
def test_create_file():
    source1 = Stream(stream_name='source1')
    source2 = Stream(stream_name='source2')

    n1 = source1.zip(source2)
    n2 = n1.map(add).scan(mul).map(lambda x: x + 1)
    n2.sink(source1.emit)

    with tmpfile(extension='png') as fn:
        visualize(n1, filename=fn)
        assert os.path.exists(fn)

    with tmpfile(extension='svg') as fn:
        n1.visualize(filename=fn, rankdir="LR")
        assert os.path.exists(fn)

    with tmpfile(extension='dot') as fn:
        n1.visualize(filename=fn, rankdir="LR")
        with open(fn) as f:
            text = f.read()

        for word in ['rankdir', 'source1', 'source2', 'zip', 'map', 'add',
                     'shape=box', 'shape=ellipse']:
            assert word in text
```
```python
/home/christopher/mc/envs/dp_dev/bin/python /home/christopher/pycharm-2016.3/helpers/pycharm/_jb_pytest_runner.py --target test_graph.py::test_create_file
Testing started at 12:43 PM ...
Launching py.test with arguments test_graph.py::test_create_file in /home/christopher/dev/streamz/streamz/tests

============================= test session starts ==============================
platform linux -- Python 3.5.4, pytest-3.2.2, py-1.4.34, pluggy-0.4.0
rootdir: /home/christopher/dev/streamz, inifile:
plugins: xonsh-0.5.12, env-0.6.0
collected 1 item
test_graph.py .Error: Unknown HTML element <lambda> on line 1 
in label of node map; <lambda>
Error: Unknown HTML element <lambda> on line 1 
in label of node map; <lambda>
Error: Unknown HTML element <lambda> on line 1 
in label of node map; <lambda>
```
I'm following up on this, I'm not certain how this broke from our previous versions.

CJ-Wright:
Also it seems that we could use a more robust assert than just checking the file's existence.

ordirules:
having same problem:
```python
In [1]: from streamz import Stream

In [2]: s = Stream()

In [3]: s2 = s.map(lambda x : x + 1)

In [4]: s.visualize()
Error: Unknown HTML element <lambda> on line 1 
in label of node map; <lambda>
---------------------------------------------------------------------------
CalledProcessError                        Traceback (most recent call last)
[...]

PR #81, opened 2017-10-06
Fixed docstrings, fixed parsing

ordirules:
proposed fix for #80. If this is good, maybe we should also add a test by testing visualize on the lambda function.

### DIFF
--- a/streamz/tests/test_graph.py
+++ b/streamz/tests/test_graph.py
@@ -29,7 +29,7 @@ def test_create_file():
     source2 = Stream(stream_name='source2')
 
     n1 = source1.zip(source2)
-    n2 = n1.map(add).scan(mul)
+    n2 = n1.map(add).scan(mul).map(lambda x : x + 1)
     n2.sink(source1.emit)
 
     with tmpfile(extension='png') as fn:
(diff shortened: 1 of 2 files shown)


ordirules:
related to #76

mrocklin:
Testing would be good if you're up for it. We might also consider
whitelisting rather than blacklisting.  For example we might only pass
through characters that are alphanumeric.

ordirules:
Actually, all that was needed in the test was @CJ-Wright 's suggestion in #76 so I will cherry pick it. Thanks @CJ-Wright.

I also added the white list and a test for the string parser.

ordirules:
Also @CJ-Wright said in #76:

> Also it seems that we could use a more robust assert than just checking the file's existence.

Any suggestions?

streamz/tests/test_graph.py
```diff
@@ -27,7 +29,7 @@ def test_create_file():
     source2 = Stream(stream_name='source2')
 
     n1 = source1.zip(source2)
-    n2 = n1.map(add).scan(mul)
+    n2 = n1.map(add).scan(mul).map(lambda x : x + 1)
```
ordirules: @CJ-Wright 's suggestion

mrocklin:
Great.  Merging.  Thanks @ordirules !

PR #91, opened 2017-10-12
Add filenames source

mrocklin:


streamz/core.py
```diff
@@ -76,12 +76,39 @@ def __init__(self, child=None, children=None, stream_name=None, **kwargs):
         self.name = stream_name
 
     @classmethod
-    def register_api(cls, func):
-        @functools.wraps(func)
-        def _(*args, **kwargs):
-            return func(*args, **kwargs)
-        setattr(cls, func.__name__, _)
-        return func
+    def register_api(cls, modifier=identity):
```
CJ-Wright: Does this mean that we can convert the main dask class to use `register_api`?

mrocklin: I think that this was possible before

PR #92, opened 2017-10-12
Also test on bad kwargs

jrmlhermitte:
Extension to #64. I realized `zip` was not correct. I made the change and also added a test that should test this. It could be good to have bad kwargs raise a `TypeError`

### DIFF
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -4,6 +4,7 @@ import json
 import operator
 from operator import add
 from time import time, sleep
+from functools import partial
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


streamz/tests/test_core.py
```diff
@@ -774,33 +775,82 @@ def test_destroy():
...
-               sin.buffer(2, stream_name=test_name),
-               sin.zip(sin2, stream_name=test_name),
-               sin.combine_latest(sin2, stream_name=test_name),
-               sin.frequencies(stream_name=test_name),
-               sin.flatten(stream_name=test_name),
-               sin.unique(stream_name=test_name),
-               sin.union(stream_name=test_name),
-               sin.pluck(0, stream_name=test_name),
-               sin.collect(stream_name=test_name),
+               partial(sin.map, lambda x : x),
```
jrmlhermitte: this is not completely necessary. my hopes of using partial was so i could reuse the list for the `bad_kwargs` test. however, I found I need to omit `map` and `scan` methods (or methods that call it) anyway.

streamz/tests/test_core.py
```diff
@@ -774,33 +775,82 @@ def test_destroy():
...
+               sin.collect,
                ]
 
+    good_kwargs = dict(stream_name=test_name)
+    bad_kwargs = dict(foo="bar")
     for s in streams:
-        assert s.name == test_name
+        # try good kwargs
+        stmp = s(**good_kwargs)
+        assert stmp.name == test_name
```
mrocklin: It looks like you use partial so that you can do this with both good and bad kwargs, but then you copy-paste the list for a different test anyway?  Presumably we would either merge the two tests and add the `pytest.raises` line within this for loop or we would avoid using partial.

I may not fully understand though.

jrmlhermitte: yeah, that's exactly what happened. i commented in line note about it as well. just thought of another way to go about it. this could work. see updated test

streamz/tests/test_core.py
```diff
@@ -771,36 +772,63 @@ def test_destroy():
...
+        del sout
+
+        with pytest.raises(TypeError):
+            sout = s(**bad_kwargs)
+            sin.emit(1)
+            # need a second emit for accumulate
+            sin.emit(1)
+            del sout
+
+    print(list(sin.parents))
```
mrocklin: remove print statement?

jrmlhermitte: oops yes, i was checking to make sure weakrefs were garbage collected. thanks

mrocklin:
Thanks @jrmlhermitte !

PR #93, opened 2017-10-12
WIP: kafka source

martindurant:
Simplistic implementation without attempt to batch or care for parallelism.

Needs tests, which probably should use the fixtures from pykafka.

### DIFF
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -2,6 +2,7 @@ from glob import glob
 import os
 
 import tornado.ioloop
+from tornado.ioloop import IOLoop
 from tornado import gen
 
 from .core import Stream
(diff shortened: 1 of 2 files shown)


streamz/sources.py
```diff
@@ -71,3 +72,42 @@ def start(self):
...
+    url: str
+        Connection string (host:port) by which to reach Kafka
+    group: str
+        Identity of the consumer. If multiple sources share the same group,
+        each message will be passed to only one of them.
+    poll_interval: number
+        Seconds that the thread sleeps between polling Kafka for new messages
+    """
+    def __init__(self, topics, url='localhost:9092', group='streamz',
+                 poll_interval=0.1):
```
mrocklin: I suggest that we just pass through the entire dictionary argument to `ck.Consumer`.  I'm not personally confident enough about Kafka that I would want to in any way control how users send keywords to the kafka client.

streamz/sources.py
```diff
@@ -71,3 +72,42 @@ def start(self):
...
+        super(Kafka, self).__init__()
+
+    @gen.coroutine
+    def poll_kafka(self):
+        while True:
+            msg = self.consumer.poll(0)
+            if msg is None or msg.error():
+                yield gen.sleep(self.sleep)
+            else:
+                self.emit(msg.value())
```
mrocklin: You'll want to yield on the emitted value here.  This is what enables back pressure (we'll stop reading data if any of the sinks can't keep up)

http://streamz.readthedocs.io/en/latest/core.html#processing-time-and-back-pressure

streamz/sources.py
```diff
@@ -71,3 +72,42 @@ def start(self):
...
+    ----------
+    topics: list of str
+        Labels of Kafka topics to consume from
+    url: str
+        Connection string (host:port) by which to reach Kafka
+    group: str
+        Identity of the consumer. If multiple sources share the same group,
+        each message will be passed to only one of them.
+    poll_interval: number
+        Seconds that the thread sleeps between polling Kafka for new messages
```
mrocklin: Slight nit, the thread doesn't sleep.  I would just say "seconds between polling kafka for new messages"

streamz/sources.py
```diff
@@ -71,3 +72,42 @@ def start(self):
                     yield last
                 else:
                     return
+
+
```
mrocklin: You can now use `@Stream.register_api(staticmethod)` here and ditch the `from_kafka` method on the Stream object (perhaps renaming this class to `from_kafka` instead)

streamz/sources.py
```diff
@@ -71,3 +72,42 @@ def start(self):
...
+    group: str
+        Identity of the consumer. If multiple sources share the same group,
+        each message will be passed to only one of them.
+    poll_interval: number
+        Seconds that the thread sleeps between polling Kafka for new messages
+    """
+    def __init__(self, topics, url='localhost:9092', group='streamz',
+                 poll_interval=0.1):
+        import confluent_kafka as ck
+        IOLoop.current().add_callback(self.poll_kafka)
```
martindurant: Given that this explicitly interacts with an IOLoop, is it a good idea to allow the user to pass in their instance?

mrocklin: Yes, I think so.  In another branch I'm doing `loop = loop or IOLoop.current()`

PR #97, opened 2017-10-14
Rename child to upstream and parent to downstream

mrocklin:
I believe that this naming scheme is more intuitive

This is a breaking change.

cc @CJ-Wright


CJ-Wright:
As much as a hate breaking API changes this one will make new user's lives easier.

Issue #95, opened 2017-10-14
Example of how to run event_loop in separate thread for using the `timed_window()` method


snth:
Hi,

Could you please add an example of how to run an event loop in a separate thread so that one can use the `timed_window()` and `delay()` methods?

The only reference I found is that it works in a Jupyter notebook because that already has an event loop running in the background. This doesn't help me as I'm not running in a Jupyter notebook.

I'm using `asyncio` in Python 3.6 and mostly that's working fine but I can't get the `timed_window()` function to work.

mrocklin:
Do you need something in a separate thread or can your entire application be asynchronous?  Presumably if you are already using `asyncio` then you should be somewhat comfortable with running asynchronous applications.  You could consider moving to the Tornado event loop, which should also support asyncio applications.

snth:
I'm not sure. I'm listening to various websockets and they're all emitting events and awaiting each other. Everything works great except if I add a `.timed_window()` to any stream then that stream locks up. I'm not sure what the problem is.

snth:
My code is in the [github.com/snth/numismatic]() package.

In particular in [https://github.com/snth/numismatic/blob/master/numismatic/exchanges.py]() on line 47 I would like to replace the `.partition(...)` with `.timed_window(interval=batch_size)`.

mrocklin:
It looks like you already have an event loop running with asyncio.  Here are the docs on using Asyncio event loops with tornado: http://www.tornadoweb.org/en/stable/asyncio.html

You can provide the `timed_window` constructor with a `loop=` keyword to give it the right event loop.

I agree with you that an example on how to use streamz with an asyncio application would be very valuable.  If I get some time I'll write something up but unfortunately I can't guarantee that this will happen in the very near future.  If you have the time I strongly encourage you to write up a small example and submit it.  I think that others like yourself might also find it valuable.

mrocklin:
Maybe @CJ-Wright or @jrmlhermitte has time here?

snth:
Sure, if I can get it to work I'll submit an example.

CJ-Wright:
I don't have a ton of experience with async programming, but I can try look at an example.

jrmlhermitte:
Same but interested. I think it's a great opportunity start more discussion on this now.

snth:
I've created an example. Please see if it works for you too.

mrocklin:
This was great.  Closing

PR #98, opened 2017-10-15
Example of using streamz with asyncio to use the timed_window() method #95

snth:
This actually turned out to be pretty easy.

The main thing to watch out for that I noticed is that the 

    AsyncIOMainLoop().install()

call has to occurr before the Stream() is created.

### DIFF
--- a/examples/fib_asyncio.py
+++ b/examples/fib_asyncio.py
@@ -1,0 +1,27 @@
+from streamz import Stream
+import asyncio
+from tornado.platform.asyncio import AsyncIOMainLoop
+AsyncIOMainLoop().install()
+
+
+source = Stream()
+s = source.sliding_window(2).map(sum)
+L = s.sink_to_list()                    # store result in a list
+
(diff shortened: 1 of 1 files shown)


mrocklin:
This looks great.  Thank you for putting this together @snth .  Merging

mrocklin:
Were you able to get your system up and running then?

snth:
@mrocklin Yes, it works great now! 

I didn't get to say this earlier but thank you for creating `streamz`. I'm loving it so far!

I think `numismatic` will be a good showcase of where the Streams functionality really simplifies the design. We're still sorting out some basics and putting in more functionality but I'll ping you once I think it's worth trying out.

mrocklin:
> I didn't get to say this earlier but thank you for creating streamz. I'm loving it so far!

I'm glad.  I encourage you to speak up in the future.  Early users are helpful in determining the direction of a project.

> I think numismatic will be a good showcase of where the Streams functionality really simplifies the design. We're still sorting out some basics and putting in more functionality but I'll ping you once I think it's worth trying out.

I look forward to it

PR #99, opened 2017-10-15
Support synchronous workloads in streamz

mrocklin:
Three main changes:

1.  Support synchronous action in streams.  This allows operations like emit to operate sensibly when called from threads that are not the main event loop.  They will block in the same way that they would wait previously when called with yield in a coroutine
2.  Connect all of the relevant operations to Dask.  Now that sync/async workflow works nicely we can simplify much of the DaskStream connection code
3.  Support `await emit` in the common case

For downstream code this means that we need to be a bit more aware about creating streams in fully asynchronous mode.  I've created a doc page that goes through some of the nuance here

I expect that this may break existing code.

### DIFF
--- a/docs/source/async.rst
+++ b/docs/source/async.rst
@@ -1,0 +1,189 @@
+Asynchronous and Synchronous Computation
+========================================
+
+*This section is only relevant if you want to use time-based functionality.  If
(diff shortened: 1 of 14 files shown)


mrocklin:
cc @snth

PR #101, opened 2017-10-17
Add missing .sink(print) in docs

maxalbert:
Looks like the `num_distinct` example in the ["Accumulating State"](http://streamz.readthedocs.io/en/latest/core.html#accumulating-state) section is missing a `.sink(print)` so that the number of distinct elements actually gets printed.


mrocklin:
Thanks @maxalbert !  Merged

PR #109, opened 2017-10-18
Added map and filter docstrings

jrmlhermitte:
in passing noticed some docstrings not up to date. this isn't everything, just what i was looking through. could add more if needed. @mrocklin I know you are also working on docs, so this may be redundant with what you may already have


mrocklin:
Thanks @jrmlhermitte !

Issue #86, opened 2017-10-10
Should streamz handle exceptions?


jrmlhermitte:
What do we do when a stream receives bad data that causes an exception to be raised. For ex:
```python
def foo(x):
    if x is None:
        raise Exception
    else:
        return x + 1
s = Stream()
s2 = s.map(foo)
s3.sink(print)

s.emit(1)
s.emit(None)
s.emit(2)
```

Here, `foo` is a point of vulnerability in the stream, where it may or may not cause the whole stream architecture to halt.

Is it worth trying to incorporate some quiet exception handling? I am not sure exactly how to tackle this so I'm being a little vague at this point. I can think of many ways of doing this. Here are a few:
1. Catch the exceptions and emit them somehow (will require defining a data type). We can also not emit (and perhaps sink errors to a global list) but this may cause unintended synchronization consequences to the user.
2. catch the exceptions in `s.emit`. Note that in this case catching the exception may be harder to find

I'll think about it, but I would like to hear opinions from @mrocklin  and @CJ-Wright  (who has already handled this in his streams extension). My current method is to wrap all mapped functions to look for exceptions, and return a document that flags the document as having encountered an exception. This works in my subclassed module only though. It would be nice to unify this I think.

mrocklin:
Yes, at some point we should handle exceptions.  Also stopping signals.  I don't have a short term plan for this or any concrete thoughts.

CJ-Wright:
This reminds me a little of the [Try-Monads talk from pygotham](https://2017.pygotham.org/talks/try-monads-with-big-data-using-pyspark/) (not that I think we should necessarily use Try-Mondads).

mrocklin:
If you're curious then I recommend looking at what ReactiveX does here

CJ-Wright:
Relevent? http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/exceptions/Exceptions.html

CJ-Wright:
This might be one area where treating the graph as a walkable DAG may be helpful. One could imagine writing a function which walked the graph, looked for nodes where exceptions could occur (map, filter, etc.) and then decorates update to handle the exceptions, potentially with some callback function, logging, or sending the data + exception to another pipeline.

I think this general pattern, keep the pipeline simple and then decorate things once the DAG has been made is interesting.

mrocklin:
We currently pass messages forward and futures backwards with the emit/update methods.  My guess is that we would do something similar with exception and stop signals.

[...]

PR #111, opened 2017-10-23
Adding messages

jrmlhermitte:
Should we allow messages to pass? I think something like this would be useful.

For example here, we send a clear message which should clear various elements of the stream that hold state. Right now, only this is performed:
```python
    s = Stream()
    # increment
    def acc1(x1, x2):
        return x1 + x2
    from streamz.core import ClearMSG

    cc = ClearMSG()

    sout = s.accumulate(acc1)

    Lout = sout.sink_to_list()

    s.emit(1)
    s.emit(2)
    s.emit(cc)
    s.emit(3)
    assert Lout == [1, 3, 3]
```

but I'd find this useful. One could change these messages to node specific, like "CLEAR_ACC", or some control stream with flags like: `Clear(clear_acc=True,...)`.

any thoughts? This is meant to open discussion. I find it more convenient than an issue in this case.
The various messages or message handler can perhaps also be handled by dispatching, to allow for more convenient subclassing.

This was a quick thought. If it's of interest in the base class we can discuss adding more.

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -28,6 +28,18 @@ thread_state = threading.local()
 logger = logging.getLogger(__name__)
 
 
+class ClearMSG:
+    pass
+
+
+class IgnoreMSG:
+    pass
+
+
+class StopMSG:
+    pass
+
+
 def identity(x):
     return x
 
@@ -277,7 +289,7 @@ class Stream(object):
         return scatter(self, **kwargs)
 
     def remove(self, predicate):
-        """ Only pass through elements for which the predicate returns False """
+        """ Only pass through elements for which the predicate returns False"""
         return self.filter(lambda x: not predicate(x))
 
     @property
@@ -399,11 +411,13 @@ class sink(Stream):
         _global_sinks.add(self)
 
     def update(self, x, who=None):
-        result = self.func(x, *self.args, **self.kwargs)
-        if gen.isawaitable(result):
-            return result
-        else:
-            return []
+        # only if not a clear msg, which is ignored
+        if not isinstance(x, ClearMSG):
+            result = self.func(x, *self.args, **self.kwargs)
+            if gen.isawaitable(result):
+                return result
+            else:
+                return []
 
 
 @Stream.register_api()
@@ -516,6 +530,7 @@ class accumulate(Stream):
 
     def __init__(self, upstream, func, start=no_default, returns_state=False,
                  **kwargs):
+        self.start = start
         self.func = func
         self.kwargs = kwargs
         self.state = start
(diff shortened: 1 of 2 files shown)


mrocklin:
Do other systems do this?  If so then what are their semantics?

The alternative here would be to ask the user to implement this sort of logic within their operations, either in their `acc1` function or as a separate stream operation.

streamz/tests/test_core.py
```diff
@@ -931,3 +931,28 @@ def slow_write(x):
...
+    s = Stream()
+    # increment
+    def acc1(x1, x2):
+        return x1 + x2
+    from streamz.core import ClearMSG, IgnoreMSG
+
+    clear_msg = ClearMSG()
+    ignore_msg = IgnoreMSG()
+
+    sout = s.accumulate(acc1)
```
mrocklin: What would happen if you had a map before the accumulate?  Does map (or all other stream operations) also need to know what to do with ClearMSGs?

CJ-Wright: I would imagine that all `update` methods will need a cutout for messages.

CJ-Wright:
I'm not certain that we can ask the user to modify everything, as there are some cases where we buffer information without user input. For example `zip_latest`. In the case of `zip_latest` you may decide that all the information in the zip_latest buffer is no longer applicable and you want it to go back into "buffer my main stream until my secondary stream emits" mode. A clear message could be helpful here.

jrmlhermitte:
I will have to get back to you about this, sorry busy.
I spent a few min looking at ReactiveX.
They split `emit` into three different functions: `onNext` (regular emit), `onError` and `onCompleted`.
The way error handling works is that when an error is received, the node 'disconnects' (`unsubscribe`s) from its parent stream and the error is passed through to the other elements in the stream. This effectively stops data from passing through the stream graph.

If you want to prevent this behaviour from happening, you must insert a [`Retry`](http://reactivex.io/documentation/operators/retry.html) node in between the node that could return an error and the node that would receive an error. Upon receiving an error, this node will not pass it through and resubscribe to its parent stream.

I don't like this method but I do like the splitting of `emit` into different methods. We could have `emit` dispatch the element to something like `onNext`, `onCompleted`, `onClear` and `onError`. All accumulators, zips etc would then have to implement their own `onClear` methods for example. The default for `onClear` would be `onNext` (`emit`).

Thoughts? This ties in with #86 as well I believe.

jrmlhermitte:
(Just for extra reference, @CJ-Wright has already implemented something similar on the `update` level [here](https://github.com/xpdAcq/SHED/blob/master/shed/event_streams.py#L271). `dispatch` will dispatch to a `start`, `event`, `descriptor` and `stop` methods which intercept different types of messages/documents.)

streamz/core.py
```diff
@@ -28,6 +28,16 @@
 logger = logging.getLogger(__name__)
 
 
+class ClearMSG:
```
CJ-Wright: You may want a MSG base class so you can just check if an item is an instance if that class

martindurant:
superceded by #249 and its discussion, I think

PR #114, opened 2017-10-25
Add 'six' to requirements

Mec-iS:


mrocklin:
Yup, that looks like a bug.  Thanks!

PR #116, opened 2017-10-27
Stream messages (branch off of @jrmlhermitte)

CJ-Wright:


CJ-Wright:
@mrocklin I could use some help when you get a moment, it seems that tornado does not like passing the `IgnoreMSG` object. General comments on the are also appreciated.

mrocklin:
This whole approach still seems a bit off to me.  It adds new complexity to *every* stream.  It's not clear to me that this is a win.

E       AttributeError: class IgnoreMSG has no attribute '__mro__'

This is caused by using old-style objects in Python 2

use the following instead

```python
# class Foo:
class Foo(object):
```

CJ-Wright:
I meant the `tornado.gen.BadYieldError: yielded unknown object <streamz.core.IgnoreMSG object at 0x7f4a84b43be0>` error.

mrocklin:
This means that at some point you are calling `yield some_msg_object` when instead you should be yielding a tornado future.

PR #129, opened 2017-11-24
Add HoloViews based plotting API

philippjfr:
Adds handling for bar plots on ``Series`` and ``Seriess`` objects. Also fixes ``Seriess.plot.line``, ``Seriess.plot.scatter`` and ``Seriess.plot.area``, ensuring the index is reset and therefore made visible to HoloViews.


mrocklin:
FYI I've changed the base of this PR to master

philippjfr:
Okay, sounds good if you're happy with this living on master already. I should have some time over the weekend to go through this more thoroughly and make sure all the options that are currently exposed are hooked up correctly and make a list of other options that might be nice to expose.

philippjfr:
Happy to work on some tests over the weekend.

mrocklin:
Things are going well.  I'm running into some issues in the Jupyter Notebook when my Stream is on a different IOLoop than the Jupyter Notebook's.  This commonly arises whenever we are using a Dask Stream, such as can be created in the following example.

https://gist.github.com/d4fffaecfe73d40061ba634e631bc433

philippjfr:
Looks like it's being passed a ``Future``, I tried handling that by resolving with Future.result() but didn't have any look. I'd have to read up more to understand what's going, do you have any immediate ideas?

mrocklin:
Ah, I've added a `gather` call in a recent commit.  This call converts to a local stream.  On normal streams it is a safe no-op.

mrocklin:
@philippjfr checking on your time availability here.  Are you available to work on tests here?  "No" is a fine answer, I'm just planning things.

philippjfr:
I probably can't devote a lot of time to it, but I should be able to at least write some tests for the core functionality over the weekend.

mrocklin:
OK, I appreciate it.   This was a hit at PyData NYC

mrocklin:
I apologize for letting this linger for so long.  Thank you @philippjfr for your recent efforts.  One last thought is that we should maybe add a small docpage or section to the dataframes page?

mrocklin:
I'm curious if anyone else has time to review this as well.  Maybe @CJ-Wright ?

streamz/dataframe/holoviews.py
```diff
@@ -0,0 +1,672 @@
+from __future__ import absolute_import
```
CJ-Wright: Top level docs?

philippjfr:
That definitely sounds like a good idea, that leaves the question of how to build the page. Since they are streaming plots and empty to begin with it doesn't make much sense to include interactive bokeh plots. Should I record some gifs or just static images to include? I obviously also don't want to bloat the repository too much.

mrocklin:
We don't necessarily even need to have images in the documentation to start
with, just a brief description of how things work and what to expect in
different situations.

philippjfr:
I've added a draft of some fairly detailed documentation, which you can view [here](http://assets.holoviews.org/tmp/streamz/plotting.html). It does include a fair number of images, which take up about 1.8 MB. That would quadruple the size of the repo, so alternatively I could add a script which generates the images. It would complicate the doc building process a bit and take a little bit of time though. Let me know what you'd prefer.

CJ-Wright:
Personally I'd prefer the doc building option, but I don't know how read the docs will handle that. However, my preference is not particularly strong.

jbednar:
Looks great to me! I'd also love to see HoloViews+Bokeh support like this in pandas itself as discussed for https://github.com/pandas-dev/pandas/issues/14130 .

> In addition, options can be passed directly to HoloViews providing greater control over the plots. The options can be provided as dictionaries via the plot_opts and style_opts keyword arguments. You can also apply options using the HoloViews API (for more information see the HoloViews User 

Can you add a brief example of doing that?  It's not obvious to me where those keyword arguments would go.

philippjfr:
>I don't know how read the docs will handle that. 

That's a good point, I might be able to do it with the [IPython Sphinx Directive](https://ipython.org/ipython-doc/3/api/generated/IPython.sphinxext.ipython_directive.html). That's what pandas seems to do.

docs/source/plotting.rst
```diff
@@ -0,0 +1,426 @@
+Plotting
+========
+
+The plotting interface on streamz DataFrame and Series objects
+attempts to mirror the pandas plotting API, but instead of plotting
+with matplotlib_ uses HoloViews_ to generate dynamically streaming
```
CJ-Wright: Do we need an a pronoun? "with matplotlib *it* uses HoloViews".

philippjfr: Yes, reads better that way.

philippjfr:
>Can you add a brief example of doing that? It's not obvious to me where those keyword arguments would go.

Sure I'll add an example.

philippjfr:
Sorry I've left this sitting for so long, been completely swamped with other stuff. The one good thing about that is that I've now designed a very similar plotting API for the intake project which is about to be open-sourced (https://github.com/ContinuumIO/intake/pull/36). The API is a bit more comprehensive and robust so I'll likely be porting some of those ideas back here. I'll try to finish this off after the HoloViews 1.10 release next week.

philippjfr:
Quick update, as I mentioned above I've been working on a similar project as part of intake. Given that this is a common need, I've therefore decided that the plotting API should be in a separate repository (https://github.com/pyviz/hvplot). This has the benefit that the code doesn't get duplicated across a bunch of libraries, stopping the API from diverging and that it will be one shared API with the goal that it can be used by pandas, dask, streamz, xarray and geopandas.

mrocklin:
Sounds sensible to me.  What is the right way for streamz to leverage this
approach?

philippjfr:
Probably very similarly to the way it works now, you'd have a ``.plot`` property on the DataFrame/Series objects passing itself to the ``hvplot`` object, which offers the varying plotting methods.

mrocklin:
Is this something that you anticipate having the time to do or is this
something that others here would have to take on?  (happy either way, just
curious)

philippjfr:
>Is this something that you anticipate having the time to do or is this
something that others here would have to take on?  (happy either way, just
curious)

Would be happy to do it, basically I'm hoping to get an initial release out and then start shopping the API around with the various projects and help integrate it if desired. I expect to natively support the following datatypes:

* pandas: DataFrame, Series
* streamz: DataFrame(s), Series(s)
* dask: DataFrame, Series, Array
* xarray: Dataset, DataArray
* geopandas: DataFrame

mrocklin:
Cool. I'm looking forward to this.

CJ-Wright:
Would it be possible to support `Streamz` classes too?

philippjfr:
>Would it be possible to support Streamz classes too?

Working on that right now, I want to port the material I've written in this PR to the [HvPlot site](https://pyviz.github.io/hvplot/index.html).

CJ-Wright:
Sorry I wrote the wrong thing, I meant the `Streams` classes themselves (not just the Pandas/Series driven classes)

philippjfr:
>Sorry I wrote the wrong thing, I meant the Streams classes themselves (not just the Pandas/Series driven classes)

Ah good question, I haven't played around with these much yet but at least in theory I see no reason why we couldn't support them as long as they emit one of the aforementioned data types.

CJ-Wright:
Do you mean that the `Stream` class would need to emit pandas or series data?

martindurant:
@philippjfr , should this issue be closed now, with the emergence of "pyviz"? At the very least, I am assuming that any work on the streamz side will now be very different than this PR, so probably best to start over.

philippjfr:
Yes, I'll close. Hoping to have an initial release of holoplot in next week and will open a new PR here.

martindurant:
"holoplot" - that will do :)  I did not see the name change before.

philippjfr:
It's the best we've got for now, not too late to find something better though :-)

mrocklin:
I'll be giving a talk on real-time dask things about a week from now.
Having some ability to do streaming plots would be valuable for that if
that is possible.

philippjfr:
If you're not averse to installing from master you can see how to do it here: https://pyviz.github.io/holoplot/user_guide/Streaming_Plots.html

Currently holoplot simply provides a hook to monkey-patch streamz objects with the plot methods. I can probably also get an early conda dev release out by then if needed.

martindurant:
Is it still structured, that you can call `holoplot.plot(stream_thing)` without the patching?

philippjfr:
>Is it still structured, that you can call holoplot.plot(stream_thing) without the patching?

Yes, in fact that's precisely how libraries that want to integrate it properly should use it. They'd define a plot property on the data containers like this:

```python
@property
def plot(self):
    from holoplot import HoloPlot
    return HoloPlot(self)
```

PR #130, opened 2017-11-30
Minor fixes

mrocklin:
1.  Log exceptions in map and accumulate
2.  Add nlargest

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -222,7 +222,7 @@ class Stream(object):
 
     def _emit(self, x):
         result = []
-        for downstream in self.downstreams:
+        for downstream in list(self.downstreams):
             r = downstream.update(x, who=self)
             if type(r) is list:
                 result.extend(r)
@@ -481,9 +481,13 @@ class map(Stream):
         Stream.__init__(self, upstream, stream_name=stream_name)
 
     def update(self, x, who=None):
-        result = self.func(x, *self.args, **self.kwargs)
-
-        return self._emit(result)
+        try:
+            result = self.func(x, *self.args, **self.kwargs)
+        except Exception as e:
+            logger.exception(e)
+            raise
+        else:
+            return self._emit(result)
 
 
 def _truthy(x):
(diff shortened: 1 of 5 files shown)


streamz/core.py
```diff
@@ -351,7 +351,8 @@ def update_frequencies(last, x):
 
         return self.scan(update_frequencies, start={}, **kwargs)
 
-    def visualize(self, filename='mystream.png', source_node=False, **kwargs):
+    def visualize(self, filename='mystream.png', source_node=False,
+                  rankdir='LR', **kwargs):
```
mrocklin: @CJ-Wright , this might affect you.  Any objections?

CJ-Wright: What was the reason for this? I'm generally ok with this, since I can override it.

mrocklin: I generally prefer left-to-right as a default over top-to-bottom.  I can go either way though.

streamz/core.py
```diff
@@ -481,9 +483,12 @@ def __init__(self, upstream, func, *args, **kwargs):
         Stream.__init__(self, upstream, stream_name=stream_name)
 
     def update(self, x, who=None):
-        result = self.func(x, *self.args, **self.kwargs)
-
-        return self._emit(result)
+        try:
+            result = self.func(x, *self.args, **self.kwargs)
+        except Exception as e:
+            logger.exception(e)
```
CJ-Wright: Do we want an option to raise?

mrocklin: Eventually we probably want to percolate exceptions as we do data or futures.  I don't know what defaults should be in the meantime.

mrocklin: If we should raise I'm fine with that

streamz/core.py
CJ-Wright: I guess we should default to something. I usually prefer having the stream direction and the names orthogonal so that we fill the page more, but I don't mind putting in the kwarg.

mrocklin: OK, removing.  Given the difference in opinion we'll stick to graphviz' default

streamz/core.py
mrocklin: We now raise

streamz/core.py
CJ-Wright: Thank you. I'll make a point of explicitly putting top-bottom in my code just in case we do make the shift.

Issue #131, opened 2017-11-30
Problem with `timed_window` with asyncio


snth:
Hi,

When I try to use the `timed_window` function with the `asyncio` event loop I get the following error:

    AttributeError: 'AsyncIOMainLoop' object has no attribute '_running'

For example if you change line 12 of the `fib_asyncio.py" example to:

    s.timed_window(1.0).sink(lambda x: print(x))

then I get:

```
$ python fib_asyncio.py
Traceback (most recent call last):
  File "fib_asyncio.py", line 14, in <module>
    source.emit(0)                          # seed with initial values
  File "/opt/venv-numismatic/lib/python3.6/site-packages/streamz/core.py", line 219, in emit
    return sync(self.loop, _)
  File "/opt/venv-numismatic/lib/python3.6/site-packages/streamz/core.py", line 1068, in sync
    if not loop._running:
AttributeError: 'AsyncIOMainLoop' object has no attribute '_running'

```

mrocklin:
Yeah, that's a bug.  The check on `IOLoop._running` is Tornado-specific.  Fortunately, that check is also a historical artifact, and can be removed without a problem.  @snth do you have any interest in submitting a PR that removes this conditional block and perhaps also includes a small test?  We should have some asyncio tests.

snth:
@mrocklin Could you describe why it is safe to remove those checks? I didn't look at more than 2 or 3 lines around the line that caused the problem so don't understand the overall context and how that code integrates into the bigger picture.

mrocklin:
They aren't used within this project.  The utility `sync` function was pulled from Dask, where that functionality was previously needed.  It has no effect in this project.

PR #132, opened 2017-11-30
Bug timed window

snth:
Very hacky workaround for the timed_window AttributeError.

It seems to me that the `AsyncIOMainLoop` in `tornado` doesn't implement `loop._running` properly (it's not implemented at all). Given that it starts with an underscore, I don't know if it forms part of the API and needs to be implemented?

This fixes the `fib_asyncio.py` example. However in the application where I discovered this and where I am trying to use it my whole app now freezes. Still investigating what's going on there. It would be good to get some more eyes on this to see if my workaround works for other people.

### DIFF
--- a/examples/fib_asyncio.py
+++ b/examples/fib_asyncio.py
@@ -9,7 +9,7 @@ s = source.sliding_window(2).map(sum)
 L = s.sink_to_list()                    # store result in a list
 
 s.rate_limit(0.5).sink(source.emit)         # pipe output back to input
-s.rate_limit(1.0).sink(lambda x: print(L))  # print state of L every second
+s.timed_window(1.0).sink(lambda x: print(x))
 
 source.emit(0)                          # seed with initial values
 source.emit(1)
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1130,7 +1130,14 @@ def sync(loop, func, *args, **kwargs):
         else:
             return gen.with_timeout(timedelta(seconds=timeout), coro)
 
-    if not loop._running:
+    try:
+        loop_is_running = loop._running
+    except AttributeError:
+        try:
+            loop_is_running = loop.asyncio_loop.is_running()
+        except AttributeError:
+            raise NotImplementedError(type(loop))
+    if not loop_is_running:
         try:
             return loop.run_sync(make_coro)
         except RuntimeError:  # loop already running
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -256,7 +256,7 @@ class Stream(object):
                     del thread_state.asynchronous
 
                 raise gen.Return(result)
-            return sync(self.loop, _)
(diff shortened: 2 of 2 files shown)


snth:
This should hopefully fix #131.

examples/fib_asyncio.py
```diff
@@ -9,7 +9,7 @@
 L = s.sink_to_list()                    # store result in a list
 
 s.rate_limit(0.5).sink(source.emit)         # pipe output back to input
-s.rate_limit(1.0).sink(lambda x: print(L))  # print state of L every second
+s.timed_window(1.0).sink(lambda x: print(x))
```
mrocklin: Did you mean to change this in the PR?

streamz/core.py
```diff
@@ -1130,7 +1130,14 @@ def make_coro():
...
 
-    if not loop._running:
+    try:
+        loop_is_running = loop._running
+    except AttributeError:
+        try:
+            loop_is_running = loop.asyncio_loop.is_running()
+        except AttributeError:
+            raise NotImplementedError(type(loop))
+    if not loop_is_running:
```
mrocklin: I recommend removing this code rather than protecting it.

examples/fib_asyncio.py
snth: Yes. It changes the functionality a bit but now this example tests the rate_limit() and timed_window() methods. I've been using this to test the other changes I've been making in this PR.

streamz/core.py
snth: Ok thanks for the pointer. I've done that. The fib_asyncio.py example runs fine now. I tried running the tests but I'm having trouble with my setup on Windows. Travis also seems to be reporting a failing check.

snth:
Unfortunately I doubt I'll be able to add any tests. Sorry about the file permissions. That was not intended. My setup on Windows under MobaXterm is not ideal. I'm not sure why the permissions were changed.

I removed the `sync` method. I hope that is what you intended. If you could please test it on your side and maybe let me know what's causing the Travis checks to fail.

mrocklin:
You can see what tests are failing by clicking on the details link for the travis-ci failure.  Here is a direct link: https://travis-ci.org/mrocklin/streamz/jobs/312773565

martindurant:
`loop._running` was removed here https://github.com/mrocklin/streamz/pull/152/files#diff-d4e512993f47710ea9e7155d66011a24L1177 , along with a firmer view on event loops and threads.

`fib_asyncio.py` appears to work OK on this end, so perhaps this PR has actually be resolved?

snth:
Thanks for the ping. I'll try to rebase on a recent master in the next few days and check if it is resolved.

martindurant:
@snth , is this still an issue? Would you like to base on current master, where things are generally passing again?

PR #137, opened 2018-01-03
FIX: connect uses append on the upstreams, we need to make sure that

CJ-Wright:
the upstreams is appendable

### DIFF
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -717,6 +717,7 @@ def test_multi_connect():
 
     assert L == [3, 5]
 
+
 def test_disconnect():
     source = Stream()
 
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -679,7 +679,7 @@ def test_connect():
     source_downstream = Stream()
     # connect assumes this default behaviour
     # of stream initialization
-    assert not(source_downstream.downstreams)
+    assert not source_downstream.downstreams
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@mrocklin ready for review

martindurant:
I can't see this causing any problem, but I also can't see a situation in which it's necessary. @CJ-Wright , did you come across a problem, perhaps something we should test for?

CJ-Wright:
@martindurant please see the test

streamz/tests/test_core.py
```diff
@@ -696,6 +696,28 @@ def test_connect():
     assert L == [3, 5]
 
 
+def test_multi_connect():
+    source0 = Stream()
+    source1 = Stream()
+    source_downstream = source0.union(source1)
+    # connect assumes this default behaviour
+    # of stream initialization
+    assert not(source_downstream.downstreams)
```
martindurant: should be `not source_downstream.downstreams` ?

CJ-Wright:
The reason why this fails is because `self.upstreams` was stored as a tuple, and `Stream.connect` tries to append to that tuple, which is invalid.

martindurant:
The test failure appears unrelated?

CJ-Wright:
I think so, rerunning to see if it reproduces (it didn't on my local)

martindurant:
Should merge from master following #160

PR #141, opened 2018-01-10
ENH: support dicts in unique

### LINKED ISSUE #140: unique fails on dicts
```python
from streamz import Stream
source = Stream()
s = source.unique().sink(print)
source.emit({'a': 1})
```
gives an error
```
~/dev/streamz/streamz/core.py in emit(self, x, asynchronous)
    243                 thread_state.asynchronous = True
    244             try:
--> 245                 result = self._emit(x)
    246                 return gen.convert_yielded(result)
    247             finally:

~/dev/streamz/streamz/core.py in _emit(self, x)
    224         result = []
    225         for downstream in list(self.downstreams):
--> 226             r = downstream.update(x, who=self)
    227             if type(r) is list:
    228                 result.extend(r)

~/dev/streamz/streamz/core.py in update(self, x, who)
    992     def update(self, x, who=None):
    993         y = self.key(x)
--> 994         if y not in self.seen:
    995             self.seen[y] = 1
    996             return self._emit(x)

TypeError: unhashable type: 'dict'
```
Is this expected behavior?

CJ-Wright:
Closes #140

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -7,6 +7,7 @@ import logging
 import six
 import sys
 import threading
+from collections import MutableMapping
 from time import time
 import weakref
 
@@ -524,7 +525,10 @@ class starmap(Stream):
         Stream.__init__(self, upstream, stream_name=stream_name)
 
     def update(self, x, who=None):
-        y = x + self.args
+        if self.args:
+            y = x + self.args
+        else:
+            y = x
         try:
             result = self.func(*y, **self.kwargs)
         except Exception as e:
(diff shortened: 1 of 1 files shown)


mrocklin:
FYI I am unlikely to review this in the next week or so.

CJ-Wright:
I'd really like this to get in before the 16th (we're deploying at the NSLS) do you want me to just merge, seek review from someone else or something else entirely. I'd rather not just work from my fork since I'd like to give the NSLS a stable release (but I can do that too if needed).

mrocklin:
I'm -1 on these changes.  This seems too special cased to me.  What about other non-dicts that are not hashable?

I recommend doing what toolz.unique does and accept a key function.

mrocklin:
> I'd really like this to get in before the 16th (we're deploying at the NSLS) do you want me to just merge, seek review from someone else or something else entirely. I'd rather not just work from my fork since I'd like to give the NSLS a stable release (but I can do that too if needed).

This seems unsustainable to me.  I think that you need to develop a long term solution that does not require my short-term involvement.  I am not comfortable with others just-merging into this fork.  I recommend forking the project for your needs or developing your own Stream objects in an auxiliary library.  If there is anything I can do to help improve extensibility so that coordination is not as important please let me know.

CJ-Wright:
I guess my best bet then is to build an extension library that adds functionality.

mrocklin:
> I guess my best bet then is to build an extension library that adds functionality.

I agree that this would probably reduce friction and allow you to develop and release according to your own needs.

mrocklin:
It looks like there already is a key= keyword to unique.  Perhaps that
solves your immediate problem.

CJ-Wright:
I guess I don't understand how that is supposed to work, do I then try to hash (or something similar) the dict in the `key` function?

mrocklin:
http://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.unique  The
key is up to you

PR #142, opened 2018-01-10
ENH: add starsink (sink analogue of starmap)

CJ-Wright:


streamz/core.py
```diff
@@ -447,6 +447,49 @@ def update(self, x, who=None):
...
 
+@Stream.register_api()
+class starsink(Stream):
+    """ Apply a function on every element
+
+    Examples
+    --------
+    >>> source = Stream()
+    >>> L = list()
+    >>> source.sink(L.append)
```
martindurant: The example here doesn't seem to demonstrate starsink at all.

CJ-Wright: Sorry, too much copy paste :sheep:

PR #150, opened 2018-03-01
Improve kafka source

### LINKED ISSUE #143: `streamz.from_X` sources not working in jupyter notebook ?
I'm playing around with `streamz` trying to get to know the project. I'm using the latest code from the `master` branch and I'm having a problem getting the Kafka source to work in a notebook.

```python
import json

import pandas as pd
import streamz


source = streamz.from_kafka(['mytopic'], {
    'bootstrap.servers': 'localhost',
    'group.id': 'notebook',
    'default.topic.config': {
        'auto.offset.reset': 'latest'   
    }
}).map(json.loads)

sdf = source.to_dataframe(example=pd.DataFrame(
    [[0.0, 0.0, 0.0]], columns=['a', 'b', 'c']
))

sdf.tail()
```

At this point I'm expecting some output in the notebook but nothing shows.
I don't see any errors in the browser's console, nor in jupyter's logs.

When trying this in a stand-alone script using tornado's IOLoop I can get the source/stream to work, I do see my data when trying `source.sink(print)` which makes me think the `load_kafka()` part works fine.

Any ideas about what I could be doing wrong or what I should investigate? If this is a bug I'm willing to help fix it and submit a PR. Thanks for the awesome project!

edit: FTR other sources work, I was able to reproduce http://matthewrocklin.com/blog/work/2017/10/16/streaming-dataframes-1, for example the `Random` plotting example at the end works fine.

martindurant:
Fixes #143

The idea is that dynamic polling should always be marked as async, and latch on to any IOLoop going, or the one passed. 

I hope this fixes the jupyter display problem - it certainly works from `print`, although that has the problem of the actual output appearing wherever the latest code was executed.

### DIFF
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -44,6 +44,11 @@ class from_textfile(Source):
     f: file or string
     poll_interval: Number
         Interval to poll file for new data in seconds
+    delimiter: str ("\n")
+        Character(s) to use to split the data into parts
+    start: bool (False)
+        Whether to start running immediately; otherwise call stream.start()
+        explicitly.
 
     Example
     -------
@@ -56,25 +61,39 @@ class from_textfile(Source):
     -------
     Stream
     """
-    def __init__(self, f, poll_interval=0.100, **kwargs):
+    def __init__(self, f, poll_interval=0.100, delimiter='\n', start=False,
+                 **kwargs):
         if isinstance(f, str):
             f = open(f)
         self.file = f
+        self.delimiter = delimiter
 
         self.poll_interval = poll_interval
         super(from_textfile, self).__init__(ensure_io_loop=True, **kwargs)
+        self.stopped = True
+        if start:
+            self.start()
+
+    def start(self):
+        self.stopped = False
+        self.loop.add_callback(self.do_poll)
 
     @gen.coroutine
-    def start(self):
+    def do_poll(self):
+        buffer = ''
         while True:
-            line = self.file.readline()
+            line = self.file.read()
             if line:
-                yield self._emit(line)
+                buffer = buffer + line
+                if self.delimiter in buffer:
+                    parts = buffer.split(self.delimiter)
+                    buffer = parts.pop(-1)
+                    for part in parts:
+                        yield self._emit(part + self.delimiter)
             else:
-                if self.poll_interval:
-                    yield gen.sleep(self.poll_interval)
-                else:
-                    return
+                yield gen.sleep(self.poll_interval)
+            if self.stopped:
+                break
 
 
 @Stream.register_api(staticmethod)
@@ -87,13 +106,16 @@ class filenames(Source):
         Directory path or globstring over which to search for files
     poll_interval: Number
         Seconds between checking path
+    start: bool (False)
+        Whether to start running immediately; otherwise call stream.start()
+        explicitly.
 
     Examples
     --------
     >>> source = Stream.filenames('path/to/dir')  # doctest: +SKIP
     >>> source = Stream.filenames('path/to/*.csv', poll_interval=0.500)  # doctest: +SKIP
     """
-    def __init__(self, path, poll_interval=0.100, **kwargs):
+    def __init__(self, path, poll_interval=0.100, start=False, **kwargs):
         if '*' not in path:
             if os.path.isdir(path):
                 if not path.endswith(os.path.sep):
@@ -102,11 +124,17 @@ class filenames(Source):
         self.path = path
         self.seen = set()
         self.poll_interval = poll_interval
-
+        self.stopped = True
         super(filenames, self).__init__(ensure_io_loop=True)
+        if start:
+            self.start()
+
+    def start(self):
+        self.stopped = False
+        self.loop.add_callback(self.do_poll)
 
     @gen.coroutine
-    def start(self):
+    def do_poll(self):
         while True:
             filenames = set(glob(self.path))
             new = filenames - self.seen
@@ -114,6 +142,8 @@ class filenames(Source):
                 self.seen.add(fn)
                 yield self._emit(fn)
             yield gen.sleep(self.poll_interval)  # TODO: remove poll if delayed
+            if self.stopped:
+                break
 
 
 @Stream.register_api(staticmethod)
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -1,10 +1,12 @@
 from glob import glob
 import os
+import weakref
 
+import time
 import tornado.ioloop
 from tornado import gen
 
-from .core import Stream
+from .core import Stream, convert_interval
 
 
 def PeriodicCallback(callback, callback_time, asynchronous=False, **kwargs):
@@ -161,35 +163,193 @@ class from_kafka(Source):
     consumer_params: dict
         Settings to set up the stream, see
         https://docs.confluent.io/current/clients/confluent-kafka-python/#configuration
+        https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
         Examples:
-        url: Connection string (host:port) by which to reach Kafka
-        group: Identity of the consumer. If multiple sources share the same
+        bootstrap.servers: Connection string(s) (host:port) by which to reach Kafka
+        group.id: Identity of the consumer. If multiple sources share the same
             group, each message will be passed to only one of them.
     poll_interval: number
         Seconds that elapse between polling Kafka for new messages
+    start: bool (False)
+        Whether to start polling upon instantiation
 
     Example
     -------
 
     >>> source = Stream.from_kafka(['mytopic'],
-    ...        dict(url='localhost:9092', group='streamz'))  # doctest: +SKIP
+    ...           {'bootstrap.servers': 'localhost:9092',
+    ...            'group.id': 'streamz'})  # doctest: +SKIP
     """
-    def __init__(self, topics, consumer_params, poll_interval=0.1):
-        import confluent_kafka as ck
+    def __init__(self, topics, consumer_params, poll_interval=0.1, start=False,
+                 **kwargs):
         self.cpars = consumer_params
-        self.consumer = ck.Consumer(consumer_params)
-        self.consumer.subscribe(topics)
+        self.consumer = None
         self.topics = topics
-        self.sleep = poll_interval
+        self.poll_interval = poll_interval
+        super(from_kafka, self).__init__(ensure_io_loop=True, **kwargs)
+        self.stopped = True
+        if start:
+            self.start()
 
-        super(from_kafka, self).__init__(ensure_io_loop=True)
-        self.loop.add_callback(self.poll_kafka)
+    def do_poll(self):
+        if self.consumer is not None:
+            msg = self.consumer.poll(0)
+            if msg and msg.value():
+                return msg.value()
 
     @gen.coroutine
     def poll_kafka(self):
         while True:
-            msg = self.consumer.poll(0)
-            if msg is None or msg.error():
-                yield gen.sleep(self.sleep)
+            val = self.do_poll()
+            if val:
+                yield self._emit(val)
             else:
-                yield self.emit(msg.value())
+                yield gen.sleep(self.poll_interval)
+            if self.stopped:
+                break
+
+    def start(self):
+        import confluent_kafka as ck
+        import distributed
+        if self.stopped:
+            finalize = distributed.compatibility.finalize
+            self.stopped = False
+            self.loop.add_callback(self.poll_kafka)
+            self.consumer = ck.Consumer(self.cpars)
+            self.consumer.subscribe(self.topics)
+
+            def close(ref):
+                ob = ref()
+                if ob is not None and ob.consumer is not None:
+                    ob.consumer.close()
+
+            finalize(self, close, weakref.ref(self))
+
+    def _close_consumer(self):
+        if self.consumer is not None:
+            self.consumer.close()
+            self.consumer = None
+        self.stopped = True
+
+
+class FromKafkaBatched(Stream):
+    """Base class for both local and cluster-based batched kafka processing"""
+    def __init__(self, topic, consumer_params, poll_interval='1s', npartitions=1,
+                 ensure_io_loop=True, **kwargs):
+        self.consumer_params = consumer_params
+        self.topic = topic
+        self.npartitions = npartitions
+        self.positions = [0] * npartitions
+        self.poll_interval = convert_interval(poll_interval)
+        self.stopped = True
+
+        super(FromKafkaBatched, self).__init__(ensure_io_loop=True, **kwargs)
+
+    @gen.coroutine
+    def poll_kafka(self):
+        import confluent_kafka as ck
+        consumer = ck.Consumer(self.consumer_params)
+
+        while not self.stopped:
+            out = []
+
+            for partition in range(self.npartitions):
+                tp = ck.TopicPartition(self.topic, partition, 0)
+                try:
+                    low, high = consumer.get_watermark_offsets(tp, timeout=0.1)
+                except (RuntimeError, ck.KafkaException):
+                    continue
+                current_position = self.positions[partition]
+                lowest = max(current_position, low)
+                out.append((self.consumer_params, self.topic, partition, lowest,
+                            high - 1))
+                self.positions[partition] = high
+
+            for part in out:
+                yield self._emit(part)
+
+            else:
+                yield gen.sleep(self.poll_interval)
+
+    def start(self):
+        self.stopped = False
+        self.loop.add_callback(self.poll_kafka)
+
+
+@Stream.register_api(staticmethod)
+def from_kafka_batched(topic, consumer_params, poll_interval='1s',
+                       npartitions=1, start=False, dask=False, **kwargs):
+    """ Get messages from Kafka in batches
+
+    Uses the confluent-kafka library,
+    https://docs.confluent.io/current/clients/confluent-kafka-python/
+
+    This source will emit lists of messages for each partition of a single given
+    topic per time interval, if there is new data. If using dask, one future
+    will be produced per partition per time-step, if there is data.
+
+    Parameters
+    ----------
+    topic: str
+        Kafka topic to consume from
+    consumer_params: dict
+        Settings to set up the stream, see
+        https://docs.confluent.io/current/clients/confluent-kafka-python/#configuration
+        https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
+        Examples:
+        bootstrap.servers: Connection string(s) (host:port) by which to reach Kafka
+        group.id: Identity of the consumer. If multiple sources share the same
+            group, each message will be passed to only one of them.
+    poll_interval: number
+        Seconds that elapse between polling Kafka for new messages
+    npartitions: int
+        Number of partitions in the topic
+    start: bool (False)
+        Whether to start polling upon instantiation
+
+    Example
+    -------
+
+    >>> source = Stream.from_kafka_batched('mytopic',
+    ...           {'bootstrap.servers': 'localhost:9092',
+    ...            'group.id': 'streamz'}, npartitions=4)  # doctest: +SKIP
+    """
+    if dask:
+        from distributed.client import default_client
+        kwargs['loop'] = default_client().loop
+    source = FromKafkaBatched(topic, consumer_params,
+                              poll_interval=poll_interval,
+                              npartitions=npartitions, **kwargs)
+    if dask:
+        source = source.scatter()
+
+    if start:
+        source.start()
+
+    return source.starmap(get_message_batch)
+
+
+def get_message_batch(kafka_params, topic, partition, low, high, timeout=None):
+    """Fetch a batch of kafka messages in given topic/partition
+
+    This will block until messages are available, or timeout is reached.
+    """
+    import confluent_kafka as ck
+    t0 = time.time()
+    consumer = ck.Consumer(kafka_params)
+    tp = ck.TopicPartition(topic, partition, low)
+    consumer.assign([tp])
+    out = []
+    while True:
+        msg = consumer.poll(0)
+        if msg and msg.value():
+            if high >= msg.offset():
+                out.append(msg.value())
+            if high <= msg.offset():
+                break
+        else:
+            time.sleep(0.1)
+            if timeout is not None and time.time() - t0 > timeout:
+                break
+    consumer.close()
+    return out
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -1063,5 +1063,17 @@ def test_share_common_ioloop(clean):
     assert aa.loop is bb.loop
 
 
+def test_start():
+    flag = []
+
+    class MySource(Stream):
+        def start(self):
+            flag.append(True)
+
+    s = MySource().map(inc)
+    s.start()
+    assert flag == [True]
+
+
 if sys.version_info >= (3, 5):
     from streamz.tests.py3_test_core import *  # noqa
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -19,7 +19,7 @@ import streamz as sz
 from streamz import Stream
 from streamz.sources import sink_to_file, PeriodicCallback
 from streamz.utils_test import (inc, double, gen_test, tmpfile, captured_logger,
-        clean)
+        clean, await_for)
 from distributed.utils_test import loop
 
 
@@ -743,12 +743,14 @@ def test_from_file():
             f.flush()
 
             source = Stream.from_textfile(fn, poll_interval=0.010,
-                                          asynchronous=True)
+                                          asynchronous=True, start=False)
             L = source.map(json.loads).pluck('x').sink_to_list()
 
             assert L == []
 
             source.start()
+
+            yield await_for(lambda: len(L) == 3, timeout=5)
 
             assert L == [1, 2, 3]
 
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -919,13 +919,14 @@ def dont_test_stream_kwargs(clean):
 
 @pytest.fixture
 def thread(loop):
-    from threading import Thread
+    from threading import Thread, Event
     thread = Thread(target=loop.start)
     thread.daemon = True
     thread.start()
 
-    while not loop._running:
-        sleep(0.01)
+    event = Event()
+    loop.add_callback(event.set)
+    event.wait()
 
     return thread
 
--- a/streamz/tests/test_kafka.py
+++ b/streamz/tests/test_kafka.py
@@ -1,0 +1,212 @@
+import atexit
+from contextlib import contextmanager
+import os
+import pytest
+import random
+import requests
+import shlex
+import subprocess
+from time import sleep
+from tornado import gen
+
+from ..core import Stream
+from ..dask import DaskStream
+from streamz.utils_test import gen_test, wait_for
+pytest.importorskip('distributed')
+from distributed.utils_test import gen_cluster  # flake8: noqa
+
+KAFKA_FILE = 'kafka_2.11-1.0.0'
+LAUNCH_KAFKA = os.environ.get('STREAMZ_LAUNCH_KAFKA', '') == 'true'
+TOPIC = 'test'
+ck = pytest.importorskip('confluent_kafka')
+
+
+def download_kafka(target):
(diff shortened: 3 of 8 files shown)


streamz/sources.py
```diff
@@ -72,10 +75,7 @@ def start(self):
             if line:
                 yield self._emit(line)
             else:
-                if self.poll_interval:
-                    yield gen.sleep(self.poll_interval)
-                else:
-                    return
+                yield gen.sleep(self.poll_interval)
```
martindurant: There is no documentation here of the intent for `poll_interval=0`

streamz/sources.py
```diff
@@ -138,29 +138,60 @@ class from_kafka(Source):
...
+    loop: None or IOLoop
+        Loop in which to run polling as a coroutine
 
     Example
     -------
 
     >>> source = Stream.from_kafka(['mytopic'],
-    ...        dict(url='localhost:9092', group='streamz'))  # doctest: +SKIP
+    ...           {'bootstrap.servers': 'localhost:9092',
+    ...            'group.id': 'streamz'})  # doctest: +SKIP
```
martindurant: This is the actual syntax used by confluent - better to suggest how to use this, than to mangle some set of keywords that we define.

streamz/sources.py
```diff
@@ -138,29 +138,60 @@ class from_kafka(Source):
...
+    def _connect_consumer(self):
+        open('/Users/mdurant/status', 'a').write('_connect')
+        import confluent_kafka as ck
+        import atexit
+        if self.loop is not None:
+            # need some IOLoop to run on
+            self.loop.add_callback(self.poll_kafka)
+        self.consumer = ck.Consumer(self.cpars)
+        self.consumer.subscribe(self.topics)
+        atexit.register(self._close_consumer)  # release on server
```
martindurant: Close is now called on both `del` and shutdown (but only once), so should not have problems with server saturation.

streamz/tests/test_kafka.py
```diff
@@ -0,0 +1,86 @@
...
+import subprocess
+import pytest
+from contextlib import contextmanager
+from tornado import gen
+
+from ..core import Stream
+from streamz.utils_test import gen_test
+
+KAFKA_FILE = 'kafka_2.11-1.0.0'
+LAUNCH_KAFKA = False
```
martindurant: Perhaps this should be an environment variable - test works fine if kafka happens to be running already. I didn't really want to import the test hooks in confluent here, which are rather more than we need.

streamz/sources.py
mrocklin: `weakref.finalize` seems to be the way the cool kids do it these days.  This is done a few times in the dask.distributed codebase if you're interested in grepping there.

mrocklin:
Glad to see a start at a test suite for kafka!

mrocklin:
My first reaction to this is that the asynchronous event loop stuff may be more complex than this.  There are some tricky issues with reusing the notebook's event loop. For example any time any cell is running we'll be blocked.  I suspect that long-term it's better to make our own event loop in a separate thread, similar to how `dask.distributed.Client` does this.

martindurant:
My feeling was that it is best to keep it simple - the kernel loop may be good enough, and if a cell's execution blocks, that can be OK.

mrocklin:
If we're going for actual production use of this then I think that there will be many cases where it will not be ok.  We're not there now though, and this is a difficult issue.  Perhaps its something that I should take on some time in the future.

martindurant:
screen shot 2018-03-06 at 15 00 40

martindurant:
@wapiflapi , the image shows the kafka stream behaving in a notebook. @mrocklin 's point stands, that it is not clear in the long run which ioloop should be used or if we should provide our own (e.g., by depending on dask.distributed), but I think that the current issue is solved.

streamz/sources.py
```diff
@@ -132,35 +131,65 @@ class from_kafka(Source):
...
 
     @gen.coroutine
     def poll_kafka(self):
         while True:
-            msg = self.consumer.poll(0)
-            if msg is None or msg.error():
-                yield gen.sleep(self.sleep)
+            val = self.do_poll()
+            if val:
+                yield self.emit(val)
```
mrocklin: We tend to use `_emit` internally when we know we're asynchronous.  This avoids some logic and keeps the overhead low.

streamz/sources.py
mrocklin: You might want the following here:

```python
import weakref
weakref.finalize(self, consumer.close)
```

If there is some way that the consumer may already have been collected, then I would use a weakref

```python
import weakref

def close(ref):
    consumer = ref()
    if consumer is not None:
        consumer.close()

weakref.finalize(self, close, weakref.ref(self.consumer))
```

streamz/sources.py
```diff
@@ -132,35 +131,65 @@ class from_kafka(Source):
...
+        atexit.register(self._close_consumer)  # release on server
+
+    def _close_consumer(self):
+        if self.consumer is not None:
+            self.consumer.close()
+            self.consumer = None
+
+    def __del__(self):
+        self._close_consumer()
+        # super(from_kafka).__del__()
```
mrocklin: You'll be able to delete this as well once finalize is in place.

streamz/tests/test_kafka.py
```diff
@@ -0,0 +1,96 @@
...
+def test_from_kafka():
+    with kafka_service() as kafka:
+        stream = Stream.from_kafka([TOPIC], ARGS)
+        yield gen.sleep(3)  # empty any lingering messages
+        out = stream.sink_to_list()
+        for i in range(10):
+            kafka.produce(TOPIC, b'value-%d' % i)
+        kafka.flush()
+        # it takes some time for messages to come back out of kafka
+        yield gen.sleep(2)
```
mrocklin: Is there something we can wait on or poll here instead?  Fixed sleeps like this make me nervous.  I find that they often come up as causes of intermittent failures.  I often do something like the following:

```python
start = time()
while not condition:
    yield gen.sleep(0.1)
    assert time() < start + 5
```

streamz/sources.py
```diff
@@ -57,13 +56,16 @@ class from_textfile(Source):
     -------
     Stream
     """
-    def __init__(self, f, poll_interval=0.100, **kwargs):
+    def __init__(self, f, poll_interval=0.100, asynchronous=True, **kwargs):
```
mrocklin: We probably can't merge with this as the default.  I'll try to find some time in the next few days to figure out a solution here.

streamz/sources.py
martindurant: Have you seen error like `TypeError: cannot create weak reference to 'cimpl.Consumer' object` before?

mrocklin: That's odd.  No I haven't.  I recommend raising that upstream.  Something with their C-Extension class must be off.

martindurant: https://github.com/confluentinc/confluent-kafka-python/issues/327

streamz/sources.py
martindurant: Note that I added it to from_textfiles too, for now.

martindurant:
(one test failed on timeout downloading docker images - I don't have the ability to restart it)

martindurant:
The batched source implements essentially what was discussed with @mrocklin elsewhere. It implements multiple partitions, but for a single topic, and sending the batch download function to a dask client directly, so that there is no need to push data to the cluster, it is fetched from a worker.

Obviously, this requires tests, but I would appreciate comments.

martindurant:
f9484ba is a proposal to do the same thing for textfile and from_files (but without an explicit close or stop method). There is a lot of common code here, perhaps all Source.from_* with polling should have a common class.

martindurant:
c35895b fixes #156 ?

martindurant:
(apparently `*parts, buffer = buffer.split()` isn't valid in py2)

streamz/tests/test_core.py
```diff
@@ -1063,5 +1064,23 @@ def test_share_common_ioloop(clean):
...
+        open(fn, 'w')  # touch file - needs to exist
+        stream = Stream.from_textfile(fn, delimiter=delimiter,
+                                      poll_interval=0.01)
+        l = stream.sink_to_list()
+        open(fn, 'a').write('notaline')
+        sleep(0.05)
+        assert l == []
+        open(fn, 'a').write(delimiter)
+        sleep(0.05)
+        assert l == ['notaline' + delimiter]
```
mrocklin: This may not trigger on slow CI systems, leading to intermittent errors.  I recommend polling while this is not true and then asserting that some wildly pessimistic timeout has not passed.  Something like the following:

```python
start = time()
while L != ['notaline' + delimiter]:
    sleep(0.01)
    assert time() < start + 2
```

streamz/tests/test_core.py
```diff
@@ -1063,5 +1064,23 @@ def test_share_common_ioloop(clean):
     assert aa.loop is bb.loop
 
 
+@pytest.mark.parametrize('delimiter', ['\n', '=='])
+def test_from_text(delimiter):
+    with tmpfile() as fn:
+        open(fn, 'w')  # touch file - needs to exist
+        stream = Stream.from_textfile(fn, delimiter=delimiter,
+                                      poll_interval=0.01)
+        l = stream.sink_to_list()
```
mrocklin: Slight preference to use the pronoun `L` rather than `l` for an anonymous list.  This matches style in the rest of the codebase and, I think, Python programs generally (though this could just be what I'm used to).

streamz/sources.py
```diff
@@ -131,35 +129,148 @@ class from_kafka(Source):
...
+                    out.append(get_message_batch(
+                        self.cpars, self.topic, partition, pos, high - 1))
+        return out
+
+    @gen.coroutine
+    def poll_kafka(self):
+        while True:
+            out = self.do_poll()
+            if any(out):
+                yield self._emit(out)
```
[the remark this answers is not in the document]
martindurant: I was wondering about this. I don't know which is preferable, it's a general library design decision on whether we want atomicity or to reduce the number of emit calls. In common cases I don't suppose there are many partitions, but we should still check for None, where there is no new data.

streamz/tests/test_core.py
martindurant: This pattern is too common, it should be in test_utils; but in this case, this is only really like a `yield`, i.e., coroutines should run just once.

streamz/tests/test_core.py
martindurant: pycharm says "function variable should be lower case" :)  I'm happy either way.

streamz/sources.py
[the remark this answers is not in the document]
mrocklin: Semantically I think of these data sources as being streams of records, not streams of batches of records.  When considering efficiency on high volume topics I think we would end up switching to the dask-friendly batched solution, even if dask is not used.  I agree that we shouldn't emit when no data arrives, I think that this will happen naturally if `self.do_poll()` returns an empty list?

streamz/tests/test_core.py
mrocklin: I think that Antoine put something like this in `distributed/utils.py::wait_for`.  In practice I find that I more often just do the while-sleep mechanism above, but I'm happy with either.

streamz/tests/test_core.py
mrocklin: Harumph

martindurant:
> Should we peel this off into a separate PR?

I have no preference here. Those changes are related but don't match the PR title well.

streamz/sources.py
```diff
@@ -131,35 +129,148 @@ class from_kafka(Source):
...
+
+        Will only consider partitions where the current max offset is at least
+        min_batch since the last time messages were fetched.
+        """
+        import confluent_kafka as ck
+        consumer = ck.Consumer(self.cpars)
+        out = []
+        for partition in range(self.npartitions):
+            tp = ck.TopicPartition(self.topic, partition, 0)
+            low, high = consumer.get_watermark_offsets(tp, timeout=0.1)
```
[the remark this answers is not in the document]
martindurant: You get a kafka exception and this partition is skipped. In practice, this doesn't happen so long as the broker is alive.

streamz/sources.py
```diff
@@ -131,35 +129,148 @@ class from_kafka(Source):
...
+        while True:
+            out = self.do_poll()
+            if any(out):
+                yield self._emit(out)
             else:
-                yield self.emit(msg.value())
+                yield gen.sleep(self.poll_interval)
+
+    def start(self):
+        if self.loop is not None:
```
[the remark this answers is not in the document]
martindurant: If simply removing the `if`, you would get `None has no attribute...` - should there be a try/except for a more informative message?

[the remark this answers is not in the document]
mrocklin: Hrm, that is surprising.  Is it sensible to have this stream without an IOLoop?  

If not then can we remove any code path that results in this case?  I wonder if the thing to do here isn't to give a more informative error message, but rather to stop this from happening in the first place.

mrocklin:
> I have no preference here. Those changes are related but don't match the PR title well.

Alternatively maybe keep them in the same PR but rebase to have two separate commits for the different functionality and then use rebase-and-merge?  In this situation I tend to use `git reset master` to clear away current commit history but keep all changes and then use `git add -i` to select out parts to form into the individual commits.  There are likely other ways to do this though.

streamz/tests/test_core.py
```diff
@@ -1070,16 +1070,23 @@ def test_from_text(delimiter):
...
+        wait_for(lambda: outlist == ['notaline' + delimiter, delimiter,
+                                     delimiter], 2)
+
+
+def test_no_loop():
+    stream = Stream.filenames('*', start=False)
+    assert stream.loop is not None
+    stream.loop = None
+    with pytest.raises(RuntimeError):
+        stream.start()
```
mrocklin: This case seems sufficiently unlikely that I'm inclined not to bother with it.

martindurant:
> rebase to have two separate commits for the different functionality

if you think it worthwhile...

mrocklin:
> if you think it worthwhile...

I do.  This seems like two different pieces of functionality that are similar, but quite distinct.  It is often useful to have these things separated in git history when doing historical debugging or pointing others to how functionality exists.  Hopefully this shouldn't take longer than a minute or two to accomplish.  I can do it for you if you prefer.

martindurant:
I'll get to it later today.

mrocklin:
Thanks for doing the rebasing.  I appreciate it.  

I'm curious on what motivated the choice of the `start=True` default.  My understanding is that this is not standard among other streaming systems.

martindurant:
> what motivated the choice of the start=True default

you mean why `True`, or why it exists at all?

mrocklin:
Why `True`.  Its existance seems fine to me as long as we're consistent.

martindurant:
There is no good reason for `True`, except that this matches the previous behaviour, where starting was implicit. In my mind, it makes sense, that real-time streaming is from "now" rather than including all data that might have accumulated in the past - this is what happens if the first poll even happens before any downstream elements are connected. Perhaps that idea is not very well expressed.

mrocklin:
It's well expressed.  I understand the intent.  My experience is that I usually want the opposite.  My workflow is often something like the following:

```python
source = Stream.from_kafka(...)  # make a source

# futz around for a bit

a = source.map(json.loads).map(...).accumulate(...)  # construct a pipeline or two
b = source.count().sink(write)  

source.start()  # great, everything is set up, start the computation going
```

This doesn't matter much in the infinite case where this is going to run forver, but it matters a great deal in the analytics or interactive cases.

Happy to go either way.  I'm not really sure which is best.  

@mariusvniekerk can you voice a preference as a potential user here?  Should streams that read from inputs start reading immediately or should they be triggered explicitly?

gccmbr:
I guess I would categorize myself as a futzer also.  I first define the source, define the source maps, and when setup, I would then start a co-routine with a source.emit(msg) running on an event loop calling a kafka's poll.

I also like to visualize() the map prior to starting the event loop which is always cool to see with streamz (similar to how we visualize dask delayed.. very cool).

The functionality is awesome and its great to have a choice. However, start=True, although similar to kafka's poll behavior, might feel more consistent with streamz/dask as start=False.   Noob vote for start=False !

mariusvniekerk:
I tend to prefer the approach where you set up all the things and then call start.  Its similar to how the other streaming systems work

martindurant:
It seems I was outvoted :|

martindurant:
btw: @mrocklin , I am suddenly worried that `Stream.from_kafka_batched` can produce data or futures, but is always a Stream as opposed to a DaskStream. Thoughts?

mrocklin:
Hrm, yes, that's an interesting conundrum.

One thought would be to change the `from_kafka_batched` method into a normal method instead of a class?  The source stream might be something that just emits the arguments for `get_message_batch` and then we choose to either call `.map(get_message_batch)` or `.scatter().map(get_message_batch)` accordingly based on some provided keyword?

mrocklin:
I suspect that even this will still have some rough edges to work around though.

Thoughts?

mariusvniekerk:
At a future point we should probably figure out how to do partitioned kafka streams where the consumption is actively done on the distributed workers.  
That is wildly out of scope though.

martindurant:
How would it be to have a new DaskSource type that only makes DaskStreams? In this case, that would mean duplicating most of the code (or having a common class to both).

martindurant:
@mariusvniekerk - this already does exactly what you say. The client side only queries for the latest offsets.

mariusvniekerk:
Ah fantastic, i was misreading things then

mrocklin:
> How would it be to have a new DaskSource type that only makes DaskStreams? In this case, that would mean duplicating most of the code (or having a common class to both).

This might make sense.  We would need to do some work on documentation though.  Currently we mostly train people to use `Stream` for everything and `scatter` and `gather` to convert to and from DaskStream objects, which are currently treated as somewhat internal.

martindurant:
Travis error is in `streamz/tests/test_core.py::test_separate_thread_without_time`, py36 only - is this known to be flaky, or has something else changed.
Output:
```

loop = <tornado.platform.asyncio.AsyncIOLoop object at 0x7f21621a1a20>
    @pytest.fixture
    def thread(loop):
        from threading import Thread
        thread = Thread(target=loop.start)
        thread.daemon = True
        thread.start()
    
>       while not loop._running:
E       AttributeError: 'AsyncIOLoop' object has no attribute '_running'
```

mrocklin:
This is likely due to the Tornado 5.0 update.  I'll take a look in a bit.

mariusvniekerk:
Is there a way in dask-stream to ensure the the offset range for a particular kafka topic-partition consistently goes to the same* dask-worker?
For quite a lot of kafka data sources the topic partition is frequently a reasonable aggregation boundary and could avoid lots of shuffling on the dask side to perform aggregations.

streamz/sources.py
```diff
@@ -131,35 +163,205 @@ class from_kafka(Source):
...
+            tp = ck.TopicPartition(self.topic, partition, 0)
+            try:
+                low, high = consumer.get_watermark_offsets(tp, timeout=0.1)
+            except (RuntimeError, ck.KafkaException):
+                continue
+            pos = self.positions[partition]
+            if high - pos > self.min_batch:
+                self.positions[partition] = high
+                if self.client is not None:
+                    future = self.client.submit(
```
mariusvniekerk: guess we can supply a worker set here if we want to keep partition affinity

martindurant:
@mariusvniekerk , no, this is not implemented. The `get_message_batch` calls are independent and go wherever the scheduler guesses there is capacity. I suppose it would be possible to attach a specific worker preference to each partition.

mariusvniekerk:
I would probably say that kind of behavior is an extension since just adding affinity gets complicated fast with uneven partitions

martindurant:
Also note that we only fetch the messages and drop any key/partition information here, so, as written, it would not be possible to group/aggregate anything by partition.

mrocklin:
I suggest that we don't deal with affinity at the moment.  This is something that ideally the scheduler would handle for us automatically.  It's good to have brought it up, I suspect that current behavior would not handle this well, but that can be changed.

mrocklin:
It sounds like that might be a common pattern.  @mariusvniekerk if you're
interested in this topic can I ask you to raise a new issue with how you
would prefer to see these sorts of things operate from a user perspective?

mrocklin:
I might take a crack at potentially refactoring this in a moment

martindurant:
@mrocklin , be my guest! This is just one possible way of doing it.

mrocklin:
Is there a trick to getting tests to work?  I'm getting the following:

    RuntimeError: Timeout waiting for kafka

mrocklin:
```python
    @contextmanager
    def kafka_service():
        if _kafka[0] is None:
            if LAUNCH_KAFKA:
                launch_kafka()
    
            producer = ck.Producer({'bootstrap.servers': 'localhost:9092'})
            producer.produce('test', b'test')
            out = producer.flush(10)
            if out > 0:
>               raise RuntimeError('Timeout waiting for kafka')
```

mrocklin:
Also, shouldn't we be getting npartitions from the consumer somehow?  It seems odd to me for us to specify this manually.

martindurant:
The first time you run kafka, it'll have to download the images, which will take relatively long on a local machine.

There is no way to query the number of partitions, that was one of the requests in https://github.com/confluentinc/confluent-kafka-python/issues/330

mrocklin:
> The first time you run kafka, it'll have to download the images, which will take relatively long on a local machine.

Do I need to trigger this manually?  It's not taking a long time, it's just erring for me.

martindurant:
`docker pull spotify/kafka` should do it.

mrocklin:
> There is no way to query the number of partitions, that was one of the requests in confluentinc/confluent-kafka-python#330

OK, maybe short-term we should switch over to pykafka instead?

martindurant:
I haven't looked at pykafka at all, happy to change things around, though.

mrocklin:
I've tried using docker pull.  I'm still getting issues.  Trying this out in IPython

```python
In [1]: from streamz.tests.test_kafka import kafka_service

In [2]: with kafka_service() as ks:
   ...:     pass
   ...: 
%3|1521666477.719|FAIL|rdkafka#producer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/bootstrap: Connect to ipv4#127.0.0.1:9092 failed: Connection refused
%3|1521666477.720|ERROR|rdkafka#producer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/bootstrap: Connect to ipv4#127.0.0.1:9092 failed: Connection refused
%3|1521666477.720|ERROR|rdkafka#producer-1| [thrd:localhost:9092/bootstrap]: 1/1 brokers are down
```

mrocklin:
OK, I've taken the liberty of pushing a few commits up to your fork.  I hope you don't mind.  Please let me know your thoughts on if you think this approach is sensible or not.

martindurant:
kafka_service only launches kafka if LAUNCH_KAFKA=True (from env var STREAMZ_LAUNCH_KAFKA), or you can call launch_kafka() manually.

mrocklin:
Should we skip those tests if that environment variable isn't set?

martindurant:
It's an option. I was thinking that some people might want to reuse whatever kafka they already had hanging around.

mrocklin:
OK, I guess that makes sense.  We probably need to improve messaging somehow.  As is I think that this will result in a confusing experience for anyone who tries to run tests here.  I might suggest that the environment variable (or perhaps a flag?) either points to an existing Kafka implementation or it creates a new one, or, by default, it skips tests.  I'm somewhat inclined to just skip tests and not worry about an external kafka deployment until that situation arises.

mrocklin:
Tests do now work for me though, so I'm fixing up what I pushed earlier.

mrocklin:
OK @martindurant the approach here could use your feedback

streamz/tests/test_kafka.py
```diff
@@ -0,0 +1,210 @@
...
+def kafka_service():
+    if _kafka[0] is None:
+        if LAUNCH_KAFKA:
+            launch_kafka()
+        else:
+            raise pytest.skip.Exception("Kafka not available. "
+                    "To launch kafka use `export STREAMZ_LAUNCH_KAFKA=true`")
+
+        producer = ck.Producer({'bootstrap.servers': 'localhost:9092'})
+        producer.produce('test-start-kafka', b'test')
```
martindurant: This ought to fail, as the topic does not exist. "test" is automatically created, and I included a `create_kafka_topic()` function in `test_kafka`, that's not actually called.

mrocklin: How hard is it to destroy topics?  It would be nice to avoid reusing the same topic between tests.  Things seem to get a bit messy.

martindurant: confluent_kafka cannot create topics, I don't know if pykafka can - you need to call a docker exec as in the function I wrote, and a similar function with `--delete` would remove it. However, If we *only* run tests on docker, the issue goes away, we can use a new topic for every test.

martindurant: (i.e,. we kill the container after use, so all topics go away)

mrocklin:
I wonder if it would be easy to wrap the command line calls here in a
context manager

martindurant:
Yes indeed, kafka_service could make a new topic each time, perhaps just with a simple counter, and yield the topic name along with the producer. Again, if we are restricting to docker, there is no need to bother cleaning up the topics (or we could).

martindurant:
(NB: the tests didn't exactly fail: py27 reports flake failure, and py36 reports a failure in something apparently unrelated)

mrocklin:
Happy to restrict to docker until that proves painful

mrocklin:
I may cycle off of this for a bit though if you'd like to take over.  How
do you feel about the approach in general.  OK to scatter if dask==True?

martindurant:
Yes, I like what you've done, quite clever way to reduce code/complexity. The documentation might need some massaging to make clear that what you get out of that will need `gather()`ing.

mrocklin:
It may be possible to configure kafka to auto-create topics on first write

martindurant:
Apparently this is possible with the kafka setting `auto.create.topics.enable`, but it may well still have the problem of the topic not being immediately "ready" to be polled.

martindurant:
(that's a setting within the kafka server system, so would need altering the image at/before launch)

martindurant:
In `streamz/tests/test_core.py::test_separate_thread_without_time`
```  
>       while not loop._running:
E       AttributeError: 'AsyncIOLoop' object has no attribute '_running'
```
should be `loop.is_running` ?

mrocklin:
I've just pushed a fix for the _running attribute.  This was removed in tornado 5.  There are ways to get it back, but generally upstream would prefer that we not use it.  The new approach is, I think, cleaner.

martindurant:
Tornado 5 turning out to require a little work across the board.

This PR is probably ready to go. I'm sure we'll iterate more on this, but the functionality here is useful to people.

mrocklin:
OK, I've added tests for some of the smaller bits I added (start, starmap), cleaned up history, and force pushed.

mrocklin:
Ah, sorry, I had force pushed to your fork with a clean history.  My apologies if I've caused some havoc here.

martindurant:
Actually, I'm not sure what has happened here, I might have messed it up on my side. Let's see if the tests can be made to pass again.

mrocklin:
My guess is that you merged in your github fork to your local fork.  Merging tends not to work well after a history-rewrite event.  Typically you would either `reset --hard` on the remote fork before doing extra work, or do work and then cleverly rebase things.

mrocklin:
I'm happy to clean things up if necessary.

martindurant:
The segmentation fault on the py2 tests likely comes from a race condition over confluent_kafka consumers and/or producer still trying to connect to the server (they maintain their own threads for this) while the docker container is being closed; that all happens while pytest is exiting. I don't see this locally, and it was working before on travis too, so it's tricky to debug.

martindurant:
Sorted.

mrocklin:
Thanks @martindurant !

amodwani:
Hello @mrocklin 
I was trying out streams with kafka as you suggested
While running this code 
```python
source = Stream.from_kafka(['test'], {
        'bootstrap.servers': 'localhost:9092',
        'group.id': '0',
        'default.topic.config': {
            'auto.offset.reset': 'latest'   
        }
    })


pipe = source.map(json.loads).map(test)
source.start()
```
gives this error 


Traceback (most recent call last):
  File "kafka.py", line 42, in <module>
    source.start()
AttributeError: 'from_kafka' object has no attribute 'start'

PR #152, opened 2018-03-08
 Manage event loops in separate threads

mrocklin:
This changes the default to create new event loops in separate threads
rather than use IOLoop.current() if asynchronous=True is not set
explicitly.

This is not perfect, but I think it's better.

### DIFF
--- a/streamz/utils_test.py
+++ b/streamz/utils_test.py
@@ -115,4 +115,4 @@ def clean():
     for loop in _io_loops:
         loop.add_callback(loop.stop)
 
-    _io_loops.clear()
+    del _io_loops[:]
--- a/streamz/utils_test.py
+++ b/streamz/utils_test.py
@@ -5,8 +5,11 @@ import six
 import shutil
 import tempfile
 
+import pytest
 from tornado import gen
 from tornado.ioloop import IOLoop
+
+from .core import _io_loops
 
 
 @contextmanager
@@ -105,3 +108,11 @@ def captured_logger(logger, level=logging.INFO, propagate=None):
         logger.setLevel(orig_level)
         if propagate is not None:
             logger.propagate = orig_propagate
+
+
+@pytest.fixture
+def clean():
+    for loop in _io_loops:
+        loop.add_callback(loop.stop)
+
+    _io_loops.clear()
--- a/docs/source/async.rst
+++ b/docs/source/async.rst
@@ -80,14 +80,9 @@ interaction) or when using Dask (next section).
 .. code-block:: python
 
    from streamz import Stream
-   from tornado.ioloop import IOLoop
-   from threading import Thread
-   loop = IOLoop()
-   thread = Thread(target=loop.start, daemon=True)
-   thread.start()
 
-   source = Stream()
-   source.map(increment).rate_limit(0.500, loop=loop).sink(write)
+   source = Stream(ensure_io_loop=True)  # starts IOLoop in separate thread
+   source.map(increment).rate_limit('500ms').sink(write)
 
    for x in range(10):
        source.emit(x)
--- a/docs/source/async.rst
+++ b/docs/source/async.rst
@@ -7,7 +7,7 @@ this section.*
 
 When using time-based flow control like ``rate_limit``, ``delay``, or
 ``timed_window`` Streamz relies on the Tornado_ framework for concurrency.
-This allows us to handle many conncurent processes cheaply and consistently
+This allows us to handle many conncurent operations cheaply and consistently
 within a single thread.  However, this also adds complexity and requires some
 understanding of asynchronous programming.  There are a few different ways to
 use Streamz with a Tornado event loop.
@@ -54,11 +54,10 @@ You may have an application that runs strictly within an event loop.
        source = Stream(asynchronous=True)  # tell the stream we're working asynchronously
        source.map(increment).rate_limit(0.500).sink(write)
 
-       for x in range(10)
+       for x in range(10):
            yield source.emit(x)
 
-   IOLoop.current().add_callback(f)
-   IOLoop.current().start()
+   IOLoop().run_sync(f)
 
 We call Stream with the ``asynchronous=True`` keyword, informing it that it
 should expect to operate within an event loop.  This ensures that calls to
@@ -68,6 +67,22 @@ should expect to operate within an event loop.  This ensures that calls to
 .. code-block:: python
 
    yield source.emit(x)  # waits until the pipeline is ready
+
+This would also work with async-await syntax in Python 3
+
+.. code-block:: python
+
+   from streamz import Stream
+   from tornado.ioloop import IOLoop
+
+   async def f():
+       source = Stream(asynchronous=True)  # tell the stream we're working asynchronously
+       source.map(increment).rate_limit(0.500).sink(write)
+
+       for x in range(10):
+           await source.emit(x)
+
+   IOLoop().run_sync(f)
 
 
 Event Loop on a Separate Thread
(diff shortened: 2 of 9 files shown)


mrocklin:
cc @martindurant

streamz/utils_test.py
```diff
@@ -105,3 +108,11 @@ def captured_logger(logger, level=logging.INFO, propagate=None):
...
         if propagate is not None:
             logger.propagate = orig_propagate
+
+
+@pytest.fixture
+def clean():
+    for loop in _io_loops:
+        loop.add_callback(loop.stop)
+
+    _io_loops.clear()
```
martindurant: This was the cause of the failures on py2. I guess `del _io_loops[:]` would do.

martindurant:
I'll work my way through this. It seems to have percolated many small changed throughout the codebase. 

Probably it could do with some more documentation than the async/loop parameter description docstrings, though.

mrocklin:
I've updated the async docs a bit.  If you are able to put on your "naive user hat" and generate a bunch of questions that would help me to generate useful docs

martindurant:
^ is that comment directed at me?

mrocklin:
Yes.  Feedback on how to make things more clear here would be useful.  If my understanding is correct you're still relatively new to async stuff and so probably still have a decent handle on what would be useful to include in documentation.

martindurant:
Some thoughts on understanding without heavy trialing yet.

I would certainly mention that some sources expect to poll on a regular time-frame (from_textfiles and filenames, never mind kafka). These sources need to be updated for the thread case, I think.

I think the naming of `asynchronous=` is confusing now, really it means "in this thread with an explicit ioloop" versus "in a thread we create for you". Both are concurrent.

Is `DaskStream` no longer meant to be used directly? The async doc still refers to it, but it's gone in the example code. Presumably there should be a caveat here about mixing async styles and dask. Or, in general since `asynchronous=False` will be the default now everywhere, the description should maybe be reordered to reflect this.

"This would also work with async-await syntax in Python 3" - at some point, we should really consider dropping py2... The tornado syntax is scary for anyone that's not seen it before (async/await less so).

mrocklin:
I've fixed the code in sources.  Thanks for pointing me to that.  At some point we'll need actual tests for that :)

To me asynchronous means "asynchronous api" but I get the general point.  This happens to be the term that Dask uses currently, so there is some inertia here.  Alternatives welcome if you have thoughts.

DaskStreams are still usable, but I'm currently recommending that people create them with scatter/gather.  I've altered the docs to be more explicit here.  

I agree about dropping Python 2, I think that that will be for another time though if that's alright.

martindurant:
> "asynchronous api"

Perhaps be more explicit, `background_thread=`?

> the user is expected to call it themselves

`from_textfiles` has this in the docstring, but `filenames` (should be `from_files`?) does not. Also, now it matters *how* the `start()` method gets called - it should be on the `.loop` - so I would expect that to be done in `__init__`.

mrocklin:
> from_textfiles has this in the docstring, but filenames (should be from_files?) does not. Also, now it matters how the start() method gets called - it should be on the .loop - so I would expect that to be done in __init__.

There are many cases where people won't want to start burning their data immediately, but will want an opportunity to set things up beforehand.  This also goes against current conventions in other streaming projects like Spark Streaming or Flink.  I suspect that eventually we'll do something like the following?

```python
def start(self):
    self.loop.add_callback(self._start)
```

But I'd like to wait on this until later.  My goal is to merge this PR soon.  I have some other things that I would like to move on to.

mrocklin:
> Perhaps be more explicit, background_thread=?

I'd rather wait on this decision for the future.  I think that the inertia behind the previous use of asynchronous in this project, and the equivalent use of asynchronous= in Dask make it so that changing this API is perhaps a longer discussion than is likely to happen immediately.

martindurant:
OK, all fair enough - was just using my naive hat. At least, `.start()` should be explicitly called out in the docstrings every time, and any doc pages referring to it, and probably kafka should do the same.

mrocklin:
> OK, all fair enough - was just using my naive hat

Ah, right.  Sorry if I got defensive.  Thank you for the feedback here.  The start thing in particular clearly needs to be spelled out more clearly, I agree.  I may leave this for whatever poor shmuck ends up implementing the next data ingest function :)

martindurant:
I feel implicated by that! 
(which remind me - is the PCAP data ingestion and dashboard published anywhere?)

mrocklin:
Here is an old gist: https://gist.github.com/mrocklin/62ec0d232fea6b9463a9eebc52582553

No guarantees on it working with versions of things today

mrocklin:
OK to merge?

martindurant:
OK. I'll fold it into my stuff and get back to you.

martindurant:
I mean, OK to merge now.

gccmbr:
@mrocklin, @martindurant  Thank you.  I am using Streamz and Dask Distributed to develop a pure python stream-processor for Kafka.  Streamz + Dask Distributed has a great opportunity to be **the** pure python streaming framework. Given the call for async noobs to contribute questions, I answer the call. 

For the Asynchronous Dask examples in the docs:

> Within the Event Loop
> We call Stream with the asynchronous=True keyword, informing it that it should expect to operate within an event loop. This ensures that calls to emit return Tornado futures rather than block.

This is perfectly clear now. A basic asynchronous IOLoop case.

> Event Loop on a Separate Thread
```
from streamz import Stream
source = Stream(asynchronous=False)  # starts IOLoop in separate thread
source.map(increment).rate_limit('500ms').sink(write)
for x in range(10):
    source.emit(x)
```
and the documentation discussing:

> In this case we pass asynchronous=False to inform the stream that it is expected to perform time-based computation (our write function is a coroutine) but that it should not expect to run in an event loop, and so needs to start its own in a separate thread. Now when we call source.emit normally without using yield or await the emit call blocks, waiting on a coroutine to finish within the IOLoop.
> 
> All functions here happen on the IOLoop. This is good for consistency, but can cause other concurrent applications to become unresponsive if your functions (like increment) block for long periods of time. You might address this by using Dask (see below) which will offload these computations onto separate threads or processes.
> 

I understand this as saying that Stream(asynchronous=False) is to run in the write coroutine's async thread, not the current thread?  If so I believe this seems clear.

**Using Dask**

> The DaskStream object is a drop-in replacement for Stream (mostly). Typically we create a Dask client, and then scatter a local Stream to become a DaskStream.
```
from dask.distributed import Client
client = Client(processes=False)  # starts thread pool, IOLoop in seaprate thread
from streamz import Stream
source = Stream()
(source.scatter()       # scatter local elements to cluster, creating a DaskStream
       .map(increment)  # map a function remotely
       .buffer(5)       # allow five futures to stay on the cluster at any time
       .gather()        # bring results back to local process
       .sink(write))    # call write locally
for x in range(10):
    source.emit(x)
```
I understand this as once you scatter, a Stream becomes a Dask Stream.  The source.emit(x) is synchronous, but the Stream is running on separate threads (or cluster). If I understand correctly the Stream is going to be run in an IOLoop created and handled by the client and scheduler not the current IOLoop.  Thus, asynchronous=False here.  If so, I think I understand the confusion over the asynchronous / background_thread naming confusion.  I go into more detail with this at the end of this comment.

**Asynchronous Dask**

> Dask can also operate within an event loop if preferred. Here you can get the non-blocking operation within an event loop while also offloading computations to separate threads.

`from dask.distributed import Client
async def f():
    client = await Client(processes=False, asynchronous=True)
    source = Stream(asynchronous=True)
    source.scatter().map(increment).rate_limit('500ms').gather().sink(write)
    for x in range(10):
        await source.emit(x)
IOLoop().run_sync(f)`

This is explaining that while Dask and Stream are running within the current IOLoop, providing non-blocking execution, while also offloading their work to different threads or cluster.

I understand the confusion over the asynchronous flag.  I had tended to just default to asynchronous=True whenever I wanted to enable Dask with Streamz, but this was incorrect.  I believe I understand now that the asynchronous=True flag is intended to mean that the Stream or Dask Client's results are returning a future running in the current IOLoop and therefore having non-blocking results.  Likewise, if asynchronous=False the Stream or Dask Client is to block and wait synchronously on the Stream or Dask Worker results that are themselves running asynchronously on their own separate IOLoop and thread.  I hope I am not too far off, but there was a call for async noobs by @mrocklin...

martindurant:
@gccmbr , thanks for your comments. Before replying to all that, are you aware of https://github.com/mrocklin/streamz/pull/150 ? I'll be folding the changes here into that, as well as making a kafka batched dispatcher for distributed. We'll be interested in your use case, what you've tried so far, and for feedback on further commits to that PR.

gccmbr:
@martindurant Thank you.  I wasn't aware of #150 and excited to see the pull request topic.  

As for my comments, please don't expend any of your valuable time other than a simple scan.  I was more trying to get it out "in words" half for myself, half for you guys, and hopefully not too confused for the remainder.

mrocklin:
I appreciate your feedback here.  Your understanding of things seems to be
what I intended, so I'm happy that it's mostly understandable :)

Issue #153, opened 2018-03-09
How to use DataFrame Aggregations?


dcferreira:
The documentation [describes how to create a DataFrame Aggregation](https://streamz.readthedocs.io/en/latest/dataframes.html#aggregations). However, it doesn't mention how to use them.

I found a partially working solution for what I want, but am not sure if this is the intended way to use it.
I created an aggregation using the described methods, and want to use it with `DataFrame` and `Window`/`Rolling`.
Then I created a new class that inherits from `DataFrame`, in which I add a method that calls my aggregation.
Assume this (rather stupid) example, in which I keep some state and do some calculations based on it.
```python
from streamz.dataframe import Random, aggregations, Aggregation

class MyAgg(Aggregation):
    def initial(self, new):
        pstate = new[new.x < 0.5].count()
        return pstate, 0
    
    def on_new(self, state, new):
        pstate, n = state
        n += 1
        pstate = pstate + new[new.x < 0.5].count()
        return (pstate, n), pstate / n
    
    def on_old(self, state, old):
        pstate, n = state
        pstate = pstate - old[old.x < 0.5].count()
        return (pstate, n), pstate / n

class MyDF(Random):
    def myagg(self):
        return self.accumulate_partitions(
            aggregations.accumulator,
            agg=MyAgg(),
            start=None, stream_type='updating',
            returns_state=True
        )
    
sdf = MyDF()
sdf
```

This now works:
```python
sdf.myagg()
```

However, this doesn't, because "`Window` has no attribute '`myagg`'":
```python
sdf.window(10).myagg()
```
I could overload the `window` method in `MyDF` with a custom window which supports `MyAgg`, but it doesn't seem like this is the intended way to do it.

I guess the intention with the `Aggregation` API is that is should also work for windows, since it knows what to do when removing data (`on_old`).

So, what is the proper way to use an `Aggregation`?
If it's not implemented, but there is a definition of how it should work, I might be able to do a PR for it.

mrocklin:
Yup, that's a fair question.  I'll clean up the API a little bit and write
docs for this shortly.

martindurant:
Currently, class `Window` has explicit methods for several aggregations, which call `._known_aggregation`. There is no registry as such, but you could always patch a method to do the same thing, if you wish. Better would be for us to provide an explicit route for you to use.

mrocklin:
See https://github.com/mrocklin/streamz/pull/154

dcferreira:
[...]

PR #154, opened 2018-03-09
Solidify API around ``.aggregate(...)`` method

mrocklin:
This provides an entry point for people to use custom aggregations
uniformly on dataframes or windowed dataframes.


martindurant:
This looks good to me. Could perhaps use a test of a non-standard aggregation with a window, could use the example in #153.

PR #160, opened 2018-04-04
Remove plotting

martindurant:


mrocklin:
`git grep plot` shows that there is also an entry in the docs and the dataframes example notebook.  For the notebook I would wait until the holoplot thing arrives but we should remove the mention in documentation along with this change.

martindurant:
@mrocklin , did you want to glance at this?

martindurant:
Fails because the coverage goes down minutely by removing code?

PR #166, opened 2018-04-07
Kafka segfaults fix

### LINKED ISSUE #165: test_from_kafka segfaults
streamz/tests/test_kafka.py::test_from_kafka <- streamz/utils_test.py /home/travis/.travis/job_stages: line 57:  4650 Segmentation fault      (core dumped) coverage run run_tests.py --runslow

cc @martindurant

martindurant:
Fixes #165 (hoping)

If this appears successful, we can push for a release of confluent_kafka (last release in September https://github.com/conda-forge/python-confluent-kafka-feedstock/pull/3 )

### DIFF
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -221,21 +221,26 @@ class from_kafka(Source):
             def close(ref):
                 ob = ref()
                 if ob is not None and ob.consumer is not None:
-                    ob.consumer.close()
+                    consumer = ob.consumer
+                    ob.consumer = None
+                    consumer.unsubscribe()
+                    consumer.close()  # may raise with latest ck, that's OK
 
             finalize(self, close, weakref.ref(self))
 
     def _close_consumer(self):
         if self.consumer is not None:
-            self.consumer.close()
+            consumer = self.consumer
             self.consumer = None
+            consumer.unsubscribe()
+            consumer.close()
         self.stopped = True
 
 
 class FromKafkaBatched(Stream):
     """Base class for both local and cluster-based batched kafka processing"""
-    def __init__(self, topic, consumer_params, poll_interval='1s', npartitions=1,
-                 ensure_io_loop=True, **kwargs):
+    def __init__(self, topic, consumer_params, poll_interval='1s',
+                 npartitions=1, **kwargs):
         self.consumer_params = consumer_params
         self.topic = topic
         self.npartitions = npartitions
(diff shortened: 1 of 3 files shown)


martindurant:
(sorry about the commit history here, I'll clean it up)

martindurant:
OK, that didn't work. I'll remove the last two commits (I think things are already better in commit bade3e4 ) , but we should release confluent_kafka anyway https://github.com/confluentinc/confluent-kafka-python/releases/tag/v0.11.4
@mrocklin , ok to push that? There still isn't an API for listing partitions/topics, but if it comes, we can always do it again.

mrocklin:
Do you mean create a new package for the newer version on conda-forge?

martindurant:
The latter

mrocklin:
I think it's always a good idea to release packages on conda-forge.  While
doing so we might also inform the upstream confluent maintainers and ask if
they're interested in taking this on in the future.

martindurant:
Yes agreed. Progress on the repo seems to be slow, though, this being a commercial invonlvement.

martindurant:
So I recommend merging this and updating the feedstock, but talking with confluent seems like a good idea too.

streamz/sources.py
```diff
@@ -221,21 +221,26 @@ def start(self):
             def close(ref):
                 ob = ref()
                 if ob is not None and ob.consumer is not None:
-                    ob.consumer.close()
+                    consumer = ob.consumer
+                    ob.consumer = None
+                    consumer.unsubscribe()
+                    consumer.close()  # may raise with latest ck, that's OK
```
mrocklin: Can you briefly explain a bit more about what is going on here?  What was failing before?  Should any of this go within finalize calls or try/finally blocks?  Should we be using thread locks around any of this?

mrocklin: > Is it safe mixing a thread lock with a finalize call? I'm not sure where you are suggesting putting in a try.

I didn't mean to suggest anything concrete, I was just listing various mechanisms that are sometimes used in situations like this.  

OK, so I understand now why we set `ob.consumer = None` when we close.  

Under what circumstances does this raise?  Should we be catching those?

streamz/sources.py
```diff
@@ -221,21 +221,26 @@ def start(self):
...
             self.consumer = None
+            consumer.unsubscribe()
+            consumer.close()
         self.stopped = True
 
 
 class FromKafkaBatched(Stream):
     """Base class for both local and cluster-based batched kafka processing"""
-    def __init__(self, topic, consumer_params, poll_interval='1s', npartitions=1,
-                 ensure_io_loop=True, **kwargs):
```
mrocklin: Nice catch

mrocklin:
> So I recommend merging this and updating the feedstock, but talking with confluent seems like a good idea too.

No objection from me.  I probably won't do anything on this personally for the next week or so.

PR #173, opened 2018-05-02
Disallow combining of DaskStream and Stream objects

### LINKED ISSUE #172: ENH: warn user if combining dask streamz with not dask streamz
Problems seem to occur if a dask stream are combined with not a dask stream (causing futures to not be evaluated when running functions). It might be nice to warn users if they have subscribed a dask stream to a non dask stream without a scatter/gather.

martindurant:
Fixes #172

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -105,11 +105,18 @@ class Stream(object):
 
     def __init__(self, upstream=None, upstreams=None, stream_name=None,
                  loop=None, asynchronous=None, ensure_io_loop=False):
+        from .dask import DaskStream
+
         self.downstreams = OrderedWeakrefSet()
         if upstreams is not None:
             self.upstreams = list(upstreams)
         else:
             self.upstreams = [upstream]
+        if upstreams:
+            if (all(isinstance(u, DaskStream) for u in upstreams) !=
+                    any(isinstance(u, DaskStream) for u in upstreams)):
+                raise ValueError('Must combine only DaskStreams '
+                                 'or no DaskStreams')
 
         self._set_asynchronous(asynchronous)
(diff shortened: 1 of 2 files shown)


martindurant:
This outright errors, which is the extreme of warning. Not sure if you wanted to go that far @CJ-Wright .

CJ-Wright:
I guess one could come up with a very wonky use case which runs operations on the `Futures` objects themselves which would require hooking up dask/not dask streams but that seems like a very janky use case and maybe we shouldn't support unless people demand it.

streamz/core.py
```diff
@@ -105,11 +105,18 @@ class Stream(object):
 
     def __init__(self, upstream=None, upstreams=None, stream_name=None,
                  loop=None, asynchronous=None, ensure_io_loop=False):
+        from .dask import DaskStream
```
mrocklin: This creates a dependency on Dask?

streamz/core.py
```diff
@@ -105,11 +105,18 @@ class Stream(object):
...
         self.downstreams = OrderedWeakrefSet()
         if upstreams is not None:
             self.upstreams = list(upstreams)
         else:
             self.upstreams = [upstream]
+        if upstreams:
+            if (all(isinstance(u, DaskStream) for u in upstreams) !=
+                    any(isinstance(u, DaskStream) for u in upstreams)):
+                raise ValueError('Must combine only DaskStreams '
+                                 'or no DaskStreams')
```
mrocklin: Are there no valid use cases of doing this?

martindurant:
OK, we can close this as overenthusiastic, following comments.

PR #176, opened 2018-05-10
add docs about crossing dask and stream nodes

CJ-Wright:
This closes #173 with docs

### DIFF
--- a/docs/source/dask.rst
+++ b/docs/source/dask.rst
@@ -135,6 +135,6 @@ For example
    a = source.scatter().map(inc)
    b = source2.combine_latest(a)
 
-In this case b is now a ``Stream`` node and does not have access to the actual
+In this case ``b`` is now a ``Stream`` node and does not have access to the actual
 data on the dask cluster.
-Any operations done downstream of b would be performed on the future.
+Any operations done downstream of ``b`` would be performed on the future.
--- a/docs/source/dask.rst
+++ b/docs/source/dask.rst
@@ -117,3 +117,24 @@ did.
     ``source.emit``.
 
 .. _Dask: https://dask.pydata.org/en/latest/
+
+
+Gotchas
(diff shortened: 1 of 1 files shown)


docs/source/dask.rst
```diff
@@ -117,3 +117,24 @@ did.
     ``source.emit``.
 
 .. _Dask: https://dask.pydata.org/en/latest/
+
+
+Gotchas
++++++++
+
+An important gotcha with ``DaskStream`` is that if a ``Stream`` node is
```
martindurant: I would work this differently: 
```
An important gotcha with ``DaskStream`` is that it is a subclass ``Stream``, and so can be used as an input 
to any function expecting a ``Stream``. If there is no intervening ``.gather()``, then the downstream node will
Dask futures instead of the data they represent:

    source = Stream()
    source2 = Stream()
    a = source.scatter().map(inc)
    b = source2.combine_latest(a)

In this case, the combine operation will get real values from ``source2``, and Dask futures. Downstream nodes would be free to operate on the futures, but more likely, the line should be:

    b = source2.combine_latest(a.gather())

martindurant: ( @CJ-Wright )

martindurant:
(I think this will need to merge from master to pass)

martindurant:
Please ignore what I said, seems to work fine and, magically, increases coverage. Perhaps the doc is picked up by doctest?

@mrocklin , happy with the words I suggested?

mrocklin:
Sorry.  I'm slammed right now and not planning to check in on this project
this week.

martindurant:
Understood.

Would you be interested in granting me rights to merge documentation PRs like this and close issues as I see fit? I would *not* merge real code which changes functionality (I know github doesn't differentiate) without your sayso.

PR #181, opened 2018-05-24
FIX: connect doesn't break zip now

CJ-Wright:


CJ-Wright:
it seems that kafka is breaking this

martindurant:
This looks like a good one to resubmit.

PR #182, opened 2018-05-26
Make streamz run tests like distributed?

mariusvniekerk:
In distributed we install dask from github master when running tests.  Should we do the same here?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
   # Install dependencies
   - conda create -n test-streamz python=$TRAVIS_PYTHON_VERSION pytest tornado toolz flake8 coverage codecov networkx graphviz python-graphviz dask distributed pandas bokeh ipython ipykernel ipywidgets python-confluent-kafka -c conda-forge -c bokeh
   - source activate test-streamz
+  - pip install git+https://github.com/dask/dask.git --upgrade --no-deps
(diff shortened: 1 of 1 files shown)


mariusvniekerk:
This seems to pass more tests but fail on some of the kafka ones?

martindurant:
The kafka tests have proven quite annoying to make work, with timing issues over when the server comes online and what data is held by it between one test and the next. I thought I had finally a foolproof setup, but maybe not.

mariusvniekerk:
Guess we must might need some more sleep bits on the Kafka launch :(

martindurant:
Well yes - the tests succeed locally, to be sure (or at least did when I last looked). However, nudging timeout values around is never a nice way to try to solve things.

mariusvniekerk:
we might want to consider using https://github.com/box/flaky ? handy utility for dealing with flaky tests

martindurant:
Recently, I have reworked the kafka tests and all seems to be good. Is there still useful work here, @mariusvniekerk ?

martindurant:
I see it's only a trivial change. I am ambivalent on whether to bother or not, the case where streamz needs something from distributed master will be very rare. Recommend closing.

mariusvniekerk:
Probably fine these days

PR #185, opened 2018-06-15
ENH: combine latest and zip latest have dashed lines to downstreams

CJ-Wright:
which they emit on

### DIFF
--- a/streamz/graph.py
+++ b/streamz/graph.py
@@ -5,6 +5,8 @@ from functools import partial
 import os
 import re
 
+from .core import combine_latest
+
 
 def _clean_text(text, match=None):
     ''' Clean text, remove forbidden characters.
@@ -29,6 +31,7 @@ def create_graph(node, graph, prior_node=None, pc=None):
     node: Stream instance
     graph: networkx.DiGraph instance
     """
+    edge_kwargs = {}
     if node is None:
         return
     t = hash(node)
@@ -40,6 +43,11 @@ def create_graph(node, graph, prior_node=None, pc=None):
                    fillcolor=node._graphviz_fillcolor)
     if prior_node:
         tt = hash(prior_node)
+        # If we emit on something other than all the upstreams vis it
+        if (isinstance(node, combine_latest)
+                and node.emit_on != node.upstreams
+                and prior_node in node.emit_on):
+            edge_kwargs['style'] = 'dashed'
         if graph.has_edge(t, tt):
             return
         if pc == 'downstream':
@@ -62,6 +70,7 @@ def create_edge_label_graph(node, graph, prior_node=None, pc=None, i=None):
     node: Stream instance
     graph: networkx.DiGraph instance
     """
+    edge_kwargs = {}
     if node is None:
         return
     t = hash(node)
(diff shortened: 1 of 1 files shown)


streamz/graph.py
```diff
@@ -40,6 +43,11 @@ def create_graph(node, graph, prior_node=None, pc=None):
                    fillcolor=node._graphviz_fillcolor)
     if prior_node:
         tt = hash(prior_node)
+        # If we emit on something other than all the upstreams vis it
+        if (isinstance(node, combine_latest)
+                and node.emit_on != node.upstreams
+                and prior_node in node.emit_on):
+            edge_kwargs['style'] = 'dashed'
```
mrocklin: This check seems too specific.  Ideally the visualization code wouldn't special case stream types.  Instead, I wonder if you can handle this by adding more configuration options like `_graphviz_shape` or, better yet, generalize that to be some general object like

```python
class combine_latest(object):
    _graphviz_node = {'shape': ellipse', 'style': 'rounded,filled', ...}
    _graphviz_edge = {...}
```

(I haven't thought through much how that would work, but something like that might be a bit cleaner and avoid mixing styles in with the graphviz code.

CJ-Wright: Hmm, ok. My main issue with that for this use case is that we don't really have the concept of a graph edge having properties in streamz. Most of the edge properties (eg if it is the emit on connection) are managed by the nodes themselves. Maybe I can create/edit the `_graphviz_edge` attribute at init/update_upstreams time so that the proper number of edges are dashed.

Maybe I can do 
```python
_graphviz_edge = {downstream: {'style': 'normal'} for downstream in downstreams}}
for downstream in downstreams:
    if downstream in emit_on:
        _graphviz_edge[downstream]['style'] = 'dashed'
```
then
```python
graph.add_edge(tt, t, **node._graphviz_edge[node])
```

mrocklin:
Perhaps nodes can state preferences about how their edges respond?
_graphviz_upstream_edge = {...} and _graphviz_downstream_edge = {...}?
Preferably none of this would require runtime logic.

CJ-Wright:
How would we know which ones are emited on and not?

mrocklin:
Ah I see, I wasn't aware of that specific feature of `combine_latest`.  I guess I'll relax  my suggestion and simply say that it might not be maintainable long term for many different stream types to introduce special cases into the graphviz code.  This seems like the kind of thing that is nice to look at short term, but cements the implementation.

I don't have strong thoughts on how to improve the situation.

mrocklin:
Understood the applicaiton.  Thanks for the explanation.

Do you understand my desired constraint on not adding special cases within the rest of the codebase for this special behavior of combine_latest?

mrocklin:
I wonder if there is some simple and general way that would allow node types with special cases to define how they want the graph around them to look.

CJ-Wright:
Yes, I understand the concern.

That is what I was trying to do [above](https://github.com/mrocklin/streamz/pull/185#discussion_r195904257) since that would allow any node to declare its node edges as needed (and moves the special case from the graph generation code to the node code itself). This exposes a general API to the graph generation code, since it can then just look at the `_graphviz_edge` attribute for the node.

martindurant:
This PR seems to have stalled. 
There is an idea above of how this might be implemented. Honestly I don't know, but if that's not to be submitted in this PR or a new one, I would recommend closing here.

CJ-Wright:
I think I can put together a better PR for graph stuff

martindurant:
OK, so let's close this one.

PR #193, opened 2018-08-08
ENH: live plot of node status

CJ-Wright:
This produces a live plot based on `grave` of the node status: running, awaiting data, failed.


mrocklin:
This seems to depend on libraries like matplotlib and grave.  It should
probably live externally to the core library.

CJ-Wright:
Ok

PR #197, opened 2018-08-09
ENH: Thread Executor based parallel streams

CJ-Wright:
This is an `ThreadPoolExecutor` driven stream system similar to the `DaskStream` system.

Note that this is mostly a proof of concept, the docs need to be fixed to remove references to Dask.

Thinks that could use some help:
1. Making the future handling pickle friendly, currently we decorate all the functions which are passed into the executor so that they properly handle the futures but that makes the functions not pickleable.
1. Checking the correct tornado behavior.
1. Better handling of the `default_client` which should most likely either return a Executor singleton or make it if it doesn't exist.

As requested :smile: 
@mrocklin

### DIFF
--- a/streamz/__init__.py
+++ b/streamz/__init__.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 from .core import *
 from .graph import *
 from .sources import *
+from .thread import *
 try:
     from .dask import DaskStream, scatter
 except ImportError:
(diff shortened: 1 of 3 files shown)


CJ-Wright:
CC: @mariusvniekerk

CJ-Wright:
see: https://github.com/xpdAcq/streamz_ext/blob/master/streamz_ext/parallel.py

PR #210, opened 2018-10-03
Replaced sync func with updated distributed/utils.py func

### LINKED ISSUE #209: Running Streamz with Dask.Distributed
I have a dask.distributed cluster setup with workers spread across multiple nodes and I am testing out a simple example to stream data to the workers to be processed:

    import dask.distributed
    import streamz

    client = dask.distributed.Client('tcp://localhost:8786')
    # client.get_metadata  # prints: <bound method Client.get_metadata of <Client: scheduler='tcp://<some ip address>:8786' processes=6 cores=96>>

    def increment(x):
        return x+1

    src = streamz.Stream()
    (src.scatter()
        .map(increment)
    )

    for x in range(10):
        src.emit(x)

However, I am receiving the following error:

    Traceback (most recent call last):
      File "./test.py", line 18, in <module>
        src.emit(x)
      File "/home/miniconda3/lib/python3.6/site-packages/streamz/core.py", line 259, in emit
        return sync(self.loop, _)
      File "/home/miniconda3/lib/python3.6/site-packages/streamz/core.py", line 1177, in sync
        if not loop._running:
    AttributeError: 'AsyncIOLoop' object has no attribute '_running'

It isn't entirely clear from the documentation what one needs to do to get around this or if this is a possible bug.

seanlaw:
close #209

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1231,14 +1231,13 @@ def sync(loop, func, *args, **kwargs):
     Run coroutine in loop running in separate thread.
     """
     # This was taken from distrbuted/utils.py
-    timeout = kwargs.pop('callback_timeout', None)
 
-    def make_coro():
-        coro = gen.maybe_future(func(*args, **kwargs))
-        if timeout is None:
-            return coro
-        else:
-            return gen.with_timeout(timedelta(seconds=timeout), coro)
+    # Tornado's PollIOLoop doesn't raise when using closed, do it ourselves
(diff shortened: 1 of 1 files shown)


mrocklin:
Looks like this wasn't as easy as I had thought :/

seanlaw:
I tried :)

mrocklin:
I appreciate you trying.  It looks like this will require more investigation.  It could also be that any number of other things changed in the interim.

PR #216, opened 2019-01-09
Added the to_kafka stream

jsmaupin:
I have added the to_kafka class.

The options were to either: 

1. Call the `.flush()` method after each `.produce(...)` call. This will seriously hurt performance as messages will be sent serially.
2. Use `atexit` to call flush at the end.  This requires the use of `atexit`, of which I'm unsure of any impact. It also means that the result of the `.emit(...)` call by the client does not yield an immediate call to the downstream. Note that the Confluent library will flush when the batch size reaches a threshold.  This is the method that is implemented here.

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1295,6 +1295,31 @@ class to_kafka(Stream):
         self.producer.flush(timeout)
 
 
+@Stream.register_api()
+class to_kafka_batched(Stream):
+    def __init__(self, upstream, topic, producer_config, n, **kwargs):
+        import confluent_kafka as ck
+
+        self.topic = topic
+        self.producer = ck.Producer(producer_config)
+        self.n = n
+        self.buffer = []
+
+        Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
+
+    def update(self, x, who=None):
+        self.buffer.append(x)
+        if len(self.buffer) >= self.n:
+            self.flush(False)
+
+    def flush(self, producer=True):
+        for x in self.buffer:
+            self.producer.produce(self.topic, x)
+        self.buffer = []
+        if producer:
+            self.producer.flush()
+
+
 def sync(loop, func, *args, **kwargs):
     """
     Run coroutine in loop running in separate thread.
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import atexit
 from collections import deque
 from datetime import timedelta
 import functools
@@ -15,6 +16,10 @@ from tornado import gen
 from tornado.locks import Condition
 from tornado.ioloop import IOLoop
 from tornado.queues import Queue
+try:
+    import confluent_kafka as ck
+except ImportError:
+    ck = None  # confluent_kafka is only required when interacting with Kafka
 try:
     from tornado.ioloop import PollIOLoop
 except ImportError:
@@ -1229,6 +1234,54 @@ class latest(Stream):
             yield self.condition.wait()
             [x] = self.next
             yield self._emit(x)
+
+
+@Stream.register_api()
+class to_kafka(Stream):
+    """ Writes data in the stream to Kafka
+
+    The Confluent client sends messages in batches and asynchronously. When a
+    message is received, or in the event of an error, this stream will emit
+    a tuple in the form of (err, msg) see
+    https://github.com/confluentinc/confluent-kafka-python
+
+    Parameters
+    ----------
+    topic : string
+        The topic which to write
+    producer_config : dict
+        Settings to set up the stream, see
+        https://docs.confluent.io/current/clients/confluent-kafka-python/#configuration
+        https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
+        Examples:
+        bootstrap.servers: Connection string(s) (host:port) by which to reach Kafka
+
+    Examples
+    --------
+    >>> source = Stream()
+    >>> source.map(lambda x: x + 1).to_kafka('test', {'bootstrap.servers': 'localhost'}).sink(
+    ...     lambda x: print(x[1].topic(), x[1].partition()))
+    >>> for i in range(3):
+    ...     source.emit(i)
+    test 0
+    test 0
+    test 0
+    """
+    def __init__(self, upstream, topic, producer_config, **kwargs):
+        self.topic = topic
+        self.producer = ck.Producer(producer_config)
+        atexit.register(self.producer.flush)
+
+        Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
+
+        _global_sinks.add(self)
+
+    def update(self, x, who=None):
+        self.producer.poll(0)
+        self.producer.produce(self.topic, str(x).encode('utf-8'), callback=self.cb)
+
+    def cb(self, err, msg):
+        self.emit((err, msg))
 
 
 def sync(loop, func, *args, **kwargs):
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1274,8 +1274,6 @@ class to_kafka(Stream):
 
         Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
 
-        _global_sinks.add(self)
-
     def update(self, x, who=None):
         self.producer.poll(0)
         self.producer.produce(self.topic, str(x).encode('utf-8'), callback=self.cb)
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-from collections import deque, namedtuple
+from collections import deque
 from datetime import timedelta
 import functools
 import logging
@@ -1254,18 +1254,16 @@ class to_kafka(Stream):
 
     Examples
     --------
+    >>> from streamz import Stream
     >>> ARGS = {'bootstrap.servers': 'localhost:9092'}
     >>> source = Stream()
-    >>> kafka = source.map(lambda x: x + 1).to_kafka('test', ARGS)
-    >>> kafka.sink(lambda x: print(x[1].topic(), x[1].partition()))
-    >>> for i in range(3):
-    ...     source.emit(i)
-    >>> kafka.flush()
-    test 0
-    test 0
-    test 0
+    >>> source.map(lambda x: str(x)).to_kafka('test', ARGS)
+    <to_kafka>
+    >>> for i in range(10):
+    ...     # In script simply use await keyword
+    ...     asyncio.wait(source.emit(i, asynchronous=True))
+
     """
-    Result = namedtuple('Result', ['err', 'msg'])
 
     def __init__(self, upstream, topic, producer_config, **kwargs):
         import confluent_kafka as ck
@@ -1275,15 +1273,30 @@ class to_kafka(Stream):
 
         Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
 
+        self.loop.add_callback(self.cb)
+
     def update(self, x, who=None):
-        self.producer.poll(0)
-        self.producer.produce(self.topic, x, callback=self.cb)
+        future = gen.Future()
+        loop = self.loop.current()
+
+        def cb(err, msg):
+            if err:
+                exc = Exception((err, msg.value() if msg else None))
+                loop.add_callback(future.set_exception, exc)
+            else:
+                loop.add_callback(future.set_result, None)
+
+        self.producer.produce(self.topic, x, callback=cb)
+
+        return future
+
+    @gen.coroutine
+    def cb(self):
+        while True:
+            self.producer.poll(65500)
 
     def flush(self, timeout=-1):
         self.producer.flush(timeout)
-
-    def cb(self, err, msg):
-        self.emit(self.Result(err, msg.value() if msg else None))
 
 
 def sync(loop, func, *args, **kwargs):
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-import atexit
-from collections import deque
+from collections import deque, namedtuple
 from datetime import timedelta
 import functools
 import logging
@@ -1236,9 +1235,10 @@ class latest(Stream):
 class to_kafka(Stream):
     """ Writes data in the stream to Kafka
 
-    The Confluent client sends messages in batches and asynchronously. When a
-    message is received, or in the event of an error, this stream will emit
-    a tuple in the form of (err, msg) see
+    The Confluent client sends messages in batches and asynchronously. This
+    stream accepts a string or bytes object. Call ``flush`` to ensure all
+    messages are pushed. When a message is received, or in the event of an
+    error, this stream will emit a tuple in the form of (err, msg) see
     https://github.com/confluentinc/confluent-kafka-python
 
     Parameters
@@ -1263,21 +1263,25 @@ class to_kafka(Stream):
     test 0
     test 0
     """
+    Result = namedtuple('Result', ['err', 'msg'])
+
     def __init__(self, upstream, topic, producer_config, **kwargs):
         import confluent_kafka as ck
 
         self.topic = topic
         self.producer = ck.Producer(producer_config)
-        atexit.register(self.producer.flush)
 
         Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
 
     def update(self, x, who=None):
         self.producer.poll(0)
-        self.producer.produce(self.topic, str(x).encode('utf-8'), callback=self.cb)
+        self.producer.produce(self.topic, x, callback=self.cb)
+
+    def flush(self, timeout=-1):
+        self.producer.flush(timeout)
 
     def cb(self, err, msg):
-        self.emit((err, msg))
+        self.emit(self.Result(err, msg.value() if msg else None))
 
 
 def sync(loop, func, *args, **kwargs):
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1254,11 +1254,13 @@ class to_kafka(Stream):
 
     Examples
     --------
+    >>> ARGS = {'bootstrap.servers': 'localhost:9092'}
     >>> source = Stream()
-    >>> source.map(lambda x: x + 1).to_kafka('test', {'bootstrap.servers': 'localhost'}).sink(
-    ...     lambda x: print(x[1].topic(), x[1].partition()))
+    >>> kafka = source.map(lambda x: x + 1).to_kafka('test', ARGS)
+    >>> kafka.sink(lambda x: print(x[1].topic(), x[1].partition()))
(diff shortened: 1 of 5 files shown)


mrocklin:
cc @martindurant you might find this interesting

martindurant:
I do! I'll try to look at it in the next couple of days.

martindurant:
Is someone maintaining this project? The errors in the tests are not obviously caused by the new code, and there doesn't seem to have been much activity in several months.

CJ-Wright:
I think if pytest is pinned to 3.10 this might be ok.

mrocklin:
> Is someone maintaining this project? The errors in the tests are not obviously caused by the new code, and there doesn't seem to have been much activity in several months.

No

mrocklin:
Or at least, I'm not.  Apparently no one else is either?

mrocklin:
Resolved the pytest things here: https://github.com/mrocklin/streamz/pull/217

mrocklin:
I've resolved the pytest failures and restarted tests here.  It looks like there are kafka failures in master.  @martindurant I suspect that you are the most familiar with that testing infrastructure (although @jsmaupin if you have energy to spend it would be good to get your eyes on the kafka testing code as well as we'll need that to write tests for your work here as well)

jsmaupin:
Sure, I'll take a look.

streamz/core.py
```diff
@@ -1231,6 +1236,52 @@ def cb(self):
...
+    def __init__(self, upstream, topic, producer_config, **kwargs):
+        self.topic = topic
+        self.producer = ck.Producer(producer_config)
+        atexit.register(self.producer.flush)
+
+        Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
+
+    def update(self, x, who=None):
+        self.producer.poll(0)
+        self.producer.produce(self.topic, str(x).encode('utf-8'), callback=self.cb)
```
kkraus14: I'm not sure casting to string is desired here, I believe the end user should create an object that is either bytes-like or str-like to be able to send and you should just send the object. The string representation of an object could be drastically different than the byte representation of the object.

martindurant: You probably do want to ensure that the thing passed is bytes-like, I'm not sure I would trust confluent's library not to segfault... (perhaps make a test for something not bytes-like to make sure it behaves?)

jsmaupin: So, if the user is manipulating text data, then they have to to have a stage upstream that will do the bytes encoding?

streamz/core.py
```diff
@@ -15,6 +16,10 @@
 from tornado.locks import Condition
 from tornado.ioloop import IOLoop
 from tornado.queues import Queue
+try:
+    import confluent_kafka as ck
```
martindurant: Won't this just make for an odd error for anyone trying to use the kafka stream without confluent? Why not just import where needed in the methods?

jsmaupin: When I looked at the Kafka code in the sources.py file, I see that the Confluent library is imported locally.  I assumed that there must have been some desire to only make this library dependent on the Confluent library only if the user was going to communicate with Kafka.  So, I did not make it mandatory.

martindurant:
The kafka-threads are the ones that appear to error (they first fail, timeout, and then don't gracefully end), so something with the event loop deadlocking, probably. All pass locally?

Note that I don't see any new tests for the functionality in this PR.

streamz/core.py
martindurant: Either way it wouldn't be mandatory, but in your version, you would get something like "None has no attribute..." error message when trying to use it, which would be confusing.

jsmaupin: Okay, I can make the import local to the to_kafka class.

streamz/core.py
jsmaupin: This change seems to work.

https://github.com/jsmaupin/streamz/commit/893300f55e8d016b91e9770471cc1109dd34a742

streamz/core.py
jsmaupin: This change seems to work.

https://github.com/jsmaupin/streamz/commit/893300f55e8d016b91e9770471cc1109dd34a742

martindurant:
Happy to see this fixed! You seem only to be failing on flake lint (lots of it, presumably not added by you).

mrocklin:
> You seem only to be failing on flake lint (lots of it, presumably not added by you).

If it's unrelated to this work then I'll take care of it later.

mrocklin:
@martindurant thanks for reviewing this.  Do you have any other concerns about the content of the implementation?

mrocklin:
I'll try to take a closer look at this later today

streamz/core.py
```diff
@@ -1231,6 +1231,59 @@ def cb(self):
...
+    def __init__(self, upstream, topic, producer_config, **kwargs):
+        import confluent_kafka as ck
+
+        self.topic = topic
+        self.producer = ck.Producer(producer_config)
+
+        Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
+
+    def update(self, x, who=None):
+        self.producer.poll(0)
```
martindurant: Is this necessary? If the data is not sent strictly synchronously, I think that's OK.

streamz/core.py
```diff
@@ -1231,6 +1231,59 @@ def cb(self):
...
+        self.producer.produce(self.topic, x, callback=self.cb)
+
+    def flush(self, timeout=-1):
+        self.producer.flush(timeout)
+
+    def cb(self, err, msg):
```
martindurant: I wonder if this should be a sink, and not emit anything? Of course, I see the usefulness in testing.

streamz/core.py
jsmaupin: I have updated the PR.  The Confluent library will take in a string or bytes, so Streamz will pass in the user provided value directly to the Confluent library

streamz/core.py
jsmaupin: Updated to the above comment in the latest PR.

jsmaupin:
I have managed to fix most of the tests.  I added another test for the new .from_kafka() class.  The main issue is described here from the author of the Confluent Python library: https://github.com/confluentinc/confluent-kafka-python/issues/156#issuecomment-286235198. 

Basically, the .subscribe() method does not immediately subscribe. There is an asynchronous operation it must complete. Subscribing and then immediately producing messages from the same client is not a normal use-case.  Most of the time, the subscriber and producer are on different locations on a network. I have added a _blocking_subscribe method that resolves this issue. We did not see this in the non-threading tests because they run first and the tests use the auto-create-topic functionality in Kafka.  There is a slight delay when producing the first message while Kafka creates the topic. There was enough time here for the .subscribe() message to completely subscribe. Once the topic was created, the rest of tests failed.

I have not been able to get the test_kafka_dask_batch test to pass.

jsmaupin:
On second thought, if this is not a normal use-case, perhaps I should move the blocking subscribe functionality to the tests?

streamz/core.py
mrocklin: I'm glad to hear it.  Thanks for tracking this down @jsmaupin

streamz/core.py
mrocklin: I'm also curious about this, and more generally how well this approach works in an asynchronous environment.  It has been a while since I looked at confluent_kafka_python so I've forgotten some things

If the producer is busy then what happens with `producer.poll(0)`?  Does it err, does it block?  It may be that neither of these outcomes is desirable.

mrocklin:
There is some discussion around `tornado`,  `confluent_kafka_python`, and async programming in https://github.com/confluentinc/confluent-kafka-python/issues/100  that might be an interesting read

mrocklin:
@jsmaupin what are your thoughts on parallel and partitioned writes?  My guess is that you wanted to start with the simple sequential case before going to the batched/parallel case.  Is this correct?

martindurant:
btw: all bar one of the flake8 issues are in test modules - can those just be skipped?

mrocklin:
They could be, but it's also easy for someone to go through and fix them.
I have it on my todo list and will get to it eventually.  If anyone wants
to beat me to it in a separate PR that would be welcome.

martindurant:
(merging from #219 will make the CI here pass)

mrocklin:
I suspect that one solution to solve the backpressure issue looks like this:

```python
async def update(self, x, who=None):
    future = gen.Future()
    loop = self.loop

    def cb(err, msg):
        if err:
            exc = Exception((err, msg.value() if msg else None))
            loop.add_callback(future.set_exception, exc)
            # future.set_exception(exc)  # want to do this, but it's not threadsafe
        else:
            loop.add_callback(future.set_result, None)
            future.set_result(None)  # want to do this, but it's not threadsafe

    self.producer.produce(self.topic, x, callback=cb)
    return future
```

This creates a Future object and passes that back to the caller.  Eventually that future will percolate back to the user and they'll `await` on it and feel pressure if kafka becomes slow.

When Kafka has finished the write it will call the local `cb` function, which will assign a result or exception to the future as appropriate.  Anyone awaiting that future will be woken up and can start work again.  

I've had to be careful above with `loop.add_callback` because I suspect that kafka's produce method will call the callback in a separate thread, and the `loop.add_callback` method is the only threadsafe way to call into the event loop.

I hope that this helps a bit.  I wouldn't expect the code above to work verbatim, but hopefully it puts folks here in the right direction.

jsmaupin:
@mrocklin, I spent a good time digging into this to make sure I understand everything. I'm confident I do now.  The only way I could get this to work was to make two major changes to your suggestion.  First, it seems to me that the reason you save off the loop is to make sure you have the `IOLoop` for the main thread. This was not working for me unless I used `self.loop.current()`.  Also, as an aside, it seems the Confluent library calls the callback from the thread on which `.flush()` was called.  So, it could easily be called from another thread. Second, if I made no other changes, the first message would be sent, but everything would pause at `await stream.emit(i)` on my client code which looked like this: 
```python
    ARGS = {'bootstrap.servers': 'localhost:9092'}
    source = Stream()
    kafka = source.map(
        lambda x: str(x) + ', ' + datetime.now().isoformat()).to_kafka('test',
                                                                       ARGS)
    kafka.sink(lambda x: print(x))

    start = time()
    for i in range(5000):
        await source.emit(i, True)

    print(str(time() - start) + 's')
```
`.poll(...)` must be called, or the callbacks from the Confluent library will not be triggered. So, my current solution is to add `self.producer.produce(self.topic, x, callback=cb)` in the constructor and have the cb function look like this:
```python
@gen.coroutine
    def cb(self):
        while True:
            self.producer.poll(65500)
```
For some reason, a value much larger than that in `.poll(...)` caused it to never block even though it should take a 32 bit floating point value.

Performance-wise, this takes about 8.1 seconds to send 5000 messages. So, it's quite slow, but the batched performance should be much better. What do you think?

jsmaupin:
Here are my tests results for various implementations.  For each I used 10,000 records.
**Outside of the streams library**
for loop: 0.039s

**to_kafka**
Awaiting on the callback return from Kafka: 13.2s
Using the BufferError exception to handle back-pressure: 7.47s

**to_kafka_batched**
With just `pass` in the `update` method: 2.15s
Flushing to Kafka (batch size 500): 2.2s

martindurant:
@jsmaupin , any objection if I push flake8 fixes here? Everything else seems to be in order, and other PRs are failing while this is unmerged.

jsmaupin:
No objections here.

martindurant:
Ah, sorry, I can't - I'm not a `streamz` maintainer. The following diff fixes things:
```diff
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -942,7 +942,7 @@ def dont_test_stream_kwargs(clean):


 @pytest.fixture
-def thread(loop):
+def thread(loop):   # noqa: E811
     from threading import Thread, Event
     thread = Thread(target=loop.start)
     thread.daemon = True
diff --git a/streamz/tests/test_dask.py b/streamz/tests/test_dask.py
index 5c007a3..9a415fd 100644
--- a/streamz/tests/test_dask.py
+++ b/streamz/tests/test_dask.py
@@ -74,7 +74,7 @@ def test_zip(c, s, a, b):


 @pytest.mark.slow
-def test_sync(loop):
+def test_sync(loop):  # noqa: E811
     with cluster() as (s, [a, b]):
         with Client(s['address'], loop=loop) as client:  # flake8: noqa
             source = Stream()
@@ -91,7 +91,7 @@ def test_sync(loop):


 @pytest.mark.slow
-def test_sync_2(loop):
+def test_sync_2(loop):  # noqa: E811
     with cluster() as (s, [a, b]):
         with Client(s['address'], loop=loop):  # flake8: noqa
             source = Stream()
@@ -132,7 +132,7 @@ def test_buffer(c, s, a, b):


 @pytest.mark.slow
-def test_buffer_sync(loop):
+def test_buffer_sync(loop):  # noqa: E811
     with cluster() as (s, [a, b]):
         with Client(s['address'], loop=loop) as c:  # flake8: noqa
             source = Stream()
@@ -158,7 +158,7 @@ def test_buffer_sync(loop):

 @pytest.mark.xfail(reason='')
 @pytest.mark.slow
-def test_stream_shares_client_loop(loop):
+def test_stream_shares_client_loop(loop):  # noqa: E811
     with cluster() as (s, [a, b]):
         with Client(s['address'], loop=loop) as client:  # flake8: noqa
             source = Stream()
```

martindurant:
(if you don't want to deal with this, @jsmaupin , happy to push to my fork instead)

martindurant:
Sorry, apparently it takes a little more work!

jsmaupin:
Yeah, np.  It looks like one version had the noqa on the decorators and yours was on the function itself.

martindurant:
It seems the other one was right! Ooops...
```
streamz/tests/test_kafka.py:148:5: E999 SyntaxError: invalid syntax
streamz/tests/test_core.py:946:12: F811 redefinition of unused 'loop' from line 23
streamz/tests/test_dask.py:76:18: W291 trailing whitespace
streamz/tests/test_dask.py:94:17: F811 redefinition of unused 'loop' from line 14
streamz/tests/test_dask.py:135:22: F811 redefinition of unused 'loop' from line 14
streamz/tests/test_dask.py:160:36: F811 redefinition of unused 'loop' from line 14
```
(plus some whitespace has crept in)

jsmaupin:
Some E811 codes should have been F811. I tested locally this time.

martindurant:
Strangely, seeing syntax error now at https://github.com/mrocklin/streamz/pull/216/files#diff-a56707adbd2d2392ea8625521ed6edd0R156 - should be tornado gen/yield pattern? Otherwise, should be in py3_test_core. But this was passing all tests before, so what happened?

jsmaupin:
There was some Python3 specific code in there.  This should be a working commit.

streamz/tests/test_kafka.py
```diff
@@ -202,6 +204,7 @@ def test_kafka_batch():
         stream = Stream.from_kafka_batched(TOPIC, ARGS)
         out = stream.sink_to_list()
         stream.start()
+        sleep(2)
```
skmatti: This may cause this test to fail, because default poll_interval is 1 sec and the test would have polled 2 batches. The second batch would be empty list and out[-1][-1] would be null always.

martindurant: I'm not sure about that, I think the time accounts for the initial setting up of the consumer, and no data has been sent yet at this point.

martindurant:
Green! Let's merge this!

martindurant:
(^ @mrocklin )

streamz/tests/test_kafka.py
jsmaupin: Yes, if I had another way to do this without a sleep, I think that would be preferable, but the delay for the topic creation is why it is there.

skmatti: Yeah, you are right! Data hasn't been sent yet.

jsmaupin:
@mrocklin, let me know if there's anything I can do to help get this closed

streamz/core.py
```diff
@@ -1231,6 +1231,95 @@ def cb(self):
...
+                try:
+                    self.producer.produce(self.topic, x, callback=self.cb)
+                    future.set_result(None)
+                    return
+                except BufferError:
+                    pass
+                except Exception as e:
+                    future.set_exception(e)
+
+        self.loop.add_callback(_)
```
mrocklin: This function above, is it likely to take a non-trivial amount of time?  If so then we probably can't place it onto the event loop.

streamz/tests/test_kafka.py
```diff
@@ -122,6 +122,7 @@ def test_from_kafka():
         stream = Stream.from_kafka([TOPIC], ARGS, asynchronous=True)
         out = stream.sink_to_list()
         stream.start()
+        sleep(5)
```
mrocklin: Why this sleep?

streamz/tests/test_kafka.py
```diff
@@ -152,6 +171,7 @@ def test_from_kafka_thread():
         stream = Stream.from_kafka([TOPIC], ARGS)
         out = stream.sink_to_list()
         stream.start()
+        sleep(2)
```
mrocklin: Why this sleep?  In general sleeps can be troublesome because:

1.  They make tests take a long time if they are overly conservative
2.  If the system running the tests is abnormally slow then they can fail (Travis is often abnormally slow)

I'm also concerned about using `time.sleep` here in an async test.  This is usually an anti-pattern.

streamz/core.py
```diff
@@ -1231,6 +1231,95 @@ def cb(self):
...
+        self.producer = ck.Producer(producer_config)
+
+        Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
+
+    def update(self, x, who=None):
+        future = gen.Future()
+
+        def _():
+            while True:
+                self.producer.poll(0)
```
mrocklin: What happens if we poll and the producer is still busy?  How do we handle back pressure?

streamz/tests/test_kafka.py
jsmaupin: This can probably be removed. I'll take a closer look

streamz/tests/test_kafka.py
jsmaupin: Yes, I totally agree. I'll take some time to come up with a better solution.

streamz/core.py
jsmaupin: The idea is that this while loop will not exit if the producer queue is full.  Am I breaking something in here with regards to how async' systems work? Will this block other parts of the program?

jsmaupin:
@mrocklin thank you much for the feedback.

martindurant:
@jsmaupin , I have extended your work and implemented Matt's suggestions in #230 . I believe that that PR is ready, and there should be no need for you to add anything here. The only thing that doesn't make the cut is to_kafka_batched, which can wait for a future PR.

jsmaupin:
So, I can close this PR?

martindurant:
I believe so. @mrocklin , are we close to merging #230 ?

martindurant:
This PR should now be closed

martindurant:
Thanks @jsmaupin for your work here.

PR #217, opened 2019-01-09
Update for use of pytest 4

mrocklin:


mrocklin:
I'm going to guess that the kafka failures here are unrelated.  I'm going to merge this in.

cc @martindurant in case he's interested in the kafka failures.

PR #219, opened 2019-01-15
Fix flake things

martindurant:


mrocklin:
Thanks for doing this @martindurant 

It looks like the kafka changes snuck in here

martindurant:
> It looks like the kafka changes snuck in here

That was on purpose, as kafka tests were not otherwise passing - we could merge this, if the original PR is OK, or that PR could cherry-pick from here.

mrocklin:
It seems like there are still questions on that PR around backpressure.  I
recommend that we just merge in the flake8 changes here and leave the kafka
PR until that solidifies.

martindurant:
Would you be OK with two of the kafka tests failing, though?

mrocklin:
They're already failing in master, correct?  If so then sure.  I think that they're sufficiently separate problems that they should ideally be handled in separate PRs.

I don't care a ton if you also want to fix the kafka tests and you feel strongly about including that fix in this PR (though I would still prefer that they be in separate PRs).  However it looks like the new `to_kafka` stuff is here, which, due to the backpressure concerns, definitely isn't ready to go in.

martindurant:
OK, done. Expect this to fail.
@jsmaupin , after this is merged and you apply to your branch, tests should go green for you.

mrocklin:
Thanks @martindurant !

PR #223, opened 2019-02-01
Fixed FutureWarning when creating DatetimeIndex

philippjfr:
Using future pandas (>0.24.0) creating a DatetimeIndex using start and endpoints will no longer be supported, which means that the Random streaming dataframe very noisily issues these FutureWarnings:

```
/Users/philippjfr/miniconda/envs/anacondaviz/lib/python3.6/site-packages/streamz/dataframe/core.py:711: FutureWarning: Creating a DatetimeIndex by passing range endpoints is deprecated.  Use `pandas.date_range` instead.
  freq=freq)
```

### DIFF
--- a/streamz/dataframe/core.py
+++ b/streamz/dataframe/core.py
@@ -706,9 +706,8 @@ class WindowedGroupBy(GroupBy):
 
 def _random_df(tup):
     last, now, freq = tup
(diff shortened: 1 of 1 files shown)


mrocklin:
Thanks @philippjfr !

PR #224, opened 2019-02-07
Add cudf support for Streamz DataFrame

skmatti:
This adds some functionalities for streaming cudf DataFrames with Streamz DataFrame. Some of the available methods are listed in here: [mrocklin/streamz/#222](https://github.com/mrocklin/streamz/issues/222)

Most of Streamz DataFrame methods are still available only with pandas and invoking these methods with cudf may result in ambiguous error messages. 

@mrocklin I changed definitions of is_series_like and is_index_like a bit to make them work for both cudf and pandas DataFrames. Not sure if I am doing this correct. 

Also, there are still a lot of references to pandas and cudf modules in Streamz DataFrame implementation. Most of these would go away after adding support for pandas like methods in cudf. I would be happy to get suggestions on removing/decreasing count of these references as we aspire to make it more general.


streamz/dataframe/aggregations.py
```diff
@@ -396,7 +397,7 @@ def grouped(self, df, grouper=None):
 
         g = df.groupby(grouper)
 
-        if self.columns is not None:
+        if self.columns is not None and isinstance(df, pd.DataFrame):
```
mrocklin: Why this change?  Was there some limitation in cudf where this didn't work?

streamz/dataframe/core.py
```diff
@@ -698,16 +733,21 @@ def _accumulate(self, Agg, **kwargs):
...
                                       window=window)
 
-        for typ, s_type in _stream_types[stream_type]:
-            if isinstance(example, typ):
+        for _, s_type in _stream_types[stream_type]:
+            if is_dataframe_like(example):
+                return s_type(outstream, example)
+            elif is_series_like(example):
+                return s_type(outstream, example)
+            elif is_index_like(example):
```
mrocklin: I'm not sure that I understand this code change.  

It looks like we always use the first `s_type` found, regardless of the type of the `example`

streamz/dataframe/core.py
```diff
@@ -3,6 +3,10 @@
 from collections import OrderedDict
 import operator
 from time import time
+try:
+    import cudf
+except ImportError:
+    cudf = None
```
mrocklin: We'll need to find a place to put this.  Either a `backends.py` file within this repository, a streaming.py file in cudf, or some third tiny repository.

streamz/dataframe/core.py
```diff
@@ -266,10 +273,10 @@ class DataFrame(Frame, _DataFrameMixin):
     --------
     Series
     """
-    _subtype = pd.DataFrame
+    _subtype = object
```
mrocklin: If we're going define this in `__init__` then we can probably remove this from the class definition entirely.

streamz/dataframe/core.py
```diff
@@ -314,7 +327,16 @@ class Series(Frame, _SeriesMixin):
...
+
+    def __init__(self, *args, **kwargs):
+        example = kwargs.get('example', None) if "example" in kwargs else args[1]
+        assert example is not None
+        if not is_series_like(example):
+            msg = "Streaming Series expects an example of Series like objects. Got: {}.".format(example)
+            raise TypeError(msg)
+        self._subtype = type(example)
+        super(Series, self).__init__(*args, **kwargs)
```
mrocklin: Maybe we can include the predicate like `is_series_like` on the class and then keep this constructor on a parent class in order to avoid duplication?

streamz/dataframe/aggregations.py
skmatti: Yes, pandas GorupBy class has separate classes for DataFrame and Series. And GroupBy DataFrame is subscriptable. But cudf has only one class and is not subscriptable.

streamz/dataframe/core.py
skmatti: I think it's a mistake, will fix this. I thought I am going to get away with not having to reference pandas and cudf classes explicitly in _stream_types.

streamz/dataframe/core.py
skmatti: Yeah, this seems redundant. I wasn't sure if this variable is referenced before initialization.

streamz/dataframe/core.py
skmatti: But it wouldn't be possible to set _subtype without the constructor. Also, the parent class is Streaming which should not have DataFrame specific code.

streamz/dataframe/aggregations.py
kkraus14: @skmatti rather than adding code to differentiate between Pandas DataFrames and cuDF DataFrames it would be better to just change the cuDF APIs to match Pandas. In situations like these can you just raise an issue on github? Thanks!

skmatti: Yes, makes sense. This PR([/rapids/ai/#895](https://github.com/rapidsai/cudf/pull/895)) may resolve this.

skmatti:
@mrocklin I hand picked a few tests from test_dataframes and modified them to work for cudf. Most of these tests worked by replacing pd.DataFrame with cudf.DataFrame. But I had to change a few things for some tests. I haven't changed anything that I didn't need to.

mrocklin:
In principle these changes look really good to me.  The cudf changes are well isolated and generally smaller than I expected.  Nice work!

We'll probably want the `cudf` import to be either:

1.  Lazily done, as with the `register_lazy` trick used by the Dask Dispatcher objects.  See the following links:
    - https://github.com/dask/dask/blob/master/dask/array/backends.py
    - https://github.com/dask/dask/blob/2376da2aaaa0d40f3680e18b439000c17043b629/dask/utils.py#L351-L380

    This would mean replacing the curent `_stream_types` dict with a Dispatch function

    This is important so that cudf isn't imported whenever a Pandas user wants to use streamz.  The cudf import time is a few seconds at this point, and may have other side effects.  

2.  In a separate `custreamz` library.  This is cleaner from an organizational point of view, but can sometimes provide a confusing user experience.  Users sometimes don't understand why they need to import custreamz when all they ever use is the streamz library.

streamz/dataframe/core.py
```diff
@@ -283,7 +289,13 @@ def concat(tup, columns=None):
                                     for k, v in args[0].items()})
             DataFrame.__init__(self, stream, example)
         else:
-            return super(DataFrame, self).__init__(*args, **kwargs)
+            example = kwargs.get('example', None) if "example" in kwargs else args[1]
+            assert example is not None
```
jsmaupin: Can we raise an AttributeError here instead of asserting?

jsmaupin:
Just a question, would this be cleaner with an oop/inheritance model where, instead of saying `if is_dataframe_like`, the code would be run in a class that had the code to handle the items in the `if` block?

Also, are we not able to run the cudf tests in every build? Could this be achieved by importing cdf?

streamz/dataframe/core.py
skmatti: Handled this in the latest commit

streamz/dataframe/core.py
skmatti: I have used enums in an attempt to get rid of code duplication. Would love to hear your thoughts on this latest approach.

skmatti:
@jsmaupin I merged those `if` blocks into a single method. 
Travis does not support GPU based builds yet. So I thought I would skip those tests on travis CI.

skmatti:
@mrocklin I spent some time looking into lazy import implementation and move cudf import to backends file. Here are my thoughts about handling this issue.

1. I don't think we need a Dispatch function for current implementation, as we are not invoking any cudf functions explicitly using cudf import in `core.py`. We just need a few dataframe like class types to identify the corresponding stream types. This can be achieved by making use of is_dataframe_like, is_series_like and is_index_like functions. But I have to add streamz dataframe specific code in `collection.py` which is undesirable. 

2. In future if we have to access `cudf.concat` or any other cudf functions explicitly in `core.py`, then we need a dispatch function. Also, adding dispatch function makes it more difficult to not have dataframe specific code in `collection.py`.

streamz/dataframe/core.py
martindurant: I recommend the first option. Probably cudf isn't particularly interested in supporting streamz, and it seems overkill to have a whole other repo just for this.

streamz/dataframe/utils.py
```diff
@@ -0,0 +1,38 @@
+from enum import Enum
+
+
+def is_dataframe_like(df):
```
martindurant: Can we replace these with imports from dask.dataframe.utils?

skmatti: Yes, this will be possible with cudf 0.6. Currently, is_series_like fails with cudf. Ref: [rapidsai/cudf/#983](https://github.com/rapidsai/cudf/issues/983)

skmatti: Also, this means non dask streamz DataFrame users are forced to install dask?

martindurant: I suppose that is an issue; I don't suppose there are many non-dask cudf users of streamz... but you are right. A pity to duplicate code, though.

streamz/dataframe/core.py
skmatti: The current implementation does not have to import cudf from Streamz DataFrame. But I needed to modify `collection.py` file outside dataframe folder.

skmatti:
@mrocklin and @martindurant I made the suggested changes. Let me know if I should be make any changes to merge this.

streamz/dataframe/utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+def is_series_like(s):
+    """ Looks like a Pandas/cudf Series """
+    return set(dir(s)) > {'index', 'dtype', 'take', 'head'} and not isinstance(s, type)
+
+
+def is_index_like(s):
+    """ Looks like a Pandas/cudf Index """
+    attrs = set(dir(s))
+    return attrs > {'take', 'values'} and 'head' not in attrs and not isinstance(s, type)
```
mrocklin: It looks like these were copied from dask.dataframe.utils.  Note that we've had to change our definitions a bit upstream as we've run into issues.  You may want to recopy from there.

streamz/dataframe/core.py
```diff
@@ -245,14 +245,16 @@ def __setitem__(self, key, value):
...
+            result = self.map_partitions(df_type.assign, self, **{key: value})
 
         self.stream = result.stream
         self.example = result.example
         return self
 
     def query(self, expr, **kwargs):
-        return self.map_partitions(pd.DataFrame.query, self, expr, **kwargs)
+        df_type = type(self.example)
+        return self.map_partitions(df_type.query, self, expr, **kwargs)
```
mrocklin: This is fine, but if this happens more often you may want to use something like

```python
from dask.utils import M
return self.map_partitions(M.query, ...)
```

See https://github.com/dask/dask/blob/967e30c3a4b17400b91b83e0fc28dd1da50c06b9/dask/utils.py#L700

mrocklin: I would leave it as is for now though, and only use the methodcaller object if this starts to happen more frequently.

.travis.yml
```diff
@@ -10,6 +10,9 @@ matrix:
     - python: 2.7
     - python: 3.6
 
+ignore:
+  - streamz/dataframe/tests/test_cudf.py
```
mrocklin: Why this line?  I would expect the `pytest.importorskip` line to take care of this.

mrocklin:
@martindurant any concerns here?  I've glanced over things briefly, but you may want to take a stronger look.

martindurant:
Thanks for reviewing, @mrocklin , I'll look it over in the near future.

streamz/dataframe/core.py
skmatti: Sure, will refer to methodcaller if I have to do this more often.

martindurant:
Note that tests did not run for the latest commits, not sure why.

streamz/collection.py
```diff
@@ -204,11 +206,13 @@ def accumulate_partitions(self, func, *args, **kwargs):
...
             _, example = example
         stream = self.stream.accumulate(func, *args, start=start,
-                returns_state=returns_state, **kwargs)
+                                        returns_state=returns_state, **kwargs)
 
-        for typ, s_type in _stream_types[stream_type]:
-            if isinstance(example, typ):
-                return s_type(stream, example)
+        try:
+            s_type = get_stream_type(example, stream_type)
```
martindurant: Having `get_stream_type` return `None` where the class is not found, and no try/except here seemed simple enough, why the change?

streamz/dataframe/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def is_index_like(s):
+    """ Looks like a Pandas Index. ** Borrowed from dask.dataframe.utils ** """
+    typ = type(s)
+    return (all(hasattr(s, name) for name in ('name', 'dtype'))
+            and 'index' in typ.__name__.lower())
+
+
+def is_frame_like(frame, method, example=None):
```
martindurant: The name implies this is a boolean test, but it actually sets an attribute on the input frame

streamz/dataframe/utils.py
```diff
@@ -0,0 +1,34 @@
+def is_dataframe_like(df):
```
martindurant: I think we should copy - or import to prevent drift - the test `dask.dataframe.tests.test_utils_dataframe::test_is_dataframe_like` (which checks all three functions). Dask is part of the test built, or could skip when dask not present; in any case, don't want to duplicate dask code (function or test) if you can think of a way to avoid it.

streamz/dataframe/tests/test_cudf.py
```diff
@@ -0,0 +1,405 @@
+from __future__ import division, print_function
```
martindurant: I appreciate the large number of tests.

The xfails are here to indicate an intention to fix in the future? Do they depend on changes upstream?

Could it be possible to combine pandas and cudf tests to avoid duplication? Is that desirable?

streamz/dataframe/core.py
```diff
@@ -698,8 +720,8 @@ def _accumulate(self, Agg, **kwargs):
                                       diff=diff,
                                       window=window)
 
-        for typ, s_type in _stream_types[stream_type]:
-            if isinstance(example, typ):
+        for fn, s_type in _stream_types[stream_type]:
```
martindurant: please add comment: fn checks whether example can be represented as s_type

streamz/dataframe/core.py
```diff
@@ -324,7 +342,7 @@ def value_counts(self):
 
 
 class Index(Series):
-    _subtype = pd.Index
```
martindurant: So this is set in `is_frame_like` now (on the instance)? Seems a bit obfuscated. Is this class useful at all?

streamz/dataframe/core.py
```diff
@@ -314,7 +321,18 @@ class Series(Frame, _SeriesMixin):
...
-    _subtype = pd.Series
+
+    def __init__(self, *args, **kwargs):
+        example = None
+        if "example" in kwargs:
+            example = kwargs.get('example')
+        elif len(args) > 1:
+            example = args[1]
+        if isinstance(self, Index):
+            is_frame_like(self, is_index_like, example)
```
martindurant: I think it would be cleaner if `is_frame_like` (with a better name) passed a value back, which was then set on the instance attribute. Still would be nice to have *something* showing that a pandas-series and cu-series instance are related, could there be more than one attribute, or some inheritance?

streamz/collection.py
```diff
@@ -242,9 +246,12 @@ def verify(self, x):
                             (self._subtype, type(x)))
 
 
-def stream_type(example, stream_type='streaming'):
+def get_stream_type(example, stream_type='streaming'):
     for typ, s_type in _stream_types[stream_type]:
-        if isinstance(example, typ):
+        if isinstance(typ, types.FunctionType):
+            if typ(example):
```
martindurant: Comment here that for a set of types, `typ` here is a function that does the check. Should the other `typ`s also be functions, would that be cleaner?

streamz/dataframe/tests/test_cudf.py
skmatti: These xfails are due to cudf functions/methods not implemented yet, but will be available soon. So, these does not depend on upstream. 
Eventually we would want to combine these two. Matt suggested me to use a separate file for tests as most of pandas tests fail with cudf.

martindurant: OK, that's fair enough, so long a it's a conscious decision. Maybe worth putting in a comment at the top of the module.

streamz/dataframe/utils.py
skmatti: Added separate tests for utils. but imports dask tests for methods borrowed from dask.datafram.utils .

Currently I can't think a way to avoid this code duplication. May be we can list dask in the project requirements for Streamz DataFrame users?.

streamz/dataframe/core.py
skmatti: Ideally `Index` should have it's own `init` method. But currently it is being initialized in it's parent class(Series) with an if condition to avoid some code duplication. Not sure if this should be avoided.

streamz/collection.py
skmatti: I would need to implement functions for all those different types. I think there are more than a few types that I need to handle if I have to do this. Having said that, it would be nice if I can avoid having a separate function for type checking Frame types.

streamz/dataframe/core.py
skmatti: Made changes to the method. Not completely sure if it's close to something that you are suggesting.

skmatti:
@martindurant Restructuring the repo is the reason for new commits not triggering a build: [https://github.com/travis-ci/travis-ci/issues/3264](https://github.com/travis-ci/travis-ci/issues/3264)?

martindurant:
I did a sync on travis and a manually triggered build did succeed - can you try pushing something to see if it builds for you?

streamz/dataframe/utils.py
martindurant: No, we probably shouldn't do that, it should be possible to use streamz-cudf or streamz-pandas without dask

martindurant:
I think this is good to go, but would like a green mark for propriety's sake. If there's something else I need to do in Travis settings or elsewhere, please let me know.

skmatti:
Should this have triggered code coverage report as well?

martindurant:
The running of the test did post coverage results to codecov (https://codecov.io/gh/python-streamz/streamz/commit/d22149db96959a54c38754ecae0e9100e580d6f1 ), but perhaps it too is trying to post it's graphical, I'm not going to worry about that. Merging this now - thank you for the hard work here.

PR #226, opened 2019-03-08
Add Kafka Consumer Factory to cache consumers for FromKafkaBatched

skmatti:
I implemented a Kafka consumer factory that caches consumers and supplies consumer on demand to poll a message batch. This prevents unnecessary creation of consumers in the current implementation. Also, cached consumer is already seeked to latest offset(`low` of the next message batch ) to start polling messages immediately after a request for the next batch. One consumer per partition is created and reused for polling next batch in that partition.  

This also adds checkpointing for the job using Kafka consumer group. Consumer commits offsets after reading each batch. Restarting a job with same consumer group name would result in processing messages from last committed offset.

### DIFF
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -237,6 +237,58 @@ class from_kafka(Source):
         self.stopped = True
 
 
+class ConsumerFactory:
+    def __init__(self, consumer_params, topic):
+        self.consumer_params = self._fix_params(consumer_params)
+        self.topic = topic
+        self.consumers = {}
+        self.ck = None
+
+    def _fix_params(self, consumer_params):
+        if 'group.id' not in consumer_params:
+            import warnings
+            import uuid
+            group_id = "consumer-{}".format(uuid.uuid4())
+            warnings.warn("Consumer group id is not specified, Job restarts consume messages from offset 0.\n" +
+                          "Use group.id: {} for restarting this job from a check point in future.".format(group_id))
+            consumer_params['group.id'] = group_id
+
+        consumer_params['enable.auto.commit'] = 'false'
+        # consumer_params['max_poll_interval_ms'] = 20*60*1000
+        return consumer_params
+
+    def get_consumer(self, partition, low):
+        topic_partition = self._topic_partition(partition, low)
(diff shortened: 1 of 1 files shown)


martindurant:
I assume this is working for you locally? We typically found that kafka running on travis is well slower than locally, possibly leading to the failure to sync you are seeing in the test. When running locally, did you find the speedup you expected? Would be interesting to see benchmarks before/after.
There being so much code now, might be good to split into a separate module.

skmatti:
I think the tests related to this implementation are passing here, `from_kafka_batch` and `from_kafka_dask_batch`. The other Kafka tests are failing locally as well. These are fixed by @jsmaupin in [mrocklin/streamz/#216](https://github.com/mrocklin/streamz/pull/216). I see some improvement in performance. I am still figure out the best way to quantify this performance improvement. Yes, I am all up for creating a separate module for kafka integration specific code.

streamz/sources.py
```diff
@@ -237,22 +237,75 @@ def _close_consumer(self):
...
+    def _fix_params(self, consumer_params):
+        if 'group.id' not in consumer_params:
+            import warnings
+            import uuid
+            group_id = "consumer-{}".format(uuid.uuid4())
+            warnings.warn("Consumer group id is not specified, Job restarts consume messages from offset 0.\n" +
+                          "Use group.id: {} for restarting this job from a check point in future.".format(group_id))
+            consumer_params['group.id'] = group_id
+
+        consumer_params['enable.auto.commit'] = 'false'
```
martindurant: What was the reason behind this and the next (commented) line?

skmatti: I am manually committing the offsets after each batch is read successfully. And the job would restart from this checkpoint and not from an auto-commit in case of a failure. 

This configuration specifies the maximum time a consumer can stay idle between two `poll` operations. A consumer might stay idle for a longer period with current implementation. For example if Dask cluster is still processing older batches and the consumer is waiting for a thread to poll next batch.

martindurant:
Closing in favour of #233 - please add any further ideas there

PR #227, opened 2019-03-13
Add socket source

martindurant:
Could be used, for example, with python logging-to-socket


martindurant:
I would also like to make a HTTP server source, i.e., one event for each incoming GET, say, containing the URL parameters and data. How, though, do you tell a tornado Application or Server to start on a given event loop, or otherwise get it to run in the right place?

CJ-Wright:
I'm not certain, I'd like this as well, since I've been looking at making a ZMQ source.

martindurant:
Of course, this fails on kafka, as all the PRs seem to now. If https://github.com/mrocklin/streamz/pull/226 is to be merged, should we simply drop the unbatched kafka sources, because they cause us so much trouble? or, @skmatti , might your reuse of confluent objects help those too?

martindurant:
@CJ-Wright OK, solved the tornado server too.

martindurant:
@mrocklin , can you review, and we should come to an agreement on how to move ahead regarding the stalling kafka tests.

skmatti:
I think unbatched kafka source failures could be fixed by a commit from [#216](https://github.com/mrocklin/streamz/pull/216/commits/3b546f31815c74b9600b31f5d948d061ccba97a7).

streamz/sources.py
```diff
@@ -148,6 +148,154 @@ def do_poll(self):
...
+        import socket
+        if self.stopped:
+            self.stopped = False
+            self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)
+            self.sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+            self.sock.setblocking(0)
+            self.sock.bind(("", self.port))
+            self.sock.listen(128)
+            self.loop.add_handler(self.sock.fileno(), self.connection_ready,
+                                  self.loop.READ)
```
mrocklin: I'm somewhat surprised to see the use of raw sockets here.  I'm curious, was there a reason to go this route rather than create a Tornado IOStream here directly?  I see that you use them later on.  

I don't know a ton about managing nonblocking sockets manually, I'm probably just missing something.

martindurant: This code is adapted from the tornado IOLoop documentation: as far as I understand it, you do not have any IOStream until you accept an incoming connection on the open socket.

mrocklin: Ah, maybe the right abstraction here then is a Tornado TCPServer?

martindurant:
@mrocklin , I added a tornado.TCPServer version of the same. I could just remove the direct socket version, but I am not convinced that either is better than the other.

mrocklin:
>  I could just remove the direct socket version, but I am not convinced that either is better than the other

Yeah, I don't know either.  I don't know non-blocking sockets that well.  I tend to just let Tornado/asyncio handle it.

martindurant:
> I tend to just let Tornado/asyncio handle it.

I strongly suspect that the two forms are actually identical, so it might just be a matter of choice - or leave both.

mrocklin:
> I strongly suspect that the two forms are actually identical, so it might just be a matter of choice - or leave both.

I recommend that we stick with TCPServer if that's ok. 

Any comments on the async test and timing questions above?

martindurant:
> Any comments on the async test and timing questions above?

Sorry, missed that. I added an async test, and changed sleeps to (a)wait_for.

streamz/sources.py
```diff
@@ -147,6 +147,134 @@ def do_poll(self):
...
+        """Start HTTP server and listen"""
+        if self.stopped:
+            self.loop.add_callback(self._start_server)
+            self.stopped = False
+
+    def stop(self):
+        """Shutdown HTTP server"""
+        if not self.stopped:
+            self.server.stop()
+            self.server = None
```
mrocklin: You remove the server here, but not in the TCP solution.  Is there a reason?

streamz/sources.py
```diff
@@ -147,6 +147,134 @@ def do_poll(self):
...
+
+            @gen.coroutine
+            def post(self):
+                yield self.source._emit(self.request.body)
+                self.write('OK')
+
+        application = Application([
+            (self.path, Handler),
+        ])
+        self.server = HTTPServer(application)
```
mrocklin: I can imagine wanting to pass in keyword arguments to the constructor and have them percolate to here (same with TCPServer) . I'm not sure though.

streamz/tests/test_sources.py
```diff
@@ -0,0 +1,81 @@
...
+        sock.connect(("localhost", port))
+        sock.send(b'data\n')
+
+        sock2 = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sock2.connect(("localhost", port))
+        sock2.send(b'data2\n')
+    finally:
+        s.stop()
+
+    wait_for(lambda: out == [b'data\n', b'data\n', b'data2\n'], 2, period=0.01)
```
mrocklin: Do we need to clean up the sockets here?  Does this leak file descriptors?

streamz/tests/test_sources.py
```diff
@@ -0,0 +1,81 @@
...
+        s.stop()
+
+
+def test_http():
+    requests = pytest.importorskip('requests')
+    port = 9875
+    s = Source.from_http_server(port)
+    out = s.sink_to_list()
+    s.start()
+    time.sleep(0.02)  # allow loop to run
```
mrocklin: Is this necessary?  If so, why this time?

In general, I'll probably challenge any sleep like this.  If it's necessary to make things run then it's likely too short in all cases (GC on travis can take arbitrarily long times) and so results in intermittent failures.  I spend a non-trivial time tracking these things down in the dask/distributed codebase, so I've grown fairly allergic to them.

streamz/sources.py
```diff
@@ -228,15 +233,18 @@ class from_http_server(Source):
...
     Example
     -------
     >>> source = Source.from_http_server(4567)  # doctest: +SKIP
     """
 
-    def __init__(self, port, path='/.*', start=False):
+    def __init__(self, port, path='/.*', start=False, server_kwargs=None):
         self.port = port
         self.path = path
+        self.server_kwargs = server_kwargs or {}
```
mrocklin: In one case you use `tcp_kwargs` and in the other `server_kwargs`.  Maybe use `server_kwargs` in both places for consistency of the API?

mrocklin:
Thanks for this @martindurant .  Should be fun to play with !

PR #230, opened 2019-03-16
Kafka fixes

martindurant:
Attempts to extend #216 following feedback 

Still contains a gen-sleep in the dask test. This could be thought OK usage, since that's not the way that people will use it in the wild. Good chances of these tests failing now again on Travis, we'll see.

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1295,6 +1295,31 @@ class to_kafka(Stream):
         self.producer.flush(timeout)
 
 
+@Stream.register_api()
+class to_kafka_batched(Stream):
+    def __init__(self, upstream, topic, producer_config, n, **kwargs):
+        import confluent_kafka as ck
+
+        self.topic = topic
+        self.producer = ck.Producer(producer_config)
+        self.n = n
+        self.buffer = []
+
+        Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
+
+    def update(self, x, who=None):
+        self.buffer.append(x)
+        if len(self.buffer) >= self.n:
+            self.flush(False)
+
+    def flush(self, producer=True):
+        for x in self.buffer:
+            self.producer.produce(self.topic, x)
+        self.buffer = []
+        if producer:
+            self.producer.flush()
+
+
 def sync(loop, func, *args, **kwargs):
     """
     Run coroutine in loop running in separate thread.
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1268,7 +1268,7 @@ class to_kafka(Stream):
 
         Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
         self.stopped = False
-        self.polltime = 0.5
+        self.polltime = 0.2
         self.loop.add_callback(self.poll)
         self.futures = []
 
@@ -1295,7 +1295,7 @@ class to_kafka(Stream):
                     self.producer.produce(self.topic, x, callback=self.cb)
                     return
                 except BufferError:
-                    yield gen.sleep(0.1)
+                    yield gen.sleep(self.polltime)
                 except Exception as e:
                     future.set_exception(e)
                     return
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import atexit
 from collections import deque
 from datetime import timedelta
 import functools
@@ -15,6 +16,10 @@ from tornado import gen
 from tornado.locks import Condition
 from tornado.ioloop import IOLoop
 from tornado.queues import Queue
+try:
+    import confluent_kafka as ck
+except ImportError:
+    ck = None  # confluent_kafka is only required when interacting with Kafka
 try:
     from tornado.ioloop import PollIOLoop
 except ImportError:
@@ -1229,6 +1234,54 @@ class latest(Stream):
             yield self.condition.wait()
             [x] = self.next
             yield self._emit(x)
+
+
+@Stream.register_api()
+class to_kafka(Stream):
+    """ Writes data in the stream to Kafka
+
+    The Confluent client sends messages in batches and asynchronously. When a
+    message is received, or in the event of an error, this stream will emit
+    a tuple in the form of (err, msg) see
+    https://github.com/confluentinc/confluent-kafka-python
+
+    Parameters
+    ----------
+    topic : string
+        The topic which to write
+    producer_config : dict
+        Settings to set up the stream, see
+        https://docs.confluent.io/current/clients/confluent-kafka-python/#configuration
+        https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
+        Examples:
+        bootstrap.servers: Connection string(s) (host:port) by which to reach Kafka
+
+    Examples
+    --------
+    >>> source = Stream()
+    >>> source.map(lambda x: x + 1).to_kafka('test', {'bootstrap.servers': 'localhost'}).sink(
+    ...     lambda x: print(x[1].topic(), x[1].partition()))
+    >>> for i in range(3):
+    ...     source.emit(i)
+    test 0
+    test 0
+    test 0
+    """
+    def __init__(self, upstream, topic, producer_config, **kwargs):
+        self.topic = topic
+        self.producer = ck.Producer(producer_config)
+        atexit.register(self.producer.flush)
+
+        Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
+
+        _global_sinks.add(self)
+
+    def update(self, x, who=None):
+        self.producer.poll(0)
+        self.producer.produce(self.topic, str(x).encode('utf-8'), callback=self.cb)
+
+    def cb(self, err, msg):
+        self.emit((err, msg))
 
 
 def sync(loop, func, *args, **kwargs):
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1274,8 +1274,6 @@ class to_kafka(Stream):
 
         Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
 
-        _global_sinks.add(self)
-
     def update(self, x, who=None):
         self.producer.poll(0)
         self.producer.produce(self.topic, str(x).encode('utf-8'), callback=self.cb)
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-from collections import deque, namedtuple
+from collections import deque
 from datetime import timedelta
 import functools
 import logging
@@ -1254,18 +1254,16 @@ class to_kafka(Stream):
 
     Examples
     --------
+    >>> from streamz import Stream
     >>> ARGS = {'bootstrap.servers': 'localhost:9092'}
     >>> source = Stream()
-    >>> kafka = source.map(lambda x: x + 1).to_kafka('test', ARGS)
-    >>> kafka.sink(lambda x: print(x[1].topic(), x[1].partition()))
-    >>> for i in range(3):
-    ...     source.emit(i)
-    >>> kafka.flush()
-    test 0
-    test 0
-    test 0
+    >>> source.map(lambda x: str(x)).to_kafka('test', ARGS)
+    <to_kafka>
+    >>> for i in range(10):
+    ...     # In script simply use await keyword
+    ...     asyncio.wait(source.emit(i, asynchronous=True))
+
     """
-    Result = namedtuple('Result', ['err', 'msg'])
 
     def __init__(self, upstream, topic, producer_config, **kwargs):
         import confluent_kafka as ck
@@ -1275,15 +1273,30 @@ class to_kafka(Stream):
 
         Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
 
+        self.loop.add_callback(self.cb)
+
     def update(self, x, who=None):
-        self.producer.poll(0)
-        self.producer.produce(self.topic, x, callback=self.cb)
+        future = gen.Future()
+        loop = self.loop.current()
+
+        def cb(err, msg):
+            if err:
+                exc = Exception((err, msg.value() if msg else None))
+                loop.add_callback(future.set_exception, exc)
+            else:
+                loop.add_callback(future.set_result, None)
+
+        self.producer.produce(self.topic, x, callback=cb)
+
+        return future
+
+    @gen.coroutine
+    def cb(self):
+        while True:
+            self.producer.poll(65500)
 
     def flush(self, timeout=-1):
         self.producer.flush(timeout)
-
-    def cb(self, err, msg):
-        self.emit(self.Result(err, msg.value() if msg else None))
 
 
 def sync(loop, func, *args, **kwargs):
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1269,14 +1269,28 @@ class to_kafka(Stream):
         self.producer = ck.Producer(producer_config)
 
         Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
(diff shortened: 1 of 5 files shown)


martindurant:
@mrocklin , I am thinking that the use pattern of the Future in to_kafka could be wrong. The Future is created when data is provided (in .update()), and set immediately when the producer is called. That happens on the event loop, but does not mean the data has been delivered. 

Should the Future only get set in .cb(), after successful delivery? If so, there can be more than one future in flight? They would need to be stored in a queue (order matters, content does not).

How would one test that backpressure was working as it should?

streamz/core.py
```diff
@@ -1231,6 +1231,111 @@ def cb(self):
...
+            if self.stopped:
+                # kill producer?
+                break
+            yield gen.sleep(self.polltime)
+
+    def update(self, x, who=None):
+        future = gen.Future()
+
+        def _():
+            while True:
```
mrocklin: I'm curious, why the `while True` here?  

If there is a `BufferError` or another exception we want to try this again and again?

streamz/core.py
```diff
@@ -1231,6 +1231,111 @@ def cb(self):
...
+                try:
+                    # this runs asynchronously, in C-K's thread
+                    self.producer.produce(self.topic, x, callback=self.cb)
```
mrocklin: Can you verify that this doesn't block if the producer is busy?

mrocklin: Also, does the `callback` get run on the event loop's thread, or C-K's thread?

mrocklin: If it gets run in C-K's thread then you may have to use `self.loop.add_callback`, which is the only threadsafe mechanism to create coroutines from other threads.

streamz/core.py
```diff
@@ -1231,6 +1231,111 @@ def cb(self):
...
+    @gen.coroutine
+    def cb(self, err, msg):
+        yield self._emit(self.Result(err, msg.value() if msg else None))
+
+    def flush(self, timeout=-1):
+        self.producer.flush(timeout)
+
+
+@Stream.register_api()
+class to_kafka_batched(Stream):
```
mrocklin: read_kafka_batched was oriented around supporting Dask.  Is this the same?

streamz/core.py
martindurant: Only for BufferError, I think: the producer queue is full, and so message can't be sent at the moment. This implies that the function should be a coroutine and gen-sleep for a while, fixed in the coming commit.

mrocklin: Ah, that makes sense.

streamz/core.py
martindurant: I should have been more verbose here. The [doc](https://docs.confluent.io/current/clients/confluent-kafka-python/#confluent_kafka.Producer.produce) says that the method call is always async (returns immediately), and I have found no case where it isn't. 

The actual marshalling of the message to kafka happens in C-K's own thread, a C-level place. But the python callbacks for delivery are triggered upon `poll()`, and so that call and the coroutine they refer to must happen in the same event loop - so `poll()` is also in a coroutine.

streamz/core.py
martindurant: I don't know. I can see if it works with Dask. I'm trying to fix the existing code rather than design it's purpose :|  I gather better performance was expected, in terms of messages/s with batches.

martindurant:
I'll have a loop at to_kafka_batched yet, and maybe remove it if I can't figure out what to do with it; it doesn't appear to be tested. It doesn't use futures, so it doesn't currently make a good sink.

mrocklin:
We could also drop that for a first pass and just get a solid `to_kafka` in?

martindurant:
OK, removed it. May come back to that - the goal here from my point of view was to get general streamz PRs green again.

streamz/core.py
```diff
@@ -1231,6 +1231,91 @@ def cb(self):
...
+
+        @gen.coroutine
+        def _():
+            while True:
+                try:
+                    # this runs asynchronously, in C-K's thread
+                    self.producer.produce(self.topic, x, callback=self.cb)
+                    return
+                except BufferError:
+                    yield gen.sleep(0.1)
```
mrocklin: Should this be `self.poll_time`?

streamz/core.py
```diff
@@ -1231,6 +1231,91 @@ def cb(self):
...
+                except Exception as e:
+                    future.set_exception(e)
+                    return
+
+        self.loop.add_callback(_)
+        return future
+
+    @gen.coroutine
+    def cb(self, err, msg):
```
mrocklin: Does confluent_kafka run the callback in a separate thread?  If so then we can't make this a coroutine.  

That might be OK.  to_kafka is likely to be used as a sink, and so we don't expect to get a future from downstream that we're supposed to wait on.  It might be enough here to just set the result and be done. 

If cb is called in the thread where `update` is called, then it might be enough to just set the result and be done

```python
def cb(self, err, msg):
    if good:
        future.set_result(...)
    else:
        future.set_exception(...)
```

However, if it's going to be called within a separate thread, then we'll need to use `loop.add_callback`, which is the only threadsafe method in torando

```python
def cb(...):
    def _():
        future.set_result(...)
        ...
    self.loop.add_callback(_)
```

streamz/core.py
martindurant: right

streamz/core.py
martindurant: > Does confluent_kafka run the callback in a separate thread?

No, it runs in the thread which calls `producer.poll()`, which is here being called in a coroutine on the current loop. Only the actual low-level delivery runs in another thread.

> it might be enough to just set the result and be done

There is no `Sink` type, though, right? I mean, it may be reasonable to expect events to be passed down a chain still.

streamz/core.py
```diff
@@ -1231,6 +1231,91 @@ def cb(self):
...
+
+        Stream.__init__(self, upstream, ensure_io_loop=True, **kwargs)
+        self.stopped = False
+        self.polltime = 0.2
+        self.loop.add_callback(self.poll)
+        self.futures = []
+
+    @gen.coroutine
+    def poll(self):
+        while True:
```
mrocklin:
```suggestion
        while not self.stopped:
```

streamz/core.py
```diff
@@ -1231,6 +1231,91 @@ def cb(self):
...
+            # executes callbacks for any delivered data, in this thread
+            # if no meesages were sent, nothing happens
+            self.producer.poll(0)
+            if self.stopped:
```
mrocklin:
```suggestion
```

streamz/core.py
```diff
@@ -1231,6 +1231,91 @@ def cb(self):
...
+                # kill producer?
+                break
```
mrocklin:
```suggestion
```

streamz/tests/test_kafka.py
```diff
@@ -149,16 +159,15 @@ def test_from_kafka_thread():
...
         out = stream.sink_to_list()
         stream.start()
         for i in range(10):
             kafka.produce(TOPIC, b'value-%d' % i)
         kafka.flush()
         # it takes some time for messages to come back out of kafka
-        startlen = len([o for o in out if o])
-        wait_for(lambda: len([o for o in out if o]) == startlen + 10, 10,
-                 period=0.1)
+        wait_for(lambda: len(out) == 10, 10, period=0.1)
```
mrocklin: If you're willing, we might want to use `await_for` here instead if this is an async test.

streamz/tests/test_kafka.py
```diff
@@ -195,18 +206,16 @@ def test_kafka_dask_batch(c, s, w1, w2):
     ARGS = {'bootstrap.servers': 'localhost:9092',
             'group.id': 'streamz-test%i' % j}
     with kafka_service() as kafka:
+        kafka, TOPIC = kafka
         stream = Stream.from_kafka_batched(TOPIC, ARGS, asynchronous=True,
                                            dask=True)
+        out = stream.gather().sink_to_list()
         stream.start()
+        yield gen.sleep(5)  # this frees the loop while dask workers report in
```
mrocklin: Would an `await_for` statement make more sense here?  Do we know what keeps the dask workers so long?

streamz/tests/test_kafka.py
```diff
@@ -141,6 +132,25 @@ def test_from_kafka():
...
+        source = Stream()
+        kafka = source.to_kafka(TOPIC, ARGS)
+        out = kafka.sink_to_list()
+
+        for i in range(10):
+            yield source.emit(b'value-%d' % i)
+
+        source.emit('final message')
+        kafka.flush()
+        wait_for(lambda: len(out) == 11, 10, period=0.1)
```
mrocklin:
```suggestion
        yield await_for(lambda: len(out) == 11, 10, period=0.1)
```

streamz/tests/test_kafka.py
```diff
@@ -141,6 +132,25 @@ def test_from_kafka():
         # absolute sleep here, since we expect output list *not* to change
         yield gen.sleep(1)
         assert out[-1] == b'final message'
+        stream._close_consumer()
```
mrocklin: It looks like we do this a bit.  Should we use `weakref.finalize` on the `stream` object to make sure that this happens when the object goes away?

streamz/tests/test_kafka.py
mrocklin:
```suggestion
        yield await_for(lambda: len(out) == 10, 10, period=0.1)
```

streamz/tests/test_kafka.py
martindurant: I do not know what's happening in this time. I tried `yield await_for(lambda: c.ncores())`, but get
```distributed/distributed/utils_test.py:945, AssertionError```, an absolute timeout in `gen_cluster` and warnings like
```
distributed.core - INFO - Event loop was unresponsive in Worker for 10.98s.
```
i.e., the gen.sleep of a single seems to release where querying the cluster is actually blocking.

Comment (mrocklin) [reply streamz/tests/test_kafka.py]: Note that `c.ncores` will be a coroutine in async mode, so you would need to `yield` the result.

However, I also don't understand the need here.  Any workers that will arrive have arrived at the start of the function.  `gen_cluster` waits until everyone is here.

Comment (martindurant) [reply streamz/tests/test_kafka.py]: Yes, I yielded for ncores - didn't work :|
Sorry, I don't know more, I can't think how to investigate this further. The following also fails, with the same timeout in gen_cluster (*not* in the test itself)
```
        t = time.time()
        while True:
            yield gen.sleep(0.2)
            ncore = yield c.ncores()
            if ncore:
                break
            assert time.time() - t < 10, "Timeout"
```

**Comment by martindurant** [comment]
Note that the `finalize` thing fails here, because it doesn't exist in py2. Distributed has an implementation, but I'm wondering whether it would be better to more to a `__del__` implementation; or just revert.

Again, I am wanting to clean this up so that the other lingering PRs can get moving.

**Comment by mrocklin** [comment]
> Note that the finalize thing fails here, because it doesn't exist in py2. Distributed has an implementation, but I'm wondering whether it would be better to more to a __del__ implementation; or just revert.

FWIW I'd be quite happy to drop Python 2 in this project.

> Again, I am wanting to clean this up so that the other lingering PRs can get moving.

OK.  If you want to revert the previous commit I'd be happy to merge.

**Comment by martindurant** [comment]
Done.

>  I'd be quite happy to drop Python 2 

Great! That would be a separate PR.
Does tornado gen-type syntax and async/await mix OK, or would all of that stuff need to be rewritten (or just use gen forever)?

File: streamz/tests/test_kafka.py (hunk 10 above)
Comment (mrocklin) [reply streamz/tests/test_kafka.py]: I'm still confused about what this sleep is doing.  I would be *very* surprised if the workers weren't already up and connected at this point.

```python
In [1]: from distributed.utils_test import gen_cluster

In [2]: @gen_cluster(client=True)
   ...: def test_foo(c, s, a, b):
   ...:     assert len(s.workers) == 2
   ...:

In [3]: test_foo()
distributed.scheduler - INFO - Clear task state
distributed.scheduler - INFO -   Scheduler at:     tcp://127.0.0.1:52172
distributed.worker - INFO -       Start worker at:      tcp://127.0.0.1:52173
distributed.worker - INFO -          Listening to:      tcp://127.0.0.1:52173
distributed.worker - INFO - Waiting to connect to:      tcp://127.0.0.1:52172
distributed.worker - INFO - -------------------------------------------------
distributed.worker - INFO -               Threads:                          1
distributed.worker - INFO -                Memory:                   17.18 GB
distributed.worker - INFO -       Local Directory: /Users/mrocklin/workspace/distributed/dask-worker-space/worker-c59n7r0g
distributed.worker - INFO - -------------------------------------------------
distributed.worker - INFO -       Start worker at:      tcp://127.0.0.1:52174
distributed.worker - INFO -          Listening to:      tcp://127.0.0.1:52174
distributed.worker - INFO - Waiting to connect to:      tcp://127.0.0.1:52172
distributed.worker - INFO - -------------------------------------------------
distributed.worker - INFO -               Threads:                          2
distributed.worker - INFO -                Memory:                   17.18 GB
distributed.worker - INFO -       Local Directory: /Users/mrocklin/workspace/distributed/dask-worker-space/worker-btnrzcte
distributed.worker - INFO - -------------------------------------------------
distributed.scheduler - INFO - Register tcp://127.0.0.1:52173
distributed.scheduler - INFO - Register tcp://127.0.0.1:52174
distributed.scheduler - INFO - Starting worker compute stream, tcp://127.0.0.1:52173
distributed.core - INFO - Starting established connection
distributed.scheduler - INFO - Starting worker compute stream, tcp://127.0.0.1:52174
distributed.core - INFO - Starting established connection
distributed.worker - INFO -         Registered to:      tcp://127.0.0.1:52172
distributed.worker - INFO - -------------------------------------------------
distributed.worker - INFO -         Registered to:      tcp://127.0.0.1:52172
distributed.worker - INFO - -------------------------------------------------
distributed.core - INFO - Starting established connection
distributed.core - INFO - Starting established connection
distributed.scheduler - INFO - Receive client connection: Client-3da15092-4b26-11e9-9923-acde48001122
distributed.core - INFO - Starting established connection
distributed.scheduler - INFO - Remove client Client-3da15092-4b26-11e9-9923-acde48001122
distributed.scheduler - INFO - Remove client Client-3da15092-4b26-11e9-9923-acde48001122
distributed.scheduler - INFO - Close client connection: Client-3da15092-4b26-11e9-9923-acde48001122
distributed.worker - INFO - Stopping worker at tcp://127.0.0.1:52173
distributed.worker - INFO - Stopping worker at tcp://127.0.0.1:52174
distributed.scheduler - INFO - Scheduler closing...
distributed.scheduler - INFO - Scheduler closing all comms
distributed.scheduler - INFO - Remove worker tcp://127.0.0.1:52173
distributed.core - INFO - Removing comms to tcp://127.0.0.1:52173
distributed.batched - INFO - Batched Comm Closed: in <closed TCP>: Stream is closed
distributed.scheduler - INFO - Remove worker tcp://127.0.0.1:52174
distributed.core - INFO - Removing comms to tcp://127.0.0.1:52174
distributed.scheduler - INFO - Lost all workers
distributed.worker - INFO - Connection to scheduler broken.  Reconnecting...
distributed.worker - INFO - Connection to scheduler broken.  Reconnecting...
distributed.worker - INFO - -------------------------------------------------
distributed.worker - INFO - -------------------------------------------------
```

We may still need a sleep here, but it would be good to know why.

mrocklin:
> Does tornado gen-type syntax and async/await mix OK, or would all of that stuff need to be rewritten (or just use gen forever)?

They mostly mix, but there are some small semantic differences.  Coroutines created with `@gen.coroutine` start once they're called.  `async def` functions have to be awaited before they start.

martindurant:
Hurray!

streamz/tests/test_kafka.py
martindurant: Agreed, but I don't know how to investigate.

Issue #199, opened 2018-08-17
[Feature request] Stream.from_textfile: only emit lines appended after `start()`


wmayner:
I think it would be useful to have a source that only emits lines that have been appended to the file since the stream was started.

This could be implemented as a keyword argument to `Stream.from_textfile` that triggers `file.seek`:

```python
def __init__(self, f, poll_interval=0.100, delimiter='\n', start=False,
             from_end=False, **kwargs):
    # ...
    self.from_end = from_end
    # ...

def start(self):
    self.stopped = False
    if self.from_end:
        self.file.seek(0, 2)
    self.loop.add_callback(self.do_poll)
```

_Use case:_
I'm running simulations of biologically-realistic neural networks with NEST, which outputs measurements of membrane potentials, etc., as lines appended to text files. These simulations can run for a long time. I want to be able to use `pd.read_csv` to quickly load existing output into a DataFrame to plot the activity, and then stream any lines that are appended afterwards so that the activity plot updates in real-time. If the existing output is large, then streaming everything instead of using `pd.read_csv` is prohibitively slow (at least in my implementation; maybe that could be improved).

PR #231, opened 2019-03-20
Reimplement #201

martindurant:
Some effect as #201, but should now pass tests

### DIFF
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -788,6 +788,25 @@ def test_from_file():
 
 
 @gen_test()
+def test_from_file_end():
+    with tmpfile() as fn:
+        with open(fn, 'wt') as f:
+            f.write('data1\n')
+            f.flush()
+
+            source = Stream.from_textfile(fn, poll_interval=0.010,
+                                          start=False, from_end=True)
+            out = source.sink_to_list()
+            source.start()
+            assert out == []
+            yield gen.sleep(0.01)
+
+            f.write('data2\n')
+            f.flush()
+            yield await_for(lambda: out == ['data2\n'], timeout=5, period=0.1)
+
+
+@gen_test()
 def test_filenames():
     with tmpfile() as fn:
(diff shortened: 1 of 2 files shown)


martindurant:
i.e., add "from_end" parameter to from_textfiles, which only makes events out of new data.

streamz/tests/test_core.py
```diff
@@ -787,6 +787,25 @@ def test_from_file():
...
+        with open(fn, 'wt') as f:
+            f.write('data1\n')
+            f.flush()
+
+            source = Stream.from_textfile(fn, poll_interval=0.010,
+                                          start=False, from_end=True)
+            out = source.sink_to_list()
+            source.start()
+            assert out == []
+            yield gen.sleep(0.01)
```
mrocklin: A little nit-picky, but why this sleep?

martindurant: The `seek()` to end doesn't happen until the coroutine runs for the first time, so if you were to write immediately here, you would always seek past the new data. This yield makes sure the coroutine runs once. Would a yield without the sleep have the same effect, or is there another way to say "run coroutines now"?

mrocklin: If you just want one cycle to go through, then try `yield gen.moment`.

If you might need a few cycles, then you should probably have an `await_for`

martindurant: I'll try `moment` tomorrow; `await_for` is tricky when you *don't* expect the output to change. Perhaps it would be reasonable to await on at attribute of the source.

martindurant:
btw: when I run flake locally, I get loads more issues than show up in travis; how is this configured?

mrocklin:
I'm not sure.  I would check the procedure in the .travis.yml file.  Maybe
a different version of flake8?

martindurant:
I believe this is ready (and should close #201)

martindurant:
Fixes #199

PR #234, opened 2019-03-25
Added from_textfile_batched() method.

chinmaychandak:
Added a from_textfile_batched() method, which takes batchSize as a parameter, and emits the corresponding-sized chunks of text downstream. 

This method is DaskStream compatible. 

Added pertinent unit test as well.

### DIFF
--- a/streamz/dask.py
+++ b/streamz/dask.py
@@ -198,3 +198,8 @@ class filenames(DaskStream, sources.filenames):
 @DaskStream.register_api(staticmethod)
 class from_textfile(DaskStream, sources.from_textfile):
     pass
+
+
+@DaskStream.register_api(staticmethod)
+class from_textfile_batched(DaskStream, sources.from_textfile_batched):
+    pass
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -98,6 +98,52 @@ class from_textfile(Source):
 
 
 @Stream.register_api(staticmethod)
+class from_textfile_batched(Source):
+    
+    """
+    Stream data from a text file in batches. Compatible with DaskStream.
+    """
+    
+    def __init__(self, f, poll_interval=0.100, delimiter='\n', start=False, batchSize=1,
+                 **kwargs):
+        if isinstance(f, str):
+            f = open(f, 'r')
+        self.file = f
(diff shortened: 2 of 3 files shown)


martindurant:
I am ambivalent on this PR, since you could easily implement the same using `from_textfile().partition`. What is the justification for a new method, does it provide a significant performance boost?

 To get tests to pass, I suspect it would be enough to merge from master and fix the conflict. If I don't hear otherwise, I'll close this in the near future.

martindurant:
(also, sorry that it's taken me so long to respond here)

chinmaychandak:
The method would not only provide flexibility to tune the delimiter, batch size of data you want to read depending on the type/volume/message size/speed of data. Also, I had experimented with the partition approach, but my method performed better in streaming use cases.

martindurant:
from_textfile does already allow you to specity the delimiter https://streamz.readthedocs.io/en/latest/api.html#streamz.from_textfile

> batch size of data

`.partition()` would allow you to do exactly this, no? You mean you want to batch by number of bytes? I don't think that's what your code does.

>  my method performed better

Would love to see some numbers here. I suspect that when the CPU load per message processing is significant, you would see no difference from this first stage.

chinmaychandak:
So, this still holds true: [https://github.com/python-streamz/streamz/issues/235].

If I remember correctly, I was able to get around 6 MBps when I used .partition() instead. The speedup in using from_textfile_batched() would come from the fact that if the batch size is large enough, the buffering happens while emitting elements downstream at the start of the streaming pipeline, as opposed to calling self._emit() for each element and then buffering using .partition(). 

Let me know what you think of this. 

> I suspect that when the CPU load per message processing is significant, you would see no difference from this first stage.

This may be true, but the speedup would still be significant if the batch sizes are large, IMO.

martindurant:
Compared to 10-12MBps, right? What size were the lines here? I am still tempted to say that it's not worth complicating the API. Perhaps you can think of a way to add `batchsize=` into the existing from_textfiles.

chinmaychandak:
> What size were the lines here?

1.1 kB

> I am still tempted to say that it's not worth complicating the API. 

Okay.

PR #242, opened 2019-05-01
Plot cyclic graphs

CJ-Wright:
This changes the graph creation code to support cyclic graphs.


CJ-Wright:
@martindurant ready for review!

CJ-Wright:
poke

martindurant:
I am planning on doing a release of all the recently-added material, and will get to this after that.

CJ-Wright:
Ah ok. I thought this one of my less controversial PRs :smile:

martindurant:
I think it has the most number of changes, and it's not so simple to understand...

CJ-Wright:
That's fair.

streamz/graph.py
```diff
@@ -21,107 +21,115 @@ def _clean_text(text, match=None):
     return text
 
 
-def create_graph(node, graph, prior_node=None, pc=None):
-    """Create graph from a single node, searching up and down the chain
+def build_node_set(node, s=None):
+    """Build a set of all the nodes in a rapidz graph
```
CJ-Wright:
```suggestion
    """Build a set of all the nodes in a streamz graph
```

streamz/graph.py
```diff
@@ -21,107 +21,115 @@ def _clean_text(text, match=None):
...
-        if pc == 'downstream':
-            graph.add_edge(tt, t, label=str(i))
-        else:
-            graph.add_edge(t, tt)
+    # Step 1 build a set of all the nodes
+    node_set = build_node_set(node)
+    if None in node_set:
+        node_set.remove(None)
+
+    # Step 2 for each node in the set add do the graph
```
CJ-Wright:
```suggestion
    # Step 2 for each node in the set add to the graph
```

streamz/graph.py
```diff
@@ -21,107 +21,115 @@ def _clean_text(text, match=None):
...
-
-    for nodes, pc in zip([list(node.downstreams), list(node.upstreams)],
-                         ['downstream', 'upstreams']):
-        for node2 in nodes:
-            if node2 is not None:
-                create_graph(node2, graph, node, pc=pc)
+    s.add(node)
+    s.update(node.upstreams)
+    s.update({n for n in node.downstreams})
+    [build_node_set(n, s) for n in list(s)]
```
martindurant: Shouldn't this be done only for *new* nodes in the set? I know that sets nicely dedup, but seems like we could have a lot unnecessary recursion here.

streamz/tests/test_graph.py
```diff
@@ -24,6 +24,21 @@ def test_create_graph():
         assert t in g
 
 
+def test_create_cyclic_graph():
```
martindurant: Would this have caused infinite recursion before?

streamz/tests/test_graph.py
```diff
@@ -24,6 +24,21 @@ def test_create_graph():
...
+
+    n1 = source1.zip(source2)
+    n2 = n1.map(add)
+    n2.connect(source1)
+
+    g = nx.DiGraph()
+    create_graph(n2, g)
+    for t in [hash(a) for a in [source1, source2, n1, n2]]:
+        assert t in g
+    assert nx.find_cycle(g)
```
martindurant: Meaning that the graph really does point back to itself?

streamz/tests/test_graph.py
CJ-Wright: Yes, this would have failed.

streamz/tests/test_graph.py
CJ-Wright: Yes, meaning that we could create a cyclic nx graph representation (which was not possible before).

CJ-Wright:
By example output do you mean a graph image?

martindurant:
> do you mean a graph image

Yes please!

Also, is there a way to specify the layout of cyclic graphs: if a line, which goes left-most, etc.

CJ-Wright:
I don't know about the specification of the layouts, graphviz kinda does its own thing. I've added a plot to the docs (under the recursive section).

martindurant:
OK, and nice to include in the doc, thanks. I bet people will ask for finer control down the line.

martindurant:
Thanks, @CJ-Wright , merging this now. The code is still a bit complex and I suspect we might revisit it later, but this is definitely an improvement for now.

PR #243, opened 2019-05-01
allow filter to take args and kwargs

CJ-Wright:


martindurant:
Can we please have docs and tests for this?
I wonder if it isn't simpler to suggest wrapping the predicate function (with partial, maybe) to do this.

CJ-Wright:
I'm happy to add docs/tests, although we might want to sort out if we want this in first.

I think it would be good to have a uniform API structure. Things like `map`, `starmap`, and `sink` take in a function and optional args and kwargs. To me it seems that mirroring this with filter would be good, since it also takes in a function which could have args and/or kwargs. I don't think going the other direction, removing these optional args and kwargs for `map` et. al. and opting for partials, would be a good API, since it adds an extra layer of indirection at the cost of removing a handful of lines of code.

martindurant:
> Things like map, starmap, and sink take in a function and optional args and kwargs. To me it seems that mirroring this with filter would be intuitive

That is a fair point. 

I think you are failing on flake.

martindurant:
Was there an associated issue?

PR #244, opened 2019-05-01
allow non hashable data to be used in unique

CJ-Wright:


streamz/core.py
```diff
@@ -1047,18 +1056,33 @@ class unique(Stream):
...
+            if isinstance(y, Hashable):
+                self.seen = dict()
+                if self.history:
+                    # if it is hashable use LRU cache
+                    if isinstance(y, Hashable):
+                        from zict import LRU
+                        self.seen = LRU(self.history, self.seen)
+            # if not hashable use deque (since it doesn't need a hash)
+            else:
+                self.seen = deque(maxlen=self.history)
```
martindurant: With a deque, we get the specified number of items, not the specified number of *unique* items. I think that, if we are allowing now to process data to get something hashable, this branch is unnecessary.

streamz/core.py
```diff
@@ -1047,18 +1056,33 @@ class unique(Stream):
...
-            self.seen = LRU(history, self.seen)
+        self.history = history
 
         Stream.__init__(self, upstream, **kwargs)
 
     def update(self, x, who=None):
         y = self.key(x)
+        # If this is the first piece of data make the cache
+        if self.seen is None:
+            if isinstance(y, Hashable):
```
martindurant: I don't think that behaviour should depend on what the first item happens to be: the cache should be set up in init

streamz/core.py
```diff
@@ -1047,18 +1056,33 @@ class unique(Stream):
...
+                self.seen = dict()
+                if self.history:
+                    # if it is hashable use LRU cache
+                    if isinstance(y, Hashable):
```
martindurant: This seems to be an `and`, and I would say that the latter should always be True

streamz/core.py
```diff
@@ -1035,6 +1035,15 @@ class unique(Stream):
     parameter.  For example setting ``history=1`` avoids sending through
     elements when one is repeated right after the other.
 
+    Parameters
+    ----------
+    history : int or None, optional
```
martindurant: `maxsize`, I think

streamz/core.py
```diff
@@ -1035,6 +1035,15 @@ class unique(Stream):
...
+        number of stored unique values to check against
+    key : function, optional
```
martindurant: You could call this hashfunc to make it clear: it is not a key (in the example, this would be simply "a"), and the purpose is to make the data hashable.

streamz/core.py
CJ-Wright: To be fair the ``key`` kwarg was always there, just not documented.

CJ-Wright: The uniqueness is guaranteed by https://github.com/python-streamz/streamz/pull/244/files#diff-d4e512993f47710ea9e7155d66011a24R1080 not the `deque` itself.

martindurant:
That does make more sense, sorry I misunderstood in the first place

martindurant:
Rather than testing whether the first input happend to be a Hashable, do you think it may be simpler to provide another argument? I'm thinking that there will be cases where the first item is hashable but subsequent ones are not.

CJ-Wright:
I can do that, although I'm not certain how much support we should give to heterogeneous streams, since usually functions don't handle more than one kind of data.

martindurant:
> streamz/core.py:3:1: F401 'collections.Hashable' imported but unused

martindurant:
Coming back to this now.

Are you certain that `deque` is the right structure here? I ask, because if we only test with `in`, then the position in the q is not updated, something can drop out of the history despite being seen relatively recently.

CJ-Wright:
Ah now I understand what you mean. Maybe we could refresh the position of the element in the queue. I like the deque structure because it handles things falling out of history nicely (we don't need to explicitly remove anything).

martindurant:
> like the deque structure because it handles things falling out

Right, but if the order isn't useful in this sense, or we have to move things around within a q, may as well use list.

In fact, it makes me wonder about the `in` operation in this sense - this will defer to the `hash()` function by default anyway, right (i.e., check `is` against each element)? So I'm not certain this has the right behaviour in any case for the non-hashable, non-hashfunction case.

martindurant:
ping

CJ-Wright:
I think that in checks equality if hash doesn't exist. For instance `{1:2} in [{1:2}, {3:4}]` is a valid statement, even though `hash({1:2})` is not.

martindurant:
mmm, ok.

Then I would request that this code be changed to keep a list instead of a deque, so that the correct order of historical uniques can be maintained.

CJ-Wright:
Ok, Do you have any suggestions on how to do the data dropping mechanic?

martindurant:
I'm afraid you'll have to do `del seen[history:]` every time; insert to the front and move to the front (`seen.remove(item); seen.insert(0, item)`). Or the same from the other end, if you prefer. This does not scale very well for very big lists.

(note that I would prefer to rename `history`, in a comment above)

CJ-Wright:
Ah, so I went back to produce a test which would get at the bug that you found with my previous implementation. However, I think the behavior for hashable and non-hashable are the same. The issue here is that checking membership of a thing in the `LRU` cache doesn't count as a "use" of the element.

For instance:
```python
from zict import LRU
lru = LRU(2, {}, on_evict=lambda k, v: print("Lost", k, v))
lru['a'] = 1
lru['b'] = 1
'a' in lru
lru['c'] = 1
```
will print that `a` has been lost.

We could want to change the current behavior so that the history gets updated.

martindurant:
`'a' in lru`  -> `lru['a']`?

CJ-Wright:
How about `lru.get('a', None)` that way we don't need to wrap it in a try except and if the result is None we know we need to add it to the cache.

martindurant:
That's fine too

CJ-Wright:
@martindurant I think this is ready for review again, I'm not certain why py2.7 is failing.

martindurant:
Hm, I don't seem to get notifications of commits on this PR. 
For py2 (although I don't really care): 
- tests that depend on distributed (master) should be skipped as it may contain py3-only syntax
- the rest looks like circular imports, which can likely be fixed by deferring, moving some imports into the methods that actually need them

CJ-Wright:
I don't care about the py2k tests either.

martindurant:
OK, so either those things should be fixed/skipped (I doubt it's more than half hour's work), or the build matrix should be updated to run only py36/7 - and then also the setup.py should reflect this change; that means we would need to remember to update the feedstock also when the time comes.

streamz/core.py
martindurant: The parameter name changed, but not this doc line (sorry :))

martindurant:
Done.

PR #248, opened 2019-05-07
FIX: connect doesn't break zip now

CJ-Wright:


CJ-Wright:
Ready for review!

streamz/core.py
```diff
@@ -181,6 +181,24 @@ def _inform_asynchronous(self, asynchronous):
                 if downstream:
                     downstream._inform_asynchronous(asynchronous)
 
+    def _add_upstream(self, upstream):
```
martindurant: Although this is internal API, I think one-line docstrings would be useful for developers.
Also, I suspect that the old way (self.upstream.downstreams.add/remove) is used in many places - do you think they ought to be changed to use this formalism now?

streamz/core.py
```diff
@@ -181,6 +181,24 @@ def _inform_asynchronous(self, asynchronous):
...
+        if self.upstreams == [None]:
+            self.upstreams = [upstream]
```
martindurant: Any reason to make a new list, instead of `self.upstreams[0] = upstream`?
I suppose you were just copying how the code looked previously.

streamz/core.py
```diff
@@ -181,6 +181,24 @@ def _inform_asynchronous(self, asynchronous):
...
+        self.downstreams.add(downstream)
+
+    def _remove_downstream(self, downstream):
+        self.downstreams.remove(downstream)
+
+    def _remove_upstream(self, upstream):
+        if len(self.upstreams) == 1:
+            self.upstreams = [None]
+        else:
+            self.upstreams.pop(self.upstreams.index(upstream))
```
martindurant: `self.upstreams.remove(upstream)`?

streamz/core.py
```diff
@@ -181,6 +181,24 @@ def _inform_asynchronous(self, asynchronous):
...
+
+    def _add_downstream(self, downstream):
+        self.downstreams.add(downstream)
+
+    def _remove_downstream(self, downstream):
+        self.downstreams.remove(downstream)
+
+    def _remove_upstream(self, upstream):
+        if len(self.upstreams) == 1:
+            self.upstreams = [None]
```
martindurant: `self.upstreams[0] = None` ?

streamz/core.py
```diff
@@ -927,6 +941,14 @@ def __init__(self, *upstreams, **kwargs):
 
         Stream.__init__(self, upstreams=upstreams2, **kwargs)
 
+    def _add_upstream(self, upstream):
+        self.buffers[upstream] = deque()
```
martindurant: So this is the meat of the change here, that instead of an upstream, we have a set of buffers for each upstream. Perhaps a comment in this method override would be useful explanation.

streamz/core.py
```diff
@@ -991,6 +1013,27 @@ def __init__(self, *upstreams, **kwargs):
...
+        self.missing.update([upstream])
+        if self.emit_on != self.upstreams:
+            super(combine_latest, self)._add_upstream(upstream)
+        else:
+            super(combine_latest, self)._add_upstream(upstream)
+            self.emit_on = self.upstreams
+
+    def _remove_upstream(self, upstream):
+        if self.emit_on == upstream:
+            raise RuntimeError("Can't remove the emit on upstream, consider"
```
martindurant: Can you make this message clearer?

streamz/core.py
```diff
@@ -991,6 +1013,27 @@ def __init__(self, *upstreams, **kwargs):
             self.emit_on = upstreams
         Stream.__init__(self, upstreams=upstreams, **kwargs)
 
+    def _add_upstream(self, upstream):
+        self.last.append(None)
+        self.missing.update([upstream])
+        if self.emit_on != self.upstreams:
```
martindurant: This equality check seems a little odd - can't we just store whether `emit_on` was originally given as `None`? Same comment for _remove_upstream.

streamz/core.py
```diff
@@ -991,6 +1013,27 @@ def __init__(self, *upstreams, **kwargs):
             self.emit_on = upstreams
         Stream.__init__(self, upstreams=upstreams, **kwargs)
 
+    def _add_upstream(self, upstream):
+        self.last.append(None)
```
martindurant: Again, I'd appreciate a comment for why an override is necessary here.

Have you now covered all classes which have multiple input streams?

streamz/core.py
CJ-Wright: I think the classes which use the old way should be migrated, otherwise they might run into issues with zip and such. I'll look to see if that pattern is used elsewhere.

martindurant: The new `slice` method has it, sorry: https://github.com/python-streamz/streamz/blob/master/streamz/core.py#L795

streamz/core.py
CJ-Wright: I haven't handled `zip_latest` yet.

martindurant: Is `zip_latest` out of scope for this PR, would you say?

streamz/core.py
martindurant: Would appreciate if you merged from master and made sure that `slice` works

streamz/core.py
CJ-Wright: Yeah, I don't use it that much (despite creating it). It turned out that it had a whole bunch of state storage issues that should most likely be managed first.

CJ-Wright:
@martindurant since `zip_latest` is out of scope, can this get another round?

streamz/core.py
martindurant: ping on this one

CJ-Wright: sorry about that, forgot it was there!

martindurant: ping...

CJ-Wright: I think this is taken care of now? https://github.com/python-streamz/streamz/pull/248/files#diff-d4e512993f47710ea9e7155d66011a24R813

martindurant: Oh I am sorry - I don't know why I didn't see your commits come in.

PR #251, opened 2019-05-15
Add support for more cudf methods with 0.7 

skmatti:
The latest release of cudf 0.7 allowed me to add support for some more cudf methods. Arithmetic operations on cudf DataFrames are now supported. I have added tests for those newly supported methods in cudf. 

GroupBy is not possible yet due to some implementation differences between Pandas and cudf. But I added some github issues for these blockers.

### DIFF
--- a/streamz/dataframe/core.py
+++ b/streamz/dataframe/core.py
@@ -13,7 +13,8 @@ from ..collection import Streaming, _stream_types, OperatorMixin
 from ..sources import Source
 from ..utils import M
 from . import aggregations
-from .utils import is_dataframe_like, is_series_like, is_index_like, get_base_frame_type
+from .utils import is_dataframe_like, is_series_like, is_index_like, \
+                    get_base_frame_type, get_dataframe_package
 
 
 class BaseFrame(Streaming):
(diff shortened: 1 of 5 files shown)


streamz/dataframe/core.py
```diff
@@ -270,17 +271,19 @@ class DataFrame(Frame, _DataFrameMixin):
     """
 
     def __init__(self, *args, **kwargs):
-        # {'x': sdf.x + 1, 'y': sdf.y - 1} - works only with pandas
+        # {'x': sdf.x + 1, 'y': sdf.y - 1}
         if len(args) == 1 and not kwargs and isinstance(args[0], dict):
-            def concat(tup, columns=None):
-                result = pd.concat(tup, axis=1)
+            def concat(tup, module=None, columns=None):
```
martindurant: Instead of making a function inline, would it not make sense to store the module as an instance attribute, and have this as a method? Then you don't have to call get_dataframe_package multiple times.

streamz/dataframe/core.py
```diff
@@ -270,17 +271,19 @@ class DataFrame(Frame, _DataFrameMixin):
...
                 return result
 
             columns, values = zip(*args[0].items())
+            base_frame_type = values[0]._subtype
+            df_package = get_dataframe_package(base_frame_type)
             stream = type(values[0].stream).zip(*[v.stream for v in values])
-            stream = stream.map(concat, columns=list(columns))
-            example = pd.DataFrame({k: getattr(v, 'example', v)
+            stream = stream.map(concat, module=df_package, columns=list(columns))
+            example = df_package.DataFrame({k: getattr(v, 'example', v)
```
martindurant: Is it absolutely required that the example be cudf if the data is?

streamz/dataframe/core.py
martindurant: Right, OK - that's annoying!

PR #254, opened 2019-05-29
Add docs and examples for cudf support

skmatti:
This adds code documentation and notebook examples for streaming cudf dataframes using Streamz DataFrame.


skmatti:
This is failing with Python2.7 but not sure if the failure is due to this PR.

martindurant:
For py2, it may be worth pinning to an earlier version of distributed - there will be some non-py2 usages creeping into that codebase.

martindurant:
Please merge from master - py2 will no longer be tested at all, so this should then pass.

skmatti:
Yes! it's green.

PR #267, opened 2019-08-18
Update build status image url in readme

mrafayaleem:
The current url doesn't seem to be showing the correct build status. This change fixes that.


CJ-Wright:
LGTM thank you very much!

PR #268, opened 2019-08-19
Add groupby aggregate tests, along with a few additions for cudf integration.

### LINKED ISSUE #266: Index name lost during groupby aggregates on Streamz Dataframes.
I am trying to do a simple groupbySum using Streamz Dataframes over cudf Dataframes. I am getting an issue wherein the result is correct, but the index name ('x' in this case) is lost. This issue is not seen with Pandas. 

I am trying to peruse through the code to see where I would have to make changes to make this work. It would be great if someone could help me out as to where exactly the GroupBy class methods do the index name check happen? 

The code to reproduce the bug is:
 
```
from streamz import Stream
from streamz.dataframe import DataFrame
import cudf
import numpy as np
import pandas as pd

df = cudf.DataFrame({'x': (np.arange(10) // 2).astype(float), 'y': [1.0, 2.0] * 5})

stream = Stream()
a = DataFrame(example=df.iloc[:0], stream=stream)

def f(x):
    return x.groupby('x').sum()

L = f(a).stream.gather().sink_to_list()

a.emit(df.iloc[:3])
first = df.iloc[:3]

print(L[0])
print(f(first))
```

The outputs I get are pasted below.
```
       y
0.0  3.0
1.0  1.0
       y
x       
0.0  3.0
1.0  1.0
```

chinmaychandak:
Most changes made to aggregations.py attempt to fix #266. Somehow, as of now, cudf does not do the index naming implicitly like Pandas. 

Other changes in the file attempt to create a temporary fallback on Pandas Timedelta API to perform window-over-time-groupby aggregations.


streamz/dataframe/tests/test_cudf_dataframes.py
```diff
@@ -33,23 +32,23 @@ def client():
         client.close()
 
 
-@pytest.fixture(params=['core', 'dask'])
+@pytest.fixture(params=["core", "dask"])
```
jsmaupin: Are we standardizing on double quotes? From what I've seen, Python code has always used single quotes for strings unless there was a reason to do otherwise.

streamz/dataframe/aggregations.py
```diff
@@ -347,8 +348,8 @@ def windowed_groupby_accumulator(acc, new, diff=None, window=None, agg=None, gro
     for o, og in zip(old, old_groupers):
         if 'groupers' in acc:
             assert len(o) == len(og)
-            if hasattr(og, 'index'):
-                assert (o.index == og.index).all()
+#             if hasattr(og, 'index'):
+#                 assert (o.index == og.index).all()
```
jsmaupin: I would rather not check in commented out code.  Just delete the code and if we need to restore it will be in the file history in Git.

streamz/dataframe/aggregations.py
```diff
@@ -202,10 +203,10 @@ def diff_loc(dfs, new, window=None):
...
     dfs = deque(dfs)
     dfs.append(new)
-    mx = max(df.index.max() for df in dfs)
-    mn = mx - window
+    mx = pd.Timestamp(max(df.index.max() for df in dfs))
+    mn = pd.Timestamp(mx) - window
     old = []
-    while dfs[0].index.min() < mn:
+    while pd.Timestamp(dfs[0].index.min()) < mn:
         o = dfs[0].loc[:mn]
```
jsmaupin: I'm curious as to how casting to a pandas timestamp helps here.

streamz/dataframe/aggregations.py
```diff
@@ -202,10 +203,10 @@ def diff_loc(dfs, new, window=None):
     """
     dfs = deque(dfs)
     dfs.append(new)
-    mx = max(df.index.max() for df in dfs)
-    mn = mx - window
+    mx = pd.Timestamp(max(df.index.max() for df in dfs))
+    mn = pd.Timestamp(mx) - window
```
jsmaupin: Isn't mx already a pd.Timestamp type, why pass it into another pd.Timestamp(...) type?

streamz/dataframe/aggregations.py
chinmaychandak: For now, cudf's `df.index.min()` (or max()) returns a `numpy.datetime64` as opposed to `pandas._libs.tslibs.timestamps.Timestamp`. Hence the explicit cast to make it compatible with Pandas Timedelta, which is required for these operations. 

The statements modified for this purpose would be redundant for Pandas, since the types are compatible with Pandas Timedelta.

streamz/dataframe/tests/test_cudf_dataframes.py
chinmaychandak: Sure, will fix this.

chinmaychandak: Fixed

streamz/dataframe/aggregations.py
chinmaychandak: Fixed

streamz/dataframe/aggregations.py
chinmaychandak: Fixed

chinmaychandak:
@martindurant Could you please review my code and merge it soon, if possible? I'd be happy to make any changes you think would be necessary.

chinmaychandak:
Hey @CJ-Wright, could you please have a look at this? I'd appreciate it if this could be merged soon! :)

CJ-Wright:
I'll try to look at it soon (I need to read in omre of the dataframe things)

chinmaychandak:
That would be great, thanks @CJ-Wright!

chinmaychandak:
@CJ-Wright Did you have a chance to look at this yet? This is a major blocker for a bigger project! :(

streamz/dataframe/tests/test_cudf_dataframes.py
CJ-Wright: If we are going to go this way we should commit to black and be done with it IMO.

CJ-Wright:
Seems reasonable to me. I wish we had codecov on this repo.

chinmaychandak:
> Next time please don't make style changes (single vs double quote) to code that you aren't substantially changing, it makes review a bit more difficult, since I need to parse which lines are logic changes and which are style changes.

Sure, will definitely keep this in mind moving forward.

@CJ-Wright Thanks a lot for reviewing and merging this, really appreciate it! Is it possible to upload the updated conda package with these changes?

CJ-Wright:
We'd need to cut a release. Can you open up an issue for this?

chinmaychandak:
Sure, will do that. Thanks!

PR #272, opened 2019-09-06
Add checkpointing conforming to Spark Streaming to from_kafka_batched().

chinmaychandak:
This PR serves two purposes: 

1. Remove all GPU-dataframe/cudf-related tests, since they would now live as part of [RAPIDS' cuDF](https://github.com/rapidsai/cudf). 

2. Implement a checkpointing feature (conforming as much as possible to Spark Streaming) in `from_kafka_batched()`. The checkpointing feature is paramount to any stream processing pipelines today in production; below is the design of the current checkpointing feature in this PR (although I expect this to get modified and become more robust with time and review(s)).

**Checkpointing:**
a. `from_kafka_batched()` has another parameter `checkpointing` which defaults to `None`. 
b. Users wanting to opt out of checkpointing can use streamz as is —  that is the consumers would read the Kafka topic from the starting offsets every single time.    
c. Users who want to use checkpointing in their pipelines can provide a path in their checkpointing parameter in `from_kafka_batched()`. If a directory doesn't exist at this path, streamz would now create one — this is where checkpoints would live. 
d. The directory structure inside this path conforms to Spark Streaming. There would be topic-level directories, and then checkpoints within these sub-directories. Checkpoints would be named after the batch number they come from, i.e., `5.txt` would be the 5th batch that streamz processed. 
e. A checkpoint file in each topic would be a text-file with a nested JSON of the form {'group.id': {partition:offset, partition:offset, ...}, 'group.id':{.., ..}}. An example is shown below:
`{'custreamz': {'0': 1, '2': 2, '6': 1, '10': 2, '11': 1, '12': 1, '19': 1, '20': 1}}`. Here, `custreamz` is the `group.id`. 
f. Like Spark Streaming, only the last few (in this PR, I have hard-coded this number to 5) checkpoints are maintained, the older checkpoints are deleted.
g. Now, if the stream is restarted, or shut down/killed for some reason,  and a consumer wants to start from the latest offsets, it can just specify the checkpoint path in `from_kafka_batched()` and the stream starts reading from the latest checkpoint stored within.

Below is a checkpointing consumer test notebook. Please have a look at the example outputs to understand how the checkpointing is working. Also, please **note** that I am printing the latest checkpoints as they get saved for now, to verify the correctness. As things get more approvable, I will make the code more mergeable.  
[checkpointing_consumer_test.ipynb.zip](https://github.com/python-streamz/streamz/files/3585082/checkpointing_consumer_test.ipynb.zip)

I would also appreciate thoughts on how/what test cases I could add.


martindurant:
Hm, I'm really not sure here - this seems like a much more general point that we might want to apply to any state-holding stream. Thoughts, @CJ-Wright ?

jsmaupin:
> Hm, I'm really not sure here - this seems like a much more general point that we might want to apply to any state-holding stream. Thoughts, @CJ-Wright ?

Are you talking about a generalized checkpointing solution? If yes, I think we can only create an abstract solution as each data source is likely going to require a different method.

chinmaychandak:
I agree with Jarod.

Checkpointing with Kafka as a source is one of the most important things today in streaming production use cases, hence this first step.

CJ-Wright:
Is this a general problem which each source solves with its own syntax? If so then we might consider having a solution which has the same streamz syntax but caters to each source in its own syntax.

chinmaychandak:
@CJ-Wright Yes, that's right. Spark provides checkpointing for a few sources — Kafka, SQS, S3, etc. But since streamz only has a single source used in production pipelines today — Kafka seems like the priority for now. Kafka consumers in the same group would share offsets, and can be resilient to streamz/Kafka failures this way. We can probably add checkpointing for text files and sockets later on, but this would be different syntax altogether, since streamz APIs for reading from Kafka and textfiles is different.

CJ-Wright:
Is it possible to de-couple this from the source itself? So that there is a checkpoint node? Is this an accumulate node which writes to a file somewhere?

chinmaychandak:
> Is it possible to de-couple this from the source itself?

There's a couple of things here, IMO. Having an API like `stream.from_kafka_batched(.., ..).checkpointing(path)` would decouple stuff, but would involve passing around a lot of information. Moreover, there would still not be a generic checkpointing implementation across sources. 

I'd be happy to implement anything cleaner in terms of API if someone can help me out here!

CJ-Wright:
When you say "still not be a generic checkpointing implementation" do you mean that each source has its own syntax or that each has their own concerns which are not expressible in a single syntax?

chinmaychandak:
The only other source I think it would make sense to implement checkpointing for would be textfiles. I meant that the checkpointing implementation for textfiles would be completely different from the one for Kafka — it would require information from `from_textfile()`. So, it's easier implemented if `checkpointing` is passed a parameter in API rather than decoupling.  

Apologies for not being clearer earlier.

chinmaychandak:
Another thing to note here is that Spark, Flink has the same syntax to read from different sources, so it is easy to decouple checkpointing. Since this is not the case with streamz, I'm not sure whether decoupling checkpointing would be the correct move here.

martindurant:
I would make a lot of sense to me from an API point of view, if there was a common `.checkpoint()` method which certain sources knew how to implement - i.e., this is an ordinary method, not a stream operation. They should probably return the source itself, to keep a similar behaviour with other stream nodes (although I don't think that the checkpointing should be counted as a node); maybe sources that don't checkpoint just return the node as a no-op.

CJ-Wright:
It might be worthwhile having `checkpoint` as a node, since that would allow any graph walking things to access it (eg. build a walker which builds pieces of a gui for all the params of a pipeline)

martindurant:
You have a point, but it also (to my mind) breaks some assumptions, by directly affecting how the upstream node works.

CJ-Wright:
That's fair. Maybe we need to document that this is different, since it would behave different than other `foo.bar()` syntax things.

chinmaychandak:
Closing this PR for now. 

After a few discussions, I realized that this implementation of checkpointing is logically incorrect  — checkpointing should be implemented after writing to a sink (Streamz only has Kafka for now), i.e., so that we can keep track of how much data has been processed. 

Currently, I was adding checkpoints as soon as data was read using from_kafka_batched(). Spark, too, only has write-stream checkpointing in order to not lose track of how much data has been 'processed'. @jsmaupin Do you want to add something to this?

I will fix this, and reopen a PR soon.

PR #273, opened 2019-09-09
cover tests

CJ-Wright:


CJ-Wright:
@martindurant I think it would be good to have coverage of the tests. We can `pragma: no_cover` things that we won't cover.

martindurant:
Is it normal to consider tests as part of coverage? We would want to know of tests that didn't run, but it doesn't seem to be a useful metric for the package as a whole. Actually, we have two separate numbers in the report, so probably don't need a limit on both. 

I don't really mind though - but had better indeed mark some things we don't use. I don't know why the coverage appears to have gone down significantly.

CJ-Wright:
I've been trying to exclude the test chunk from the library chunk.
I think the coverage is low because we don't run many of the dataframe tests.

The coverage of the tests help make certain that we are running the tests, and show if we miss some chunks of test for some reason or another. (I stole the target from matplotlib)

martindurant:
Whilst I agree,
1) So why are we not running those DF tests?
2) why is the coverage diff negative?

CJ-Wright:
1. Many of the tests are cudf tests, which we can't really run on travis. We can leave them in, but we'll need to list them as no cover.
1. When the not covered tests are averaged in with the covered things it drops the percentage (I'm still working on test exclusion)

martindurant:
OK, so you can blanket ignore cuDF tests. It would be nice if coverage had a mechanism like pytest marks to set this automatically.

martindurant:
please fix the merge conflict, if you have the time.

chinmaychandak:
I guess we can close this now, since we have `#pragma no cover` and also `flaky` markers for tests? Or are there places where we can improve?

CJ-Wright:
Do those cover the purpose of this PR? This PR was supposed to add an additional check to see the coverage of the test code itself. That way we could be certain that the entire test suite was being run (in case we missed anything and pieces of the tests were actually not being run)

chinmaychandak:
Oh, I understand now. I misunderstood the purpose of this PR then. Apologies!

CJ-Wright:
No problem, I should really merge this with master

chinmaychandak:
We no longer have any `cuDF` tests in streamz, though, so the coverage is back on track. And branches of execution which are not tested because of cuDF/cuStreamz-specific functionality have been marked with `#pragma no cover`. Maybe that should be enough? How can we check if we still need to cover tests?

PR #281, opened 2019-11-14
add installation docs

CJ-Wright:


docs/source/index.rst
```diff
@@ -66,6 +66,14 @@ flow of data.  Typically people rely on tools like ``itertools.tee``, and
 However this quickly become cumbersome, especially when building complex
 pipelines.
 
+Installation
+------------
+
+To install either use:
+
+- conda-forge: ``conda install streamz -c conda-forge``
+- pip: ``python -m pip install streazm``
```
martindurant: typo here; also, I prefer `pip install astreamz` alone.

docs/source/index.rst
```diff
@@ -66,6 +66,14 @@ flow of data.  Typically people rely on tools like ``itertools.tee``, and
...
+- local pip: ``git clone git@github.com:python-streamz/streamz.git; python -m pip install -e streamz``
```
martindurant: `pip install git+https://github.com/python-streamz/streamz` ?

CJ-Wright: Does that install a local editable copy? (I'm mostly thinking of future streamz developers here)

docs/source/index.rst
CJ-Wright: Does that run into issues with the wrong pip being used?

docs/source/index.rst
martindurant: No, it uses a temporary copy.
Instead, then, I would say
```
- dev: git clone https://github.com/python-streamz/streamz
       pip install -e streamz/
```

docs/source/index.rst
martindurant: Not in my experience. If people have pip on their path from a different environment than python, then I suspect other things would go wrong.

CJ-Wright:
@martindurant now I think this is ready (it also fixes some networkx bugs)

CJ-Wright:
Thank you!

PR #283, opened 2019-11-25
Add filter class to dask and do the tests for it

xuzifan08:
Added filter class to dask.py and four tests associated with it.

### DIFF
--- a/streamz/dask.py
+++ b/streamz/dask.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, division, print_function
+from functools import wraps
 
+from .core import _truthy
 from operator import getitem
 
 from tornado import gen
@@ -9,6 +11,21 @@ from distributed.client import default_client
 
 from .core import Stream
 from . import core, sources
+
+
+NULL_COMPUTE = "~~NULL_COMPUTE~~"
+
+
+def return_null(func):
+    @wraps(func)
+    def inner(x, *args, **kwargs):
+        tv = func(x, *args, **kwargs)
+        if tv:
+            return x
+        else:
+            return NULL_COMPUTE
+
+    return inner
 
 
 class DaskStream(Stream):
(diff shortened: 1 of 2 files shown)


streamz/tests/test_dask.py
```diff
@@ -131,6 +131,67 @@ def test_buffer(c, s, a, b):
     assert source.loop == c.loop
 
 
+@pytest.mark.slow
+def test_filter(backend):
```
CJ-Wright: You need to remove the `backend` arg here, and in the `scatter` statement. Streamz only has a dask backend at the moment.

streamz/dask.py
```diff
@@ -140,6 +157,24 @@ def update(self, x, who=None):
         return self._emit(result)
 
 
+@DaskStream.register_api()
+class filter(DaskStream):
```
CJ-Wright: I think you also need the modifications to the gather and other nodes as well.

streamz/tests/test_dask.py
xuzifan-lab: Should I remove the backend for all the four tests?

xuzifan08: Should I remove the backend for all the four tests I have added?

streamz/dask.py
xuzifan08: I made changes to gather already. I also compared other nodes. There is a slightly difference between the dask.starmap and parallel.starmap. Do I need to change that one?

xuzifan08: I made changes to gather already. I also compared other nodes. There is a slightly difference between the dask.starmap and parallel.starmap. Do I need to change that one?

streamz/tests/test_dask.py
CJ-Wright: yes please

xuzifan08: Already changed the four tests but the build is still failed

xuzifan08:
Hi, @martindurant @mrocklin, this is Zifan. I'm working on this branch with @CJ-Wright . 

We added class filter() to dask and modified class gather() according to parallel.py (including the associated tests). However, we ran into an issue which causes the test test_integration_from_stream fail in the test_dataframe.py. The test passes on the Stream but fails on Daskstream. After troubleshooting for a while, we guess there is something wrong with the process of serializing the code to different dask clusters, but we are not sure what the specific issue is. 

Can we pick your brains about how we can resolve it? Thanks!

martindurant:
Sorry for not following up here - are you still interested in this, @xuzifan08 ?

PR #286, opened 2019-12-03
Added base Conda environment yml file

jdye64:


jdye64:
I think more docs is always great! Specifically which doc building things are you referring? I can add them pretty quickly if you can point me to them.

CJ-Wright:
https://github.com/python-streamz/streamz/blob/master/docs/requirements-docs.txt

jdye64:
Ah, yes I misunderstood what you were talking about. I will add those now.

conda/environments/streamz_dev.yml
```diff
@@ -0,0 +1,23 @@
+name: streamz_dev
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - python>=3.7
+  - cython>=0.29,<0.30
```
CJ-Wright: Where is the cython dep coming from?

CJ-Wright:
@martindurant thoughts?

conda/environments/streamz_dev.yml
martindurant: cytoolz? I don't suppose we're building anything.

martindurant:
I'm generally OK with the idea, but wonder whether the CI tests should explicitly use the same environment.

conda/environments/streamz_dev.yml
CJ-Wright: Right, if one of the deps needs it conda should figure that out.

jdye64: ahh ... my bad I meant to remove that. I was experimenting with implementing librdkafka with cython and forgot to remove it. I will remove them now.

On a side note that is working very well BTW and might open something up around that as well. Working on the best way to make cython not a hard dependency at the moment since I'm sure several people don't using streamz don't necessarily want it.

martindurant: > Working on the best way to make cython not a hard dependency

Typically you can include the C file for those that want to compile (source dist) and of course, if you build wheels and conda packages, those should not need cython.

CJ-Wright: I'm confused, do you mean that you are using librdkafka with cython or using cython to speed up `streamz` kafka bindings?

jdye64: > I'm confused, do you mean that you are using librdkafka with cython or using cython to speed up `streamz` kafka bindings?

Actually both. I was replacing the current kafka sources with my own sources (not the confluent-python-kafka library) which were cython modules that interacted directly with librdkafka++. There are many many advantages to that especially in the realm of partition balancing that the new KafkaConsumer offers (not implemented in the confluent python library yet) for distributed systems with many processes appearing/disappearing.

In a nutshell I was seeing about 15-20X speedups in my consumers when using my cython implementations.

CJ-Wright:
@jdye64 would you be interested in changing the `.travis.yaml` to use the `environment.yml` file to build the testing env?

jdye64:
Sure. I noticed there are a few other dependencies in the .travis.yml file that were being installed so I'm going to add those to the environment.yml file as well.

jdye64:
@CJ-Wright Ok had to step away for a bit but looks like its all good to go and using the conda environment file correctly in Travis now.

CJ-Wright:
Thank you very much!

Issue #287, opened 2019-12-03
Dockerfile for Streamz


jdye64:
Offering a Dockerfile would help in a couple of ways.

1. Although already easy, would make for an even more simple "quick start" for users.
2. Testing new features with more complicated dependencies would be easier.
3. Act as a reference for users to view start to finish setup and installation by viewing the operations performed by the Dockerfile.

I'm more than happy to take this on.

PR #288, opened 2019-12-03
Initial streamz docker image to lower barrier for users accessing str…

jdye64:
…eamz exmaples. Closes: #287

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,6 @@ EXPOSE 2181
 EXPOSE 9092
 
 # Install Jupyter-lab so that users can quickly interact with the streamz examples
-RUN pip install jupyterlab
+RUN pip install jupyterlab numpy pandas
 
 CMD ["/streamz/docker/scripts/entry.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,9 +21,19 @@ SHELL ["/bin/bash", "-c"]
 # Add Streamz source code to the build context
 ADD . /streamz/.
 
+# Create the conda environment
+RUN conda env create --name streamz-dev -f /streamz/conda/environments/streamz_dev.yml
+
 # Build streamz from source
-RUN cd /streamz && \
+RUN source activate streamz-dev && \
+    cd /streamz && \
     python setup.py install
+
+# Install optional dependencies in the conda environment
+RUN source activate streamz-dev && \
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -2,3 +2,5 @@ tornado
 toolz
 zict
 six
+numpy
```
CJ-Wright: These are optional deps I think so we might consider if they are needed or not.

jdye64: That is a good point. Let me remove them from requirements.txt and rather just install them with pip directly in the Dockerfile itself. 

If the conda PR merges at another point that could also be used instead to make the Dockerfile even cleaner.

jdye64:
@CJ-Wright I modified the Dockerfile to use the conda environment from the optional dependencies now instead of pip. This also ensures that as things are added to the conda environment file over time they will be automatically added to the Docker container at build time.

I also added some documentation on building in docs/index.rst. Let me know what you think.

martindurant:
@CJ-Wright , were you reviewing this?

martindurant:
ping @CJ-Wright

CJ-Wright:
Sorry for the delay

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+USER root
+
+ENV DEBIAN_FRONTEND noninteractive
+ENV SCALA_VERSION 2.11
+ENV KAFKA_VERSION 2.3.0
+ENV KAFKA_HOME /opt/kafka_"$SCALA_VERSION"-"$KAFKA_VERSION"
+
+# Install system dependencies and convenience utilities
+RUN apt-get update -y && \
+    apt-get install -y wget \
```
CJ-Wright: Is there any reason not to get these two from conda-forge?

CJ-Wright:
This looks good to me, one minor comment.

jdye64:
Sorry for the major delay folks. Yep agree its dumb of my to not just add them via conda. I'll make that change and commit now after a validation build.

CJ-Wright:
cycle for ci

CJ-Wright:
This is failing on flake8 I'll push a fix to the branch

CJ-Wright:
I don't know why flake8 is unhappy some of the offending lines don't even exist at the most recent commit.

martindurant:
I see no reason not the merge this - @CJ-Wright , agreed?

PR #290, opened 2019-12-09
STREAMZ-289: Upgraded version of python in Travis CI from 3.6 -> 3.7.…

jdye64:
… This closes: 289

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ language: python
 
 matrix:
   include:
-    - python: 3.6
+    - python: 3.7
 
 env:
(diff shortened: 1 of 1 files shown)


CJ-Wright:
To be fair do we actually use the python version that travis supplies? (since we are now using the envrionment.yaml)

jdye64:
Honestly I’m not enough of a Travis expert to know if it’s using that in some other capacity or not? Thought maybe just be on the safe side

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

PR #294, opened 2019-12-11
STREAMZ-293: Added vscode to list of gitignore directories

### LINKED ISSUE #293: Add .vscode to .gitignore
I personally use vscode and I think some others do as well. Thought it would be worth adding those vscode environment specifics to the .gitignore so they don't accidentally get checked into the codebase.

jdye64:
This closes #293


CJ-Wright:
Thank you!

PR #297, opened 2020-01-11
Added checkpointing

jsmaupin:
This works by creating a future at the source for each batch.  The futures are passed down the pipeline with the data.  Each type of function does something different depending on that function's operation.  Some functions will mark a future as down with 'drop' to tell the source not to commit the offsets. Others, such as `map` will pass the future downstream without affecting it.


martindurant:
Thanks for the large PR. I can't promise when I can give it a thorough review, but I will keep it on my list. @CJ-Wright , any thoughts here, in the light of the long checkpointing discussion? Without looking at the code, I am surprised that futures should be passed *forward* rather than as for back-pressure.
@jsmaupin , I note that CI is still failing.

CJ-Wright:
I'll take a longer look.
I think some prose documentation would be great so we can better understand what is going on in the code.

CJ-Wright:
How does this work with things like `combine_latest` where you want to commit the offset only when it leaves the cache? Does a single future only represent a datum from one source or is that the entire state of the graph? If it represents the data from one source do the futures need to be cached for nodes that cache?

CJ-Wright:
Overall I think that the concept of the pipeline declaring when it is "done" with a piece of data to be quite valuable and currently not very intuitive, especially when considering caching nodes.

This could help provide provenance for non-deterministic pipelines, since you could be confident that a piece of data was no longer needed, even if you couldn't replay the analysis from scratch.
You could also consider building this into some status/dashboarding system showing exactly what data was being used at any point in time.

jsmaupin:
After looking at this pull request, we have realized that we aren't handling the case where there are multiple down-streams very well.  After discussion we have come up with a plan and we are hoping to get feedback on this idea.  

Firstly, a reference to the source will be kept in each of the components of a stream.  Each stage will know the source of the data.  This way, when a downstream is attached, the source can be notified and a future for each downstream pipeline can be created.  Likewise, when the downstream is disconnected, the source will be notified.  

When the data is pushed downstream and a case is encountered where there are multiple down-streams, the futures will be divided evenly among them.  The offsets will be committed once all futures have been marked as done.

CJ-Wright:
>Firstly, a reference to the source will be kept in each of the components of a stream. Each stage will know the source of the data. This way, when a downstream is attached, the source can be notified and a future for each downstream pipeline can be created. Likewise, when the downstream is disconnected, the source will be notified.

I don't know if you need to keep a reference to the source at each node. You could have the init of the downstream sinks walk the graph and notify any relevant sources, same thing for attaching and removing nodes.

Otherwise this seems reasonable, I think you may need a different name, since futures currently correspond to the objects going up the call stack rather than these which go down.

Also how are you planning on handling caching nodes?

jsmaupin:
> I don't know if you need to keep a reference to the source at each node. You could have the init of the downstream sinks walk the graph and notify any relevant sources, same thing for attaching and removing nodes.
We were considering this. We thought saving the reference would be more efficient, but I'm fine with walking the graph.
> Also how are you planning on handling caching nodes?
If you're talking about passing the tracking objects, they get cached with the data.

chinmaychandak:
> Also how are you planning on handling caching nodes?

@CJ-Wright By caching nodes, do you mean `partition()` or `window()` operations? In these cases, the future is only passed on when the window/partition size is complete; partial futures when the partition or window is being updated to be passed along the pipeline, are dropped.

@jsmaupin Please correct me if I am wrong here.

CJ-Wright:
I'm mostly worried about things like `zip` and `combine_latest` which need to keep a cache of data for at least some of the incoming streams. For instance one stream's data is done, but the other stream's data is being kept in a cache to they can be reused.

CJ-Wright:
> We were considering this. We thought saving the reference would be more efficient, but I'm fine with walking the graph.

I think it is a bit cleaner for members of the graph to not have to keep track of the global graph state (in this case the identities of the data sources) unless the have to (in the cases of sources and sinks).

jsmaupin:
So, for `combine_latest`, for example, the futures are all dropped until we pass data forward.  Let's say we drop 4 data frames. Those futures are marked as done with a result of 'drop'. The offsets are not committed. Then, on the 5th frame, which is passed downstream, the offset is passed along with it. If this is the behavior, then are we maintaining any state?

CJ-Wright:
Hmm, maybe I don't understand properly how would it work in this case?

```python
from streamz import Stream
s1 = Stream()
s2 = Stream()

pipeline = s1.combine_latest(s2).sink(print)

s2.emit(0)
s1.emit(0)
s1.emit(1)
s1.emit(2)
```

My understanding is that if both s1 and s2 were kafka sources then s1.emit(0) would mean that s1 could commit an offset once the computation was finished. Is that not correct?

jsmaupin:
It seems that the way the code is now, the future associated with the first `emit` is dropped, while the next three are passed forward.  I suppose what should be happening is that the future from the first emit his held until the data changes on `s2`?

CJ-Wright:
Right, that way if another emit came into s2 we could let consumers know that we had finished with the current s2 value.

I think that pattern would also scale to the other caching nodes. On emit you could check if any values that are in the futures cache are no longer in the actual cache, if true then let consumers know that the value is now finished.

Would there be any interest in expanding this from futures to metadata? That way additional information can be passed along with the data, including something that links a piece of data to an external cache. At that point caching source/sink pairs would just have some specific metadata that they write/access.

CJ-Wright:
Essentially this is fancy reference counting. I don't think we can use actual reference counting since some sinks may increase the reference count themselves (for instance `sink_to_list`). Really we want to keep track of only the number of references to a piece of data coming from pieces of the pipeline. When a pipeline no longer refers to a piece of data (or any of its derivatives) then that data is finished.

I don't know if I have a fully formed approach to this yet.

CJ-Wright:
Maybe the approach is

1. When data is produced by a source also create an entry in the reference counter thing
2. When/if that data comes to a cache increment the reference count for that data
4. When the data is removed from the cache decrement the counter
5. When the data comes to a sink, check the reference count if zero let consumers know the data is finished.

The main issue with this approach is associating any derived/merged/transformed data with its source. This will mostly need some merging of the metadata (for instance a datum coming out of zip will need to include reference counts for both incoming datums).

jsmaupin:
> Would there be any interest in expanding this from futures to metadata? That way additional information can be passed along with the data, including something that links a piece of data to an external cache. At that point caching source/sink pairs would just have some specific metadata that they write/access.

Personally, yes. I would like to see this happen

So, the reference counter is basically a hash lookup?  This sounds like it would replace the current futures strategy?

CJ-Wright:
Right, I think this would replace the futures system.

There may need to be some increment and decrements around the emit method so that this doesn't trigger early.

satishvarmadandu:
@CJ-Wright after discussing with @jsmaupin , @chinmaychandak .. we believe we don't need to replace existing futures system. We might need to handle futures for combine_latest() bit differently. @jsmaupin would be summarizing

CJ-Wright:
Sounds good!

jsmaupin:
Yes, I think the behavior of the current implementation is just a bug in the `combine_latest` function. It just needs to be changed to hold on to the future of the `s2` stream in @CJ-Wright 's example until another element is emitted through `s2`.  

So, to illustrate:

```python
s2.emit(0)  # combine_latest will hold on to the future passed here
s1.emit(0)  # combine_latest will emit (0, 0), but will not pass the future forward for s2
s1.emit(1)  # combine_latest will emit (1, 0), but will not pass the future forward for s2
s1.emit(2)  # combine_latest will emit (2, 0), but will not pass the future forward for s2
s2.emit(1)  # combine_latest will emit (2, 1) and will release the future that was held for s2
```

CJ-Wright:
Does this handle the `zip` case as well?
```python
s1 = Stream()
s2 = Stream()

s1.zip(s2).sink(print)

s1.emit(0)
s1.emit(1)
s1.emit(2)
s2.emit(0)
s2.emit(1)
```

chinmaychandak:
> Does this handle the `zip` case as well?

@CJ-Wright Yes, I would say so. 

@jsmaupin would be sharing a spreadsheet of how the futures are being handled in other nodes as well. This would be awesome for everyone to review and debug in case we are currently missing out anything.

streamz/core.py
```diff
@@ -532,8 +540,12 @@ def __init__(self, upstream, func, *args, **kwargs):
         Stream.__init__(self, upstream, stream_name=stream_name)
         _global_sinks.add(self)
 
-    def update(self, x, who=None):
+    def update(self, x, futures=None, who=None):
         result = self.func(x, *self.args, **self.kwargs)
+        if futures:
+            for future in futures:
+                if not future.done():
+                    future.set_result(None)
```
CJ-Wright: What happens if there are multiple sinks?

My understanding of this line is that this sink says the datum is finished, even if there are more sinks that need to be processed after this one.

jsmaupin: Yes, this is a problem.  I have previously commented on how we still need to solve the multiple down-streams and provided a potential solution in the PR

CJ-Wright: In your solution, how do you know if all futures are done?

jsmaupin: I'm not sure I get your question.  The source will wait on them all.

CJ-Wright: Ok so the source needs to keep a reference to all the futures, including the ones created partway through the pipeline.

jsmaupin: Yes

jsmaupin:
Closing in leu of the reference counting PR

PR #304, opened 2020-02-24
Read key along with value for messages from Kafka

chinmaychandak:
Most streaming data processors like Spark read both the key and value for Kafka messages. Most streaming jobs have Kafka message keys as Null, but some of them which need custom load balancing/data ordering across partitions, work more efficiently with keys.

### DIFF
--- a/streamz/dataframe/tests/test_dataframes.py
+++ b/streamz/dataframe/tests/test_dataframes.py
@@ -388,7 +388,7 @@ def test_setitem_overwrites(stream):
     pytest.param({}, 'count', marks=pytest.mark.slow),
     ({'ddof': 0}, 'std'),
     pytest.param({'quantile': 0.5}, 'quantile', marks=pytest.mark.slow),
-    ({'arg': {'sum', 'min'}}, 'aggregate')
+    ({'arg': {'x':'sum'}}, 'aggregate')
 ])
 @pytest.mark.parametrize('window', [
     pytest.param(2, marks=pytest.mark.slow),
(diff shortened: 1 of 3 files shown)


martindurant:
Seems like a reasonable change, although it may be better to
- have the output as a dictionary, `{'key': , 'value': }`
- have the extraction of the key be optional, so that we maintain the earlier simpler behaviour for the majority that don't care about they key.

Also, this could use a test where a specific key is passed.

chinmaychandak:
@martindurant Thanks a lot for responding! 

I have made the output a dictionary, and also updated the tests to reflect custom key extraction. 

As for making the extraction of keys optional, we would have to pass a `keys: True` parameter in `from_kafka_batched()` which would then get passed along to `get_message_batch()` for every tuple which gets added to the stream, which seems unnecessary. Do you have a better option in mind?

chinmaychandak:
Also, with the rolling "aggregate" function removed because of deprecated way of renaming of columns, the coverage seems to gone down. Any way we can avoid that?

martindurant:
Find something else to add coverage to? :)
I think we can live with the decreased coverage.

> As for making the extraction of keys optional, we would have to pass a keys: True parameter in from_kafka_batched() which would then get passed along to get_message_batch() for every tuple which gets added to the stream, which seems unnecessary. Do you have a better option in mind?

^ @CJ-Wright , thoughts on this kind of thing?

chinmaychandak:
> I think we can live with the decreased coverage.
Yay! 

> As for making the extraction of keys optional, we would have to pass a keys: True parameter in from_kafka_batched() which would then get passed along to get_message_batch() for every tuple which gets added to the stream, which seems unnecessary. Do you have a better option in mind?

@martindurant, @CJ-Wright: On second thoughts, we are passing `kafka_params` for every tuple anyway, so I think passing the `keys` boolean should be okay as well.

chinmaychandak:
@martindurant I have made extracting the keys optional.

martindurant:
I'm happy with this. Any further comments?

chinmaychandak:
Nope, all good from my side. Jarod and I are working on getting the checkpointing PR in soon.

martindurant:
Can you please add the keys parameter to the from-kafka source docstrings?

chinmaychandak:
@martindurant Done!

PR #305, opened 2020-02-28
Added reference counting and Kafka checkpointing

jsmaupin:
This change is to use reference counters to check when data is done. Then, those counters are used by sources to know when to commit offsets. There are two primary types of nodes for the purposes here: synchronous nodes and asynchronous nodes. Synchronous do not exit from the `update` method before calling `_emit`. The reference counting happens in the `Stream` base class in the `_emit` function.  Asynchronous nodes, however do return and will call `_emit` at some time in the future. These nodes are more complicated because the metadata must be managed.

## Synchronous nodes
The only thing changed here is that the metadata is passed forward to the next node.
### Core
* accumulate
* filter
* map
* pluck
* sink
* rate_limit
* slice
* starmap
* union
* unique
* flatten - The only special case for sync' nodes. The reference to the reference counters can not be duplicated for each messages or their counts will be decremented below zero. The metadata is emitted with the last element.

### Dask
* accumulate
* map
* starmap

#### Input to output is 1:1
Couple the metadata with the data on the cache and pull the metadata off the cache with the data and emit them together.
  * buffer
  * delay
  * latest

#### Input single values and output collections
The metadata is cached with the data and emitted a collection.
  * collect
  * partition
  * sliding_window
  * timed_window

#### Multiple stream inputs combined into a single output 
All datum that is emitted downstream will have it's accompanying metadata emitted with it as a collection.
  * combine_latest
  * zip
  * zip_latest
  
#### A yield is performed before _emit
The metadata is a simple pass-through just as in sync nodes, but the reference counter will need to incremented here as there is a yield before the `_emit` call
  * gather
  * scatter

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -57,6 +57,26 @@ def identity(x):
     return x
 
 
+class RefCounter:
+    def __init__(self, initial=0, cb=None, loop=None):
+        self.loop = loop if loop else get_io_loop()
+        self.count = initial
+        self.cb = cb
+
+    def retain(self, x=1):
+        self.count += x
+
+    def release(self, x=1):
+        self.count -= x
+        if self.count <= 0 and self.cb:
+            self.loop.add_callback(self.cb)
+
+    def __str__(self):
+        return '<RefCounter count={}>'.format(self.count)
+
+    __repr__ = __str__
+
+
 class Stream(object):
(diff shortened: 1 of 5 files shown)


jsmaupin:
Some Kafka tests did not pass. I'm going to close this and open a new PR with those tests fixed

PR #306, opened 2020-02-29
Added reference counting and Kafka checkpointing

jsmaupin:
This change is to use reference counters to check when data is done. Then, those counters are used by sources to know when to commit offsets. There are two primary types of nodes for the purposes here: synchronous nodes and asynchronous nodes. Synchronous do not exit from the `update` method before calling `_emit`. Asynchronous nodes, however do return and will call `_emit` at some time in the future. These nodes are more complicated because the metadata must be managed.

## Synchronous nodes
The only thing changed here is that the metadata is passed forward to the next node.
### Core
* accumulate
* filter
* map
* pluck
* sink
* rate_limit
* slice
* starmap
* union
* unique
* flatten - The only special case for sync' nodes. The reference to the reference counters can not be duplicated for each messages or their counts will be decremented below zero. The metadata is emitted with the last element.

### Dask
* accumulate
* map
* starmap

#### Input to output is 1:1
Couple the metadata with the data on the cache and pull the metadata off the cache with the data and emit them together.
  * buffer
  * delay
  * latest

#### Input single values and output collections
The metadata is cached with the data and emitted a collection.
  * collect
  * partition
  * sliding_window
  * timed_window

#### Multiple stream inputs combined into a single output 
All datum that is emitted downstream will have it's accompanying metadata emitted with it as a collection.
  * combine_latest
  * zip
  * zip_latest
  
#### A yield is performed before _emit
The metadata is a simple pass-through just as in sync nodes, but the reference counter will need to incremented here as there is a yield before the `_emit` call
  * gather
  * scatter


CJ-Wright:
Thank you for putting this in! I need some time to read through it

jsmaupin:
Thanks, @CJ-Wright.  I appreciate your time.

streamz/core.py
```diff
@@ -319,18 +339,35 @@ def remove_stream(change):
 
         return output._ipython_display_(**kwargs)
 
-    def _emit(self, x):
+    def _emit(self, x, metadata=None):
```
martindurant: The use and structure of metadata, with particular reference to the reference counter, should be included here and for `emit()`

streamz/core.py
```diff
@@ -319,18 +339,35 @@ def remove_stream(change):
...
+        if metadata is None:
+            metadata = {}
+
+        refs = []
+        if 'refs' in metadata:
```
martindurant: Remove L343/344 in favour of `refs = metadata.get('refs', []) if metadata else []`

streamz/core.py
```diff
@@ -319,18 +339,35 @@ def remove_stream(change):
...
+            refs = metadata['refs']
+            if not isinstance(refs, list):
```
martindurant: Explain in a comment, when is refs a list versus a single instance?

streamz/core.py
```diff
@@ -319,18 +339,35 @@ def remove_stream(change):
...
+                refs = [refs]
+
         result = []
-        for downstream in list(self.downstreams):
-            r = downstream.update(x, who=self)
+        downstreams = list(self.downstreams)
```
martindurant: This line is not necessary, just refer to `self.downstreams` in the lines below. `len()` and iteration will work regardless.

streamz/core.py
```diff
@@ -319,18 +339,35 @@ def remove_stream(change):
...
+        for downstream in downstreams:
+            r = downstream.update(x, who=self, metadata=metadata)
+
             if type(r) is list:
                 result.extend(r)
             else:
                 result.append(r)
 
+            for ref in refs:
+                ref.release()
```
martindurant: So even in completely sync operations, the references value goes up while the chain of tasks is executed, and then down again. This is unnecessary computing, but I suppose insignificant?

streamz/core.py
```diff
@@ -497,6 +534,28 @@ def to_batch(self, **kwargs):
         from .batch import Batch
         return Batch(stream=self, **kwargs)
 
+    def _retain_refs(self, metadata):
+        if not isinstance(metadata, list):
```
martindurant: Again, no docstrings here or even comments; and more `isinstance` calls than I am comfortable with. `metadata` is a dict or a list of dict?

streamz/core.py
```diff
@@ -497,6 +534,28 @@ def to_batch(self, **kwargs):
...
+            metadata = [metadata]
+        for m in metadata:
+            if m and 'refs' in m:
```
martindurant: Is `m` ever `None`? 
How about `refs = m.get('refs', []`)`
Also, no need to create the list every time
```
if isinstance(refs, list):
    for ref in refs:
        ref.retain(1)
else:
    refs.retain(1)
```
Note that the value of 1 is already the default value.

streamz/sources.py
```diff
@@ -453,12 +453,13 @@ def _close_consumer(self):
...
     """Base class for both local and cluster-based batched kafka processing"""
     def __init__(self, topic, consumer_params, poll_interval='1s',
-                 npartitions=1, **kwargs):
+                 npartitions=1, keys=False, **kwargs):
         self.consumer_params = consumer_params
         self.topic = topic
         self.npartitions = npartitions
         self.positions = [0] * npartitions
         self.poll_interval = convert_interval(poll_interval)
+        self.keys = keys
```
martindurant: This is already in master?

streamz/sources.py
```diff
@@ -467,6 +468,30 @@ def __init__(self, topic, consumer_params, poll_interval='1s',
     def poll_kafka(self):
         import confluent_kafka as ck
 
+        def commit(_part):
```
martindurant: Can these functions be moved outside this def? Dynamic function closures have a reputation for holding references.

streamz/sources.py
```diff
@@ -467,6 +468,30 @@ def __init__(self, topic, consumer_params, poll_interval='1s',
...
+
+        tps = []
+        for partition in range(self.npartitions):
+            tps.append(ck.TopicPartition(self.topic, partition))
+
+        while True:
+            try:
+                committed = self.consumer.committed(tps, timeout=1)
+            except ck.KafkaException:
+                pass
```
martindurant: Does an exception here cause rapid-fire attempts to commit? This might seize the whole program

streamz/sources.py
```diff
@@ -481,11 +506,11 @@ def poll_kafka(self):
                     lowest = max(current_position, low)
                     if high > lowest:
                         out.append((self.consumer_params, self.topic, partition,
-                                    lowest, high - 1))
+                                    self.keys, lowest, high - 1))
                         self.positions[partition] = high
 
                 for part in out:
-                    yield self._emit(part)
+                    self.loop.add_callback(checkpoint_emit, part)
```
martindurant: why not `yield` here?

streamz/sources.py
martindurant: **This** is the line that is causing coverage to go down...

streamz/core.py
```diff
@@ -57,6 +57,26 @@ def identity(x):
     return x
 
 
+class RefCounter:
+    def __init__(self, initial=0, cb=None, loop=None):
+        self.loop = loop if loop else get_io_loop()
+        self.count = initial
+        self.cb = cb
+
+    def retain(self, x=1):
```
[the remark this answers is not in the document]
jsmaupin: I used the Objective-C convention as Obj-C uses the reference counting for memory management : https://www.roseindia.net/iphone/objectivec/memory-management.shtml

I'm fine changing it if you feel strongly about this.

streamz/core.py
jsmaupin: We are talking about documentation?

martindurant: Yes please

streamz/core.py
jsmaupin: While I agree with this, I'll need to figure out a way to reduce these.

jsmaupin:
Thanks, a lot @martindurant. This is great.  I'll go through each of these and commit updates.

streamz/sources.py
jsmaupin: I'll have to re-import confluent_kafka every time the function is called

streamz/sources.py
jsmaupin: This could cause a rapid fire requests of already committed offsets. I suppose an alternative is to either try once or a specified number of times. If this fails n times, then we could honor the `auto.offset.reset` configuration in the `consumer_params` arugment. 

What do you think?

streamz/core.py
jsmaupin: I believe this is necessary because the source or creator of the ref counter cannot make assumption about what the pipeline looks like. The source knows that the data is done due to the value of the reference counter changing from greater than zero to zero. If we do not use counters for the sync operations, then the counter will never trigger the done callback.

streamz/core.py
```diff
@@ -57,6 +57,26 @@ def identity(x):
     return x
 
 
+class RefCounter:
+    def __init__(self, initial=0, cb=None, loop=None):
```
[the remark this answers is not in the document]
jsmaupin: I have added the doc strings.

streamz/core.py
jsmaupin: done.

streamz/core.py
jsmaupin: This has been cleaned up. In the code we never have a list of refs, only a list of metadata.

streamz/core.py
jsmaupin: Cleaned up. Refs is always a single instance

streamz/core.py
jsmaupin: This code has been removed and simplified.

streamz/sources.py
jsmaupin: I removed this and it will be merged in when pulled to master.

streamz/sources.py
jsmaupin: I have added a  `yield` here.

jsmaupin:
I have addressed all the issues above including adding a new collection of tests for reference counters in all the async'/cache nodes.

streamz/core.py
jsmaupin: fixed.

streamz/core.py
jsmaupin: I have added docstrings here and simplifed the code by assuming that metadata is always a list

streamz/sources.py
chinmaychandak: What about changes like this one that would be needed: [checkpointing + keys](https://github.com/chinmaychandak/streamz/blob/a951ffa91e626a8ce2a00fee16ac7309a5a693ce/streamz/sources.py#L473).

martindurant:
The one outstanding question is of exceptions (i.e., if the callback fails, such as updating kafka offsets). Streamz doesn't currently do anything useful with exceptions anyway, so this can be a future PR: maybe the metadata object collects exceptions and reports them to the callback when the count reaches zero.

jsmaupin:
From what I've seen in other frameworks, when the fail to commit the offset for whatever reason, all they do is log a message and keep processing. 

However, I get your point,. There could be a more critical exception elsewhere. For anything that would result in a loss of data, I would prefer to just stop the pipeline with an error message.

streamz/core.py
```diff
@@ -1131,19 +1272,26 @@ def _remove_upstream(self, upstream):
...
         super(combine_latest, self)._remove_upstream(upstream)
         if self._initial_emit_on is None:
             self.emit_on = self.upstreams
 
-    def update(self, x, who=None):
+    def update(self, x, who=None, metadata=None):
+        self._retain_refs(metadata)
+        idx = self.upstreams.index(who)
+        if self.metadata[idx]:
+            self._release_refs(self.metadata[idx])
```
CJ-Wright: Is this releasing the reference too early?
It seems that you would need the reference until the tuple is emitted (line 1294)

chinmaychandak: I think we would need to release the reference when we see a new element in the stream that would now be the 'latest' for that stream. So this the current code is correct?

CJ-Wright: Ah ok, sorry I misread that. I would move this down so that it is matched up with `self.last[idx] = x` which is doing the same thing but for the data. Or maybe this needs a comment saying that we're bouncing an element out from that position (if there is an element at that position) and therefore need to release the reference count

jsmaupin: I do not think this is too early. Here `self.metadata` holds the metadata for the previous call to `update`. When a new value comes in, that new value is emitted and not the value that is in the `self.metadata` buffer. The new data replaces the value in the buffer and the old value is "released" as it is no longer being "retained" by the `combine_latest` node.

jsmaupin:
Hey, @CJ-Wright, sure. I'm happy to add clarity where it would be of benefit to others.  I just want to be clear on what you're suggesting.  Are you talking about comments, docstrings, or something else?

CJ-Wright:
Comments would be good, maybe also a short docs page on metadata handling (how is metadata propagated, how is it used, when implementing a new node how do you need to think about it)

For instance we have prose docs around [async usage](https://streamz.readthedocs.io/en/latest/async.html) maybe we need a developer's notes page.

jsmaupin:
I added some documentation. It didn't seem to make sense to add documentation for each and every function as the docs are for the user and should be based on the API. I added sections in the core doc for both metadata and reference counting.

jsmaupin:
Let me know if you think the documentation not enough, or if I need to rebase so merging is easier.

CJ-Wright:
I like the documentation, thank you!

martindurant:
All good, then, @CJ-Wright ? Please press the big green button if so.
We should probably release after this change.

CJ-Wright:
Thank you everyone who contributed to this and all its iterations! This is a really awesome feature!

PR #310, opened 2020-03-12
Adding Kafka message keys PR overwritten by Checkpointing + max_batch_size parameter for from_kafka_batched

chinmaychandak:
Adding #304 back after being overwritten by #306.

This PR also adds `max_batch_size` as a parameter to `from_kafka_batched` — maximum number of messages per partition to be read per batch from Kafka. Currently, if there's a million messages in topic partition, streamz attempts to read all of them in one go, which can be a problem and overload/overflow memory.

### DIFF
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -453,12 +453,13 @@ class from_kafka(Source):
 class FromKafkaBatched(Stream):
     """Base class for both local and cluster-based batched kafka processing"""
     def __init__(self, topic, consumer_params, poll_interval='1s',
-                 npartitions=1, **kwargs):
+                 npartitions=1, keys=False, **kwargs):
         self.consumer_params = consumer_params
         self.topic = topic
         self.npartitions = npartitions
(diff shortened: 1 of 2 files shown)


martindurant:
Ooops :)

chinmaychandak:
Adding another small PR to this one, so as to not have only redundant code from the previous keys PR. :)

streamz/sources.py
```diff
@@ -503,9 +505,11 @@ def checkpoint_emit(_part):
                         continue
                     current_position = self.positions[partition]
                     lowest = max(current_position, low)
+                    if high > lowest + self.max_batch_size:
+                        high = lowest + self.max_batch_size
```
martindurant: This is not tested, right?

chinmaychandak: Correct, I am just writing up a test right now.

chinmaychandak: Done

martindurant:
Thanks!

PR #311, opened 2020-03-12
Add parameter to allow reading from latest offsets in Kafka.

chinmaychandak:
Currently, a stream starts reading from the earliest message available in the topic. Sometimes, if a topic has a million messages, or if a stream restarts, you might not want to read from the earliest record. This would allow the stream to start reading from the 'latest' offset in each partition.

Would love to have ideas here, if there's a more efficient implementation possible for this.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -95,6 +95,7 @@ Definitions
 
 .. autofunction:: filenames
 .. autofunction:: from_kafka
+.. autofunction:: from_kafka_batched
 .. autofunction:: from_textfile
 
 .. currentmodule:: streamz.dask
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -457,7 +457,7 @@ class FromKafkaBatched(Stream):
                  keys=False, **kwargs):
         self.consumer_params = consumer_params
         # Override the auto-commit config to enforce custom streamz checkpointing
-        self.consumer_params["enable.auto.commit"] = False
+        self.consumer_params['enable.auto.commit'] = False
         self.topic = topic
         self.npartitions = npartitions
         self.positions = [0] * npartitions
@@ -507,8 +507,11 @@ class FromKafkaBatched(Stream):
                             tp, timeout=0.1)
                     except (RuntimeError, ck.KafkaException):
                         continue
-                    if self.latest is True or self.consumer_params["auto.offset.reset"] == "latest":
+                    if self.latest is True:
                         self.positions[partition] = high
+                    if 'auto.offset.reset' in self.consumer_params.keys():
+                        if self.consumer_params['auto.offset.reset'] == "latest":
+                            self.positions[partition] = high
                     current_position = self.positions[partition]
                     lowest = max(current_position, low)
                     if high > lowest + self.max_batch_size:
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -453,17 +453,15 @@ class from_kafka(Source):
 class FromKafkaBatched(Stream):
     """Base class for both local and cluster-based batched kafka processing"""
     def __init__(self, topic, consumer_params, poll_interval='1s',
-                 npartitions=1, max_batch_size=10000, latest=False,
-                 keys=False, **kwargs):
+                 npartitions=1, max_batch_size=10000, keys=False, **kwargs):
         self.consumer_params = consumer_params
         # Override the auto-commit config to enforce custom streamz checkpointing
-        self.consumer_params['enable.auto.commit'] = False
+        self.consumer_params['enable.auto.commit'] = 'false'
         self.topic = topic
         self.npartitions = npartitions
         self.positions = [0] * npartitions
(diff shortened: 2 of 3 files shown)


jsmaupin:
This parameter already exists in the Kafka config.  Would it make sense to honor that parameter in the config provided by the client?

jsmaupin:
Also, this might apply to the `enable.auto.offset.store` config as well.

chinmaychandak:
Do you think the behavior of `from_kafka_batched` forming `partition, start_offset, end_offset` tuples to be read from Kafka would change if the `auto.offset.reset: True` is set in the consumer config? It would still pull the low, high watermarks and then compare with committed offsets and form tuples.

I will test this out, though. If it works, you're right. We can just honor the config values instead of having a new parameter.

jsmaupin:
Firstly, I think it would be confusing and misleading to be able to define these parameters in two places. How does the user know which one will be used and which will be ignored? 

> Do you think the behavior of from_kafka_batched forming partition, start_offset, end_offset tuples to be read from Kafka would change if the auto.offset.reset: True is set in the consumer config? It would still pull the low, high watermarks and then compare with committed offsets and form tuples.

I think that, because the current implementation in Streamz specifically tells the Confluent lib where to read instead of allowing the Confluent lib to decide, this parameter is meaningless to the Confluent lib.  It will only have meaning Streamz implementation on top.

However, for the `enable.auto.offset.store`, the Confluent lib may try to save the offsets in addition to the Streamz implementation.  Maybe this config can be in another PR, but we may have to force it to false in the constructor in order to prevent the Confluent lib from deciding when to push our checkpoints.

chinmaychandak:
> How does the user know which one will be used and which will be ignored?

We can have an either-or check, then? Most Python developers aren't aware of specific Kafka  configs like this one, I think. So, for important configs like this, we might want to consider having an additional parameter.

> this parameter is meaningless to the Confluent lib. It will only have meaning Streamz implementation on top.

So, we need something like this?
```
if latest is True or configs['auto.offset.reset'] is True:
    self.positions[partition] = high
``` 

> Maybe this config can be in another PR, but we may have to force it to false in the constructor in order to prevent the Confluent lib from deciding when to push our checkpoints.

So, `enable.auto.offset.store` store offsets to auto-commit. I agree that we need to force `enable.auto.commit`  to false to ensure that streamz' checkpointing mechanism gets used instead.

@martindurant, @CJ-Wright Any thoughts?

martindurant:
I don't know much about the kafka config, but I agree that it can be useful to explicitly surface important ones, so long as we are explicit about how to merge them with those contained in the config dict.

chinmaychandak:
I've added code to enforce `"enable.auto.commit": False` to ensure streamz' checkpointing is followed. 

I've also added the condition that either setting `latest = True` in the `from_kafka_batched` parameters or setting `"auto.offset.reset": "latest"` in the Kafka consumer config will work if one wants to start reading from the latest offsets in the stream.

martindurant:
@jsmaupin , more thoughts on how things stand here?

jsmaupin:
Mostly, I think this is fine, but I'm not sure I see the need for another parameter in the growing list of parameters in `from_kafka_batched`. The parameter name in the Kafka consumer parameters seems less ambiguous about the underlying behavior.

martindurant:
@jsmaupin , you would prefer good doc instead?

jsmaupin:
Yes, that would be good, I think.

chinmaychandak:
Spark Streaming exposes _options_ for `maxOffsetsPerTrigger` and `startingOffsets` since these are two of the most importantly used streaming configs, that’s why I thought maybe it made sense to expose both `max_batch_size` and `latest` parameters directly in the streamz API, too.

chinmaychandak:
> Spark Streaming exposes _options_ for `maxOffsetsPerTrigger` and `startingOffsets` since these are two of the most important streaming configs, that’s why I thought maybe it made sense to expose both `max_batch_size` and `latest` parameters directly in the streamz API, too.

If you guys still feel strongly otherwise, I can just ensure that `auto.offset.reset` from the config gets used for this functionality. 

> you would prefer good doc instead?

@martindurant did you mean doc string, or something else?

martindurant:
Yes, docstrings at the minimum. Obviously, where this is described in the RTD docs, links and description to the various options would be good.

chinmaychandak:
> Yes, docstrings at the minimum. Obviously, where this is described in the RTD docs, links and description to the various options would be good.

I don't see a `sources.rst` anywhere, so I don't think one exists as of now. I already have a doc string update in this PR, although that would change a little now that we don't use a new parameter for this functionality. Would that be okay?

chinmaychandak:
@martindurant, @jsmaupin Done!

jsmaupin:
> Use only auto.offset.reset for latest functionality; add doc links

Excellent!

martindurant:
Thanks for this.

PR #312, opened 2020-03-13
Union of Kafka streams should not freeze when either of the sources freezes/fails.

### LINKED ISSUE #308: Multi-kafka for high availability
We are setting up Kafka in multiple data centers to achieve high availability. I expected that Streamz could do a union between both of these clusters.  Everything works as expected, but when I stop one of the Kafka clusters, the stream freezes. I'm thinking there is a callback somewhere that is blocking operations?

```python
import asyncio

import confluent_kafka as ck
from streamz import from_kafka_batched


async def produce(broker, cluster_id, topic):
    p = ck.Producer({'bootstrap.servers': broker})

    for i in range(100000):
        p.poll(0)
        p.produce(topic, 'cluster-{} value-{}'.format(cluster_id, i))
        p.flush(timeout=0.5)
        await asyncio.sleep(1)


async def main():
    topic = 'test-topic'
    args1 = {'bootstrap.servers': 'localhost:9092', 'group.id': 'test-group'}
    args2 = {'bootstrap.servers': 'localhost:9093', 'group.id': 'test-group'}
    a = from_kafka_batched(topic, args1, npartitions=1, start=True,
                           asynchronous=True)
    b = from_kafka_batched(topic, args2, npartitions=1, start=True,
                           asynchronous=True)

    a.union(b).sink(print)

    asyncio.create_task(produce('localhost:9092', 1, topic))
    asyncio.create_task(produce('localhost:9093', 2, topic))

    while True:
        await asyncio.sleep(1)


if __name__ == '__main__':
    asyncio.run(main())
```

chinmaychandak:
PR tries to resolve #308. 

In the event when CK encounters an exception at `get_watermark_offsets`, we typically need to recreate (resubscribe to the topic) the consumer in case we want to consume the next time around when Kafka is up again. I'm thinking there might be a cleaner way than this, though.

Below is the example notebook I ran to confirm if this works. I have also attached the output of each cell. 

```
from streamz import Stream
import random
import confluent_kafka as ck

PRODUCER_ARGS = {
    "bootstrap.servers": "localhost:9092"
}
producer = ck.Producer(PRODUCER_ARGS)
j1 = random.randint(0, 10000)
CONSUMER_ARGS1 = {
    "bootstrap.servers": "localhost:9092",
    "group.id": 'streamz-test%i' % j1,
    "enable.auto.commit": "false",
    "enable.partition.eof": "true",
    "session.timeout.ms": "60000",
    "auto.offset.reset": "earliest"
}
TOPIC1 = "prod"
TOPIC2 = "backup"

a = Stream.from_kafka_batched(TOPIC1, CONSUMER_ARGS1)
a.start()

b = Stream.from_kafka_batched(TOPIC2, CONSUMER_ARGS1)
b.start()

a.union(b).sink(print)
```
```
producer.produce(TOPIC1, '{"name":"Prod", "amount":50}', "")
producer.produce(TOPIC2, '{"name":"Backup", "amount":100}', "")
```
[b'{"name":"Backup", "amount":100}']
[b'{"name":"Prod", "amount":50}']

### After Deleting prod topic ###

```
producer.produce(TOPIC2, '{"name":"Backup", "amount":150}', "")
producer.produce(TOPIC2, '{"name":"Backup", "amount":200}', "")
```
[b'{"name":"Backup", "amount":150}', b'{"name":"Backup", "amount":200}']

### After Recreating prod topic ###

```
producer.produce(TOPIC1, '{"name":"Prod", "amount":50}', "")
producer.produce(TOPIC1, '{"name":"Prod", "amount":250}', "")
producer.produce(TOPIC1, '{"name":"Prod", "amount":300}', "")
producer.produce(TOPIC2, '{"name":"Backup", "amount":350}', "")
```

[b'{"name":"Prod", "amount":250}', b'{"name":"Prod", "amount":300}']
[b'{"name":"Backup", "amount":350}']

Should we add a test for this?

### DIFF
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -504,6 +504,12 @@ class FromKafkaBatched(Stream):
                         low, high = self.consumer.get_watermark_offsets(
                             tp, timeout=0.1)
                     except (RuntimeError, ck.KafkaException):
+                        print("Recreating consumer")
+                        self.consumer.unsubscribe()
+                        self.consumer.close()
+                        self.consumer = ck.Consumer(self.consumer_params)
+                        self.consumer.subscribe([self.topic])
+                        time.sleep(5)
                         continue
                     if self.latest is True:
                         self.positions[partition] = high
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -504,10 +504,6 @@ class FromKafkaBatched(Stream):
                         low, high = self.consumer.get_watermark_offsets(
                             tp, timeout=0.1)
                     except (RuntimeError, ck.KafkaException):
-                        self.consumer.unsubscribe()
-                        self.consumer.close()
-                        self.consumer = ck.Consumer(self.consumer_params)
-                        self.consumer.subscribe([self.topic])
                         continue
                     if self.latest is True:
                         self.positions[partition] = high
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -504,6 +504,10 @@ class FromKafkaBatched(Stream):
                         low, high = self.consumer.get_watermark_offsets(
                             tp, timeout=0.1)
                     except (RuntimeError, ck.KafkaException):
+                        self.consumer.unsubscribe()
+                        self.consumer.close()
+                        self.consumer = ck.Consumer(self.consumer_params)
+                        self.consumer.subscribe([self.topic])
                         continue
                     if self.latest is True:
                         self.positions[partition] = high
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -453,13 +453,15 @@ class from_kafka(Source):
 class FromKafkaBatched(Stream):
     """Base class for both local and cluster-based batched kafka processing"""
     def __init__(self, topic, consumer_params, poll_interval='1s',
-                 npartitions=1, max_batch_size=10000, keys=False, **kwargs):
+                 npartitions=1, max_batch_size=10000, latest=False,
+                 keys=False, **kwargs):
         self.consumer_params = consumer_params
         self.topic = topic
         self.npartitions = npartitions
         self.positions = [0] * npartitions
         self.poll_interval = convert_interval(poll_interval)
         self.max_batch_size = max_batch_size
+        self.latest = latest
         self.keys = keys
         self.stopped = True
 
@@ -503,6 +505,8 @@ class FromKafkaBatched(Stream):
                             tp, timeout=0.1)
                     except (RuntimeError, ck.KafkaException):
                         continue
+                    if self.latest is True:
+                        self.positions[partition] = high
                     current_position = self.positions[partition]
                     lowest = max(current_position, low)
                     if high > lowest + self.max_batch_size:
@@ -511,6 +515,7 @@ class FromKafkaBatched(Stream):
                         out.append((self.consumer_params, self.topic, partition,
                                     self.keys, lowest, high - 1))
                         self.positions[partition] = high
+                self.latest = False
 
                 for part in out:
                     yield self.loop.add_callback(checkpoint_emit, part)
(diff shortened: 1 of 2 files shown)


streamz/sources.py
```diff
@@ -502,6 +502,10 @@ def checkpoint_emit(_part):
                         low, high = self.consumer.get_watermark_offsets(
                             tp, timeout=0.1)
                     except (RuntimeError, ck.KafkaException):
+                        self.consumer.unsubscribe()
```
martindurant: Perhaps you'll need to patch the consumer for testing, so that it throws a RuntimeError at some point

chinmaychandak: I can't seem to think of a way to do that (mocking the consumer) which wouldn't affect the semantics of `from_kafka_batched`. Do you have any ideas?

Is there way we can actually delete (this would throw the RuntimeError like the notebook I've attached here) and recreate the docker container's Kafka topic in the test?

martindurant: Really taking down kafka would be a more realistic test, but also rather heavy. I'm thinking like
```
def make_err(*_, **_):
    raise RuntimeError

... # values get through
prev = source.consumer. get_watermark_offsets
source.consumer. get_watermark_offsets = make_err
... # no values from the one source, continues from other, no error
source.consumer. get_watermark_offsets = prev
... # values from both
```

jsmaupin:
From my test, the Confluent lib did a pretty good job of reconnecting when the cluster came back.  I wonder if everything will work if we add a `yield gen.sleep(1)` where there's a `pass` to yield control back to the event loop.

martindurant:
The `yield` sounds right, perhaps with exponential backoff

streamz/sources.py
chinmaychandak: Okay, this test's working well! :)

```
def test_multi_kafka_failover():
    CONSUMER_ARGS = {
        "bootstrap.servers": "localhost:9092",
        "group.id": 'streamz-test',
        "enable.auto.commit": "false",
        "enable.partition.eof": "true",
        "session.timeout.ms": "60000",
        "auto.offset.reset": "earliest"
    }
    TOPIC1 = "prod"
    TOPIC2 = "backup"

        kafka, TOPIC = kafka

        a = Stream.from_kafka_batched(TOPIC1, CONSUMER_ARGS)
        a.start()

        b = Stream.from_kafka_batched(TOPIC2, CONSUMER_ARGS)
        b.start()

        out = a.union(b).sink_to_list()

        kafka.produce(TOPIC1, '50')
        kafka.produce(TOPIC2, '100')

        wait_for(lambda: any(out) and len(out) == 2, 10, period=0.2)

        # Delete prod topic
        os.system("kafka-topics.sh --zookeeper 127.0.0.1:2181 --delete --topic prod")
        time.sleep(5)

        kafka.produce(TOPIC2, '150')
        kafka.produce(TOPIC2, '200')

        wait_for(lambda: any(out) and out[-1] == [b'150', b'200'], 10, period=0.2)

        # Recreate prod topic
        os.system("kafka-topics.sh --zookeeper 127.0.0.1:2181 --create" 
                  "--partitions 1 --replication-factor 2 --topic prod")
        time.sleep(5)

        kafka.produce(TOPIC1, '50')
        kafka.produce(TOPIC1, '250')
        kafka.produce(TOPIC1, '300')
        kafka.produce(TOPIC2, '350')

        wait_for(lambda: any(out) and len(out) == 5, 10, period=0.2)
        assert out[-2] == [b'50', b'250', b'300'] and out[-1] == [b'350']
```

jsmaupin:
I think I see what the issue here is.  I set up a test with two local Kafka clusters.  However, I can demonstrate this with just one.  It seems that `get_watermark_offsets` does not respect the timeout when Kafka is not available. So, the test is:

1. Start Kafka
2. Start the script below
3. Kill Kafka with `ctrl+c`
4. Notice the output

```python
from time import sleep

import confluent_kafka as ck


def main():
    params = {'bootstrap.servers': 'localhost:9092',
              'group.id': 'test-group',
              'enable.auto.offset.store': False}

    c = ck.Consumer(params)

    while True:
        tp = ck.TopicPartition('test-topic', 0)
        print(c.get_watermark_offsets(tp, timeout=1))
        print('tick')
        sleep(1)


if __name__ == '__main__':
    main()
```

jsmaupin:
I see there's an issue for this already in the confluent repo: https://github.com/confluentinc/confluent-kafka-python/issues/413

Also this: https://github.com/confluentinc/confluent-kafka-python/issues/412

PR #316, opened 2020-03-20
Fixed metadata forwarding and added associated tests

jsmaupin:
In order for checkpointing to work in a pipeline, the metadata must be in a list and not a nested list. If the metadata is forwarded as a nested list, then managing it can get very complicated. Previously there were tests to verify that the reference counters were being incremented and decremented properly. The tests in this PR ensure that the metadata is forwarded properly for all functions that emit collections.


jsmaupin:
It looks like there's an intermittent failure that's unrelated to the changes?

streamz/tests/test_core.py
```diff
@@ -199,6 +210,21 @@ def test_sliding_window_ref_counts():
...
+
+    L = []
+    sink = sink_metadata_to_list(stream, L)
+    stream.connect(sink)
+
+    source.emit(0)
+    source.emit(1, metadata=[{'v': 1}])
+    source.emit(2, metadata=[{'v': 2}])
+    source.emit(3, metadata=[{'v': 3}])
+    assert L == [[{'v': 1}], [{'v': 1}, {'v': 2}], [{'v': 2}, {'v': 3}]]
```
martindurant: I am confused by why this looks like it does. I would have expected
`[{'v': 1}, {'v': 2}, {'v': 3}]`

streamz/tests/test_core.py
```diff
@@ -199,6 +210,21 @@ def test_sliding_window_ref_counts():
         r_prev = r
 
 
+def test_sliding_window_metadata():
+    source = Stream()
+    stream = source.sliding_window(2)
+
+    L = []
+    sink = sink_metadata_to_list(stream, L)
+    stream.connect(sink)
```
martindurant: Ordinary `sink_to_list` doesn't need to be primed with a list, or have the sink object assigned and connected. This should probably do the same (I realise it's only for testing).

streamz/tests/test_core.py
jsmaupin: I do not believe that `[{'v': 1}, {'v': 2}, {'v': 3}]` is correct. Note that the above is a list with three lists.  This is how I see the result:

```python
    assert L == \
           [
               [{'v': 1}],            # First emit, because 0 has no metadata 
               [{'v': 1}, {'v': 2}],  # Second emit
               [{'v': 2}, {'v': 3}]   # Third emit
           ]
```

martindurant: Sorry, totally missed the sliding window, you must be right

jsmaupin: Maybe I can comment on all the tests like above.

jsmaupin: done.

streamz/utils_test.py
```diff
@@ -125,3 +125,9 @@ def await_for(predicate, timeout, fail_func=None, period=0.001):
             if fail_func is not None:
                 fail_func()
             pytest.fail("condition not reached until %s seconds" % (timeout,))
+
+
+class metadata(Stream):
```
martindurant: I like this - could even be useful as a user class perhaps

jsmaupin: Yeah, I agree. Should that be in this PR or in another?

martindurant: Don't bother for now - in case someone asks :)

martindurant:
I'm good with this, will merge over the weekend unless there are other comments.

Issue #317, opened 2020-03-27
Unable to use buffer after partition


roveo:
# Demonstration

```py
from streamz import Stream

source = Stream()
source.partition(3).buffer(10).sink(print)

for i in range(20):
    source.emit(i)
```

```
Traceback (most recent call last):
  File "test.py", line 4, in <module>
    source.partition(3).buffer(10).sink(print)
TypeError: 'list' object is not callable
```

# Why

`partition` overwrites `buffer` in `__init__`:
```py
    def __init__(self, upstream, n, **kwargs):
        self.n = n
        self.buffer = []
        Stream.__init__(self, upstream, **kwargs)
```

# Solution

Rename `partition.buffer` into something else.

martindurant:
I am not sure I understand why the attribute name is a problem
*but* - your code here is, essentially, a test case. I would recommend you add it to the tests, and show that your proposed change makes it pass, and submit a new PR.

roveo:
Will do.

> I am not sure I understand why the attribute name is a problem

In core.py buffer is registered as a method of Stream class. But when `partition` stream is constructed, `self.buffer` get overwritten. So you can't use buffer right after partition (and also timed_window and several other stream kinds that also overwrite buffer).

martindurant:
aha, understood. Perhaps indeed rename all those instance attributes to `_buffer`.

PR #318, opened 2020-03-27
fix some streams overwriting Stream.buffer method

roveo:
closes https://github.com/python-streamz/streamz/issues/317


martindurant:
Thank you!

PR #328, opened 2020-04-13
pass metadata when using LRU too, fixes #327

### LINKED ISSUE #327: unique not passing metadata downstream
Here it emits the message without incoming metadata (line 1412):

https://github.com/python-streamz/streamz/blob/994b7308387ae6f9113adc55dbcc9cf558d85327/streamz/core.py#L1397-L1412

I'd make a PR but I don't know how to test for this because none of the built-in core operations have access to metadata.

I only noticed it because I use custom versions of `.map` and `.sink` that receive metadata instead of args and kwargs passed at build time.

roveo:


roveo:
#327

streamz/tests/test_core.py
```diff
@@ -699,6 +699,20 @@ def test_unique_key():
     assert L == [1, 2, 3]
 
 
+def test_unique_metadata():
+    @Stream.register_api()
```
jsmaupin, line 703: There is already a `metdata()` function used for testing.  Can we reuse that? See the test `test_sliding_window_metadata()`.

```python
    source = Stream()
    L = metadata(source.sliding_window(2)).sink_to_list()
```

streamz/tests/test_core.py
```diff
@@ -699,6 +699,20 @@ def test_unique_key():
...
+    class get_metadata(Stream):
+        def update(self, x, metadata=None, who=None):
+            self._emit(metadata)
+
+    source = Stream()
+    L = source.unique().get_metadata().flatten().sink_to_list()
```
jsmaupin, line 709: Is the metadata not already flat? All functions should be forwarding a flat list.

roveo:
```
>>> from streamz import Stream
>>> from streamz.utils_test import metadata
>>> src = Stream()
>>> L = metadata(src).sink_to_list()
>>> src.emit(0, metadata=[{'a': 1}])
>>> src.emit(1, metadata=[{'b': 2}])
>>> L
[[{'a': 1}], [{'b': 2}]]
```
As far as I understood [the docs](https://github.com/python-streamz/streamz/blob/master/docs/source/core.rst#metadata), metadata should always be a list. In this situation, the result of `metadata(...)` is a list of lists.

streamz/tests/test_core.py
roveo: Will do.

streamz/tests/test_core.py
jsmaupin: Oh yes, I see. There is one element on the output for each emit.

roveo:
Looks like `test_from_kafka` failed, on the surface nothing to do with my commit.

jsmaupin:
> Looks like `test_from_kafka` failed, on the surface nothing to do with my commit.

Yeah, I had the same issue on my commit. This test is flaky.

roveo:
> This test is flaky.

What do I do? Any way to rerun the build? (except for making a dummy commit)

jsmaupin:
I'm not sure.  I think @CJ-Wright is the person to make the call on that.

martindurant:
Restarted the test build

PR #330, opened 2020-04-21
Add Accelerated Kafka datasource.

chinmaychandak:
This PR adds an `engine` parameter to `from_kafka_batched`. If `engine="cudf"`, messages (for now, only JSON is supported) will be read from Kafka directly into a cuDF dataframe. 

`custreamz.kafka` has the exact same API as Confluent Kafka, so it serves as a drop-in replacement in `from_kafka_batched` with minimal duplication of code. But under the hood, it reads messages from librdkafka and directly uploads them to the GPU as a cuDF dataframe instead of gathering all the messages back from C++ into Python.

This essentially avoids the GIL issue we encountered in the Confluent Kafka consumer, and hence enables reading from Kafka in a faster fashion with fewer processes. This accelerated reader also adheres to the current checkpointing mechanism.

Folks interested in trying out custreamz would benefit from this accelerated Kafka reader. If someone does not want to use GPUs, they can use streamz as is, with the default `engine=None`.

I am skipping the tests for this PR, since we would be having gpuCI tests in cuDF/cuStreamz once things consolidate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ language: python
 
 matrix:
   include:
-    - python: 3.6
+    - python: 3.7
 
 env:
   - STREAMZ_LAUNCH_KAFKA=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ language: python
 
 matrix:
   include:
-    - python: 3.7
+    - python: 3.6
 
 env:
   - STREAMZ_LAUNCH_KAFKA=true
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 Streamz
 =======
 
-|Build Status| |Doc Status| |Version Status|
+|Build Status| |Doc Status| |Version Status| |RAPIDS custreamz gpuCI|
 
 Streamz helps you build pipelines to manage continuous streams of data. It is simple to use in simple cases, but also supports complex pipelines that involve branching, joining, flow control, feedback, back pressure, and so on.
 
@@ -21,3 +21,5 @@ BSD-3 Clause
    :alt: Documentation Status
 .. |Version Status| image:: https://img.shields.io/pypi/v/streamz.svg
    :target: https://pypi.python.org/pypi/streamz/
+.. |RAPIDS custreamz gpuCI| image:: https://img.shields.io/badge/gpuCI-custreamz-green
+   :target: https://github.com/jdye64/cudf/blob/kratos/python/custreamz/custreamz/kafka.py
--- a/conda/environments/streamz_dev.yml
+++ b/conda/environments/streamz_dev.yml
@@ -6,6 +6,7 @@ channels:
 dependencies:
   - python>=3.7
   - pytest
+  - pytest-tornasync
   - flake8
   - black
   - isort
--- a/conda/environments/streamz_dev.yml
+++ b/conda/environments/streamz_dev.yml
@@ -7,7 +7,6 @@ dependencies:
   - python>=3.7
   - pytest
   - pytest-tornasync
-  - pytest-async
   - flake8
   - black
   - isort
--- a/conda/environments/streamz_dev.yml
+++ b/conda/environments/streamz_dev.yml
@@ -6,7 +6,6 @@ channels:
 dependencies:
   - python>=3.7
   - pytest
-  - pytest-tornasync
   - flake8
   - black
   - isort
--- a/conda/environments/streamz_dev.yml
+++ b/conda/environments/streamz_dev.yml
@@ -7,6 +7,7 @@ dependencies:
   - python>=3.7
   - pytest
   - pytest-tornasync
+  - pytest-async
   - flake8
   - black
   - isort
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -498,6 +498,7 @@ class FromKafkaBatched(Stream):
                 for tp in committed:
                     self.positions[tp.partition] = tp.offset
                 break
+
         try:
             while not self.stopped:
                 out = []
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -586,6 +586,8 @@ def from_kafka_batched(topic, consumer_params, poll_interval='1s',
         Number of partitions in the topic
     start: bool (False)
         Whether to start polling upon instantiation
+    max_batch_size: int
+        The maximum number of messages per partition to be consumed per batch
     keys: bool (False)
         Whether to extract keys along with the messages. If True, this will yield each message as a dict:
         {'key':msg.key(), 'value':msg.value()}
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -533,11 +533,11 @@ class FromKafkaBatched(Stream):
 
     def start(self):
         import confluent_kafka as ck
-        if self.engine == "cudf":
+        if self.engine == "cudf":  # pragma: no cover
             from custreamz import kafka
 
         if self.stopped:
-            if self.engine == "cudf":
+            if self.engine == "cudf": # pragma: no cover
                 self.consumer = kafka.Consumer(self.consumer_params)
             else:
                 self.consumer = ck.Consumer(self.consumer_params)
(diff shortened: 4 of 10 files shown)


streamz/sources.py
```diff
@@ -583,10 +593,15 @@ def from_kafka_batched(topic, consumer_params, poll_interval='1s',
     keys: bool (False)
         Whether to extract keys along with the messages. If True, this will yield each message as a dict:
         {'key':msg.key(), 'value':msg.value()}
+    engine: str (None)
+        If engine is "cudf", streamz reads data (messages must be JSON) from Kafka
+        in an accelerated manner directly into cudf dataframes.
+        Please refer to API here: github.com/jdye64/cudf/blob/kratos/python/custreamz/custreamz/kafka.py
```
chinmaychandak, line 599: This would be updated in the future.

chinmaychandak:
FYI: All Dask tests with `@gen_cluster` are failing for some reason. This is completely unrelated to this PR.

chinmaychandak:
Okay, so installing `pytest-tornasync` and making the tests with `@gen_cluster` start with `async def` worked locally.

martindurant:
> Can we do install pytest-tornasync on travisCI

I don't see why not

chinmaychandak:
> I don't see why not

I added it to travis, but tests are still failing. Can someone take a look?

chinmaychandak:
I see this got merged last week: https://github.com/dask/distributed/pull/3706. But marking tests with async def should work. 

Maybe we need to change from Python 3.6.

streamz/sources.py
```diff
@@ -503,7 +507,7 @@ def checkpoint_emit(_part):
                     try:
                         low, high = self.consumer.get_watermark_offsets(
                             tp, timeout=0.1)
-                    except (RuntimeError, ck.KafkaException):
+                    except (RuntimeError, ck.KafkaException, ValueError):
                         continue
```
jsmaupin, line 511: Do we need a sleep statement here to prevent a spam query for offsets?  Also, I wonder if we need logging to here to explain the reason for the `ValueError`

conda/environments/streamz_dev.yml
```diff
@@ -6,6 +6,7 @@ channels:
 dependencies:
   - python>=3.7
   - pytest
+  - pytest-tornasync
```
jsmaupin, line 9: Do we need to add this dependency if because of the introduction of the `async` keyword?

.travis.yml
```diff
@@ -7,7 +7,7 @@ language: python
 
 matrix:
   include:
-    - python: 3.6
+    - python: 3.7
```
jsmaupin, line 10: What does this fix?

streamz/tests/test_dask.py
```diff
@@ -15,7 +15,7 @@
 
 
 @gen_cluster(client=True)
-def test_map(c, s, a, b):
+async def test_map(c, s, a, b):
```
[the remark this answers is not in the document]
chinmaychandak: There's been an upstream change in Dask, that's causing tests unrelated to the PR to fail. Posted the relevant information and links in the comments.

streamz/sources.py
```diff
@@ -529,8 +533,14 @@ def checkpoint_emit(_part):
 
     def start(self):
         import confluent_kafka as ck
+        if self.engine == "cudf":
+            from custreamz import kafka
```
CJ-Wright, line 537: Is this a complete replacement for `ck`? (does it include `TopicPartition`)

chinmaychandak: Yes, if you see the API calls like `commit`, `committed`, `get_watermark_offsets`, they're the exact same as CK, including the `TopicPartition` class. We've tried to do this deliberately to keep code duplication as minimal as possible.

CJ-Wright: ok, why not do `from custreamz import kafka as ck` and remove the conditionals?

chinmaychandak: Oh, I see what you're saying. A little clarification here is that the API for custreamz.kafka are exactly same as CK in the sense that they take the `TopicPartition` object(s). But, the `TopicPartition` class does not exist in custreamz.kafka.

chinmaychandak: Deserialization happens in these API and there is not concept of `TopicPartition` since this a librdkafka C++ implementation underneath: https://github.com/jdye64/cudf/blob/7307cbe4f33a67c7d31a8d48955d755d55b03e9d/python/custreamz/custreamz/kafka.py#L72

CJ-Wright:
Should these changes also be made in `from_kafka`?

chinmaychandak:
> Should these changes also be made in `from_kafka`?

There's no checkpointing in `from_kafka`. Also, I'm assuming people wanting to leverage GPUs for streaming would use `from_kafka_batched` since that's the general convention to read from Kafka as it is faster. Even with CPUs, I think `from_kafka` should only be used for naive, simple use cases where the incoming data is low throughput.

chinmaychandak:
> I see this got merged last week: [dask/distributed#3706](https://github.com/dask/distributed/pull/3706). But marking tests with async def should work.

@CJ-Wright Any idea about how we can resolve this? Tests are passing locally.

jsmaupin:
It seems like maybe this was an effort to use native Python features in 3.x, because 2.x is no longer supported.  Tornado is python 2/3 compatible with how it does not use these new Python 3 keywords.

chinmaychandak:
Okay, with the upstream Dask error fixed: https://github.com/dask/distributed/pull/3738, all tests are now passing.

martindurant:
Clearly coverage suffers here since we don't test for CUDA on Travis. Indeed, that means that we take it on trust that this code works... Can you think of how we can get around that?

chinmaychandak:
> Clearly coverage suffers here since we don't test for CUDA on Travis. Indeed, that means that we take it on trust that this code works... Can you think of how we can get around that?

Yes, coverage definitely takes a small hit here. All the API that serve as a replacement for CK would be tested as part of cudf/custreamz nightly builds (we get a new build every 2-3 hours, so this is pretty reliable) on gpuCI. If those pass, then it is guaranteed that the individual API added as part of `custreamz.kafka` work just fine, and since the reading-in-batches-from-kafka logic remains the same, the `engine="cudf"` should work fine too. 

As for actually having a test for `from_kafka_batched` with `engine="cudf`, I do not know of any way we can do that here. @jdye64, any ideas?

martindurant:
At the minimum, I would suggest adding a badge to the front page (README) linking to those nightly tests, and describe this somewhere, and adding no-coverage markers to GPU-specific branches.

chinmaychandak:
> adding a badge

Should I name it as `gpuCI`?



> describe this somewhere

I have added the description to the doc string. Let me make it more elaborate. Would that be sufficient?



> no-coverage markers to GPU-specific branches

Would `# pragma: no cover` be okay to use, or should we prefer something else?

chinmaychandak:
> At the minimum, I would suggest adding a badge to the front page (README) linking to those nightly tests, and describe this somewhere, and adding no-coverage markers to GPU-specific branches.

Done. Please do have a look.

README.rst
```diff
@@ -21,3 +21,5 @@ BSD-3 Clause
    :alt: Documentation Status
 .. |Version Status| image:: https://img.shields.io/pypi/v/streamz.svg
    :target: https://pypi.python.org/pypi/streamz/
+.. |RAPIDS custreamz gpuCI| image:: https://img.shields.io/badge/gpuCI-custreamz-green
+   :target: https://github.com/jdye64/cudf/blob/kratos/python/custreamz/custreamz/kafka.py
```
chinmaychandak, line 25: This link will be updated in the future.

streamz/sources.py
```diff
@@ -583,10 +593,36 @@ def from_kafka_batched(topic, consumer_params, poll_interval='1s',
...
+        Folks interested in trying out custreamz would benefit from this accelerated Kafka reader.
+        If one does not want to use GPUs, they can use streamz as is, with the default engine=None.
 
-    Important Kafka Configurations:
+        To use this option, one must install RAPIDS custreamz using:
+        https://anaconda.org/rapidsai-nightly/custreamz
+        This will install all GPU dependencies, including streamz.
+
+        Please refer to RAPIDS custreamz.kafka API here:
+        github.com/jdye64/cudf/blob/kratos/python/custreamz/custreamz/kafka.py
```
chinmaychandak, line 617: This link will be updated in the future.

CJ-Wright:
LGTM to the extent that I understand this.

martindurant:
OK, putting it in then - thank you.
I hope you accept responsibility for cuda-related issues, @chinmaychandak  :)

chinmaychandak:
Thanks @martindurant, @CJ-Wright

chinmaychandak:
I will keep the links to custreamz.kafka and the gpuCI up to date.

PR #331, opened 2020-05-01
Update doc string links for accelerated Kafka datasource

chinmaychandak:
Updating doc string for #330.

### DIFF
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -597,22 +597,25 @@ def from_kafka_batched(topic, consumer_params, poll_interval='1s',
         If engine is set to "cudf", streamz reads data (messages must be JSON) from Kafka
         in an accelerated manner directly into cuDF (GPU) dataframes.
 
+        Please refer to RAPIDS cudf API here:
+        https://docs.rapids.ai/api/cudf/stable/
+
         This is done using the custreamz.kafka module in cudf. custreamz.kafka has the exact
-        same API as Confluent Kafka, so it serves as a drop-in replacement in from_kafka_batched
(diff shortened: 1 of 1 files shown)


streamz/sources.py
```diff
@@ -597,27 +597,32 @@ def from_kafka_batched(topic, consumer_params, poll_interval='1s',
...
 
         Folks interested in trying out custreamz would benefit from this accelerated Kafka reader.
         If one does not want to use GPUs, they can use streamz as is, with the default engine=None.
 
-        To use this option, one must install RAPIDS custreamz using:
-        https://anaconda.org/rapidsai-nightly/custreamz
-        This will install all GPU dependencies, including streamz.
+        To use this option, one must install custreamz (use the appropriate CUDA version recipe) using
+        the following command, which will install all GPU dependencies, including streamz:
+        conda env create --name custreamz --file
```
martindurant, line 617: Seems strange to require the user to create a new environment; also, this should only work with the raw version of the file, not the rendered github page

chinmaychandak: > Seems strange to require the user to create a new environment

That's true, but the alternative is to install quite a few packages instead. The .yml script installs everything cleanly. Also, this is temporary (as mentioned at the end of the doc string) until the code gets merged into cudf/custreamz, after which it be a single line nightly package install. 

> this should only work with the raw version of the file, not the rendered github page

Correct, I was assuming people would understand this by themselves, that they would need to download the .yml file and then run the command pointing it to that file.

martindurant: Perhaps better to write this out in full, even if it is only temporary

chinmaychandak: Done

PR #345, opened 2020-07-16
Tags for 0.5.4 release

chinmaychandak:
Just noticed that the tags needed an update with the new release. Do we need to make a change anywhere else in the code or docs?


martindurant:
d'oh :)

PR #347, opened 2020-08-03
fix: is_dataframe_like compatibility with pandas.DataFrame v1.1

### LINKED ISSUE #346: Stop using the presence of `value_counts` to determine that an object is not a pandas.DataFrame
From version 1.1 the `pandas.DataFrame` contains `value_counts` so the https://github.com/python-streamz/streamz/blob/de166a40babc9690be17d1b92dcd64fa5e149362/streamz/dataframe/utils.py#L7 doesn't work.

see an issue in dask - https://github.com/dask/dask/issues/5109

- https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.value_counts.html

bednar:
Closes #346 

Fixed checking of DataFrame for pandas 1.1+.

### DIFF
--- a/streamz/dataframe/utils.py
+++ b/streamz/dataframe/utils.py
@@ -7,11 +7,11 @@ import sys
 def is_dataframe_like(df):
     """ Looks like a Pandas DataFrame. ** Borrowed from dask.dataframe.utils ** """
     typ = type(df)
-    return (all(hasattr(typ, name)
-                for name in ('groupby', 'head', 'merge', 'mean'))
-            and all(hasattr(df, name) for name in ('dtypes',))
-            and not any(hasattr(typ, name)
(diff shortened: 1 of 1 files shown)


bednar:
The https://github.com/python-streamz/streamz/blob/de166a40babc9690be17d1b92dcd64fa5e149362/streamz/dataframe/tests/test_dataframes.py#L782 should be updated, because currently is not compatible with panda.DataFrame v1.1

martindurant:
> should be updated

Do you intend to do this?

bednar:
Nope, I tried it but it is over my capabilities :(

martindurant:
Fix was merged in #348 , including for series and index. OK to close here?

bednar:
Yes, thanks for Fix.

PR #348, opened 2020-08-06
Update docstring for custreamz accelerated Kafka reader

chinmaychandak:
@martindurant, @CJ-Wright Could you please have a look? 

I have also added `|` at the start of each line to ensure correct indentation, since currently the indentation doesn't reflect correctly: https://streamz.readthedocs.io/en/latest/api.html#streamz.from_kafka_batched

### DIFF
--- a/streamz/dataframe/tests/test_dataframes.py
+++ b/streamz/dataframe/tests/test_dataframes.py
@@ -275,8 +275,9 @@ def test_getitem(stream):
                                      lambda a: ['x']])
 @pytest.mark.parametrize('indexer', [lambda g: g.y,
                                      lambda g: g,
-                                     lambda g: g[['y']],
-                                     lambda g: g[['x', 'y']]])
+                                     lambda g: g[['y']]
+                                     #lambda g: g[['x', 'y']]
+                                    ])
 def test_groupby_aggregate(agg, grouper, indexer, stream):
     df = pd.DataFrame({'x': (np.arange(10) // 2).astype(float), 'y': [1.0, 2.0] * 5})
 
@@ -723,7 +724,8 @@ def test_windowing_n(func, n, getter):
 @pytest.mark.parametrize('indexer', [lambda g: g.x,
                                      lambda g: g,
                                      lambda g: g[['x']],
-                                     lambda g: g[['x', 'y']]])
+                                     #lambda g: g[['x', 'y']]
+                                    ])
 def test_groupby_windowing_value(func, value, getter, grouper, indexer):
     index = pd.date_range(start='2000-01-01', end='2000-01-03', freq='1h')
     df = pd.DataFrame({'x': np.arange(len(index), dtype=float),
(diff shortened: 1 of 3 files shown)


chinmaychandak:
Did my changes just break CI? :(

martindurant:
Seems to be failing `is_frame_like(example)`, perhaps because of pandas release?

martindurant:
@chinmaychandak , do you have a screenshot of the difference? I have never used "|" characters like this in docstrings.

chinmaychandak:
> do you have a screenshot of the difference?

You mean generate the docs locally first? I haven't done this yet. 

> I have never used "|" characters like this in docstrings.

Neither have I. I just read on [Stack Overflow](https://stackoverflow.com/questions/7033239/how-to-preserve-line-breaks-when-generating-python-docs-using-sphinx) that Sphinx-generated RST docs need a `|` to force a newline.

chinmaychandak:
> Seems to be failing is_frame_like(example), perhaps because of pandas release?

Let me take a look.

martindurant:
> I haven't done this yet.

Please do, it will help us

chinmaychandak:
@martindurant Here's the difference, seems to be working well. Need to figure out the tests failing though. 

image

chinmaychandak:
> Seems to be failing `is_frame_like(example)`, perhaps because of pandas release?

The dask function for `is_dataframe_like` seems to have changed: https://github.com/dask/dask/blob/58bab6b2ad352fa15b9db6fba0d50a43925b9e34/dask/utils.py#L1118.

Should I make this change here, too?

martindurant:
In utils we have 
```
def is_dataframe_like(df):
    """ Looks like a Pandas DataFrame. ** Borrowed from dask.dataframe.utils ** """
    typ = type(df)
    return (all(hasattr(typ, name)
                for name in ('groupby', 'head', 'merge', 'mean'))
            and all(hasattr(df, name) for name in ('dtypes',))
            and not any(hasattr(typ, name)
                        for name in ('value_counts', 'dtype')))
```
but dask's definition has been updated to:
```
def is_dataframe_like(df):
    """ Looks like a Pandas DataFrame """
    typ = type(df)
    return (
        all(hasattr(typ, name) for name in ("groupby", "head", "merge", "mean"))
        and all(hasattr(df, name) for name in ("dtypes", "columns"))
        and not any(hasattr(typ, name) for name in ("name", "dtype"))
    )
```
I suggest updating with the current dask version.

Also should check for the series and index variants

chinmaychandak:
The `sdf.groupby()` and `sdf.window().groupby()` objects were failing when using the `std()` aggregation with overlapping indexes. I have removed these tests for now. Seems random, not sure why this is happening.

chinmaychandak:
Does this look good now?

chinmaychandak:
@martindurant, should we cut a release because these changes might be needed for users with the new Pandas versions?

martindurant:
Decent idea. Nudge me again next week if nothing happens :)

chinmaychandak:
Cool, thanks Martin!

martindurant:
0.5.5 on pypi

chinmaychandak:
Awesome, thanks a lot Martin. Should we update the tags as well?

martindurant:
https://github.com/python-streamz/streamz/releases/tag/0.5.5

Issue #336, opened 2020-05-18
Problem processing dictionary type events using dask-distributed


pablo-campillo:
I've realized that events of dict type are not processed properly when using dask-distributed. It seems that only the first event is processed and replicated.

streamz=0.5.3
dask=2.14.0

**Sequential version:**

```python
from streamz import Stream
import random
import time

def my_function(record):
    record['x'] = random.randint(0,100)
    return record

source = Stream()
a = source.map(my_function).sink(print)

for i in range(10):
    source.emit({'i': i})
```

Output:

```
{'i': 0, 'x': 44}
{'i': 1, 'x': 28}
{'i': 2, 'x': 5}
{'i': 3, 'x': 0}
{'i': 4, 'x': 75}
{'i': 5, 'x': 37}
{'i': 6, 'x': 70}
{'i': 7, 'x': 29}
{'i': 8, 'x': 16}
{'i': 9, 'x': 78}
```

**Parallel version:**

```python
from dask.distributed import Client
from streamz import Stream
import random
import time

client = Client()

def my_function(record):
    record['x'] = random.randint(0,100)
    return record

source = Stream()
a = source.scatter().map(my_function).buffer(10).gather().sink(print)

for i in range(10):
    source.emit({'i': i})
```

Output:

```
{'i': 0, 'x': 91}
{'i': 0, 'x': 91}
{'i': 0, 'x': 91}
{'i': 0, 'x': 91}
{'i': 0, 'x': 91}
{'i': 0, 'x': 91}
{'i': 0, 'x': 91}
{'i': 0, 'x': 91}
{'i': 0, 'x': 91}
{'i': 0, 'x': 91}
```

CJ-Wright:
Does this work correctly with just Dask?

pablo-campillo:
I've never used Dask. I've tryed a Dask version and it works, but I don't know if it is equivalent:

```python
from dask.distributed import Client
import random

client = Client()

def my_function(record):
    record['x'] = random.randint(0,100)
    return record

futures = client.map(my_function, [{'i': i} for i in range(10)])
results = client.gather(futures)

print(results)
```

martindurant:
Can you please check what's going on in the dask dashboard? My inkling is, that perhaps the data is not getting tokenised properly, so you are getting the same result back no matter what you send.

pablo-campillo:
You can find bellow the output of the dashboard. Using stream only one task is executed.

Streamz distributed version:

[...]

PR #349, opened 2020-08-12
Wrap the items passed to `client.scatter` with a dict

nils-braun:
As discussed in #336 dasks `client.scatter` treats dicts and lists different from any other object. This is good for raw dask usages, but not wanted in streamz. Therefore I always wrap with a dict now, so that we get consistent behavior.
The key of the dict will be used to identify the data on the cluster. I am following the same convention as dask itself by choosing the key.

### DIFF
--- a/streamz/dask.py
+++ b/streamz/dask.py
@@ -109,7 +109,7 @@ class scatter(DaskStream):
         # lists and dicts. So we always use a dict here to be sure
         # we know the format exactly. The key will be taken as the
         # dask identifier of the data.
-        tokenized_x = type(x).__name__ + "-" + tokenize(x)
+        tokenized_x = f"{type(x).__name__}-{tokenize(x)}"
         future_as_dict = yield client.scatter({tokenized_x: x}, asynchronous=True)
         future = future_as_dict[tokenized_x]
         f = yield self._emit(future, metadata=metadata)
--- a/streamz/dask.py
+++ b/streamz/dask.py
@@ -5,6 +5,7 @@ from operator import getitem
 from tornado import gen
 
 from dask.compatibility import apply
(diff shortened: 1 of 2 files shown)


streamz/dask.py
```diff
@@ -103,7 +104,14 @@ def update(self, x, who=None, metadata=None):
         client = default_client()
 
         self._retain_refs(metadata)
-        future = yield client.scatter(x, asynchronous=True)
+        # We need to make sure that x is treated as it is by dask
+        # However, client.scatter works internally different for
+        # lists and dicts. So we always use a dict here to be sure
+        # we know the format exactly. The key will be taken as the
+        # dask identifier of the data.
+        tokenized_x = type(x).__name__ + "-" + tokenize(x)
```
CJ-Wright, line 112: I would have done this as an fstring

streamz/tests/test_dask.py
```diff
@@ -28,6 +28,28 @@ def test_map(c, s, a, b):
     assert all(isinstance(f, Future) for f in futures_L)
 
 
+@gen_cluster(client=True)
+def test_map_on_dict(c, s, a, b):
+    # dask treats dicts differently, so we have to make sure
+    # the user sees no difference in the streamz api.
+    # Regression test agains #336
```
CJ-Wright, line 35:
```suggestion
    # Regression test against #336
```

streamz/dask.py
martindurant: We *just* moved to py>=3.6 :)

nils-braun: Nice :-) I just copied this from dask, but you are very correct!

nils-braun:
Thanks for the review!

PR #358, opened 2020-09-01
Make npartitions optional when streaming from Kafka and set auto.offset.reset to latest as default.

chinmaychandak:
We currently need to specify the number of topic partitions (`npartitions`) to start a stream from Kafka, which is not ideal in case of production workloads where we need to add partitions to a topic on the fly, etc. The stream should be able to handle these changes on its own. No other streaming stacks need the end-user to specify #partitions for a stream. I am using Confluent Kafka's [list_topics](https://docs.confluent.io/current/clients/confluent-kafka-python/#confluent_kafka.Consumer.list_topics) API for this purpose. 

**[EDIT]** On further discussion (please read entire thread below), I have changed `npartitions` to default to `None`. I have skipped the auto-refresh capability (to support adding partitions to a topic on the fly and being handled by the stream) for now. I will add it in a follow-up PR via an additional parameter as discussed below. This PR just basically targets making the requirement of specifying number of partitions optional.

**NOTE:**
This PR also changes the default `auto.offset.reset` Kafka consumer config to `latest` conforming to the librdkafka configuration here. I needed to fix a small condition in addition to this to make sure checkpointing was working with this latest config. Currently, if a stream with `auto.offset.reset: latest` restarts, the offsets are again set to `latest` offsets (`high` watermark offset) rather than using the checkpoint created by the first run. This needed fixing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ tests = [p + '.tests' for p in packages]
 
 
 setup(name='streamz',
-      version='0.5.6',
+      version='0.6.0',
       description='Streams',
       url='http://github.com/python-streamz/streamz/',
       maintainer='Matthew Rocklin',
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ tests = [p + '.tests' for p in packages]
 
 
 setup(name='streamz',
-      version='0.5.5',
+      version='0.5.6',
       description='Streams',
       url='http://github.com/python-streamz/streamz/',
       maintainer='Matthew Rocklin',
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ tests = [p + '.tests' for p in packages]
 
 
 setup(name='streamz',
-      version='0.5.6',
+      version='0.5.5',
       description='Streams',
       url='http://github.com/python-streamz/streamz/',
       maintainer='Matthew Rocklin',
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -459,7 +459,7 @@ class FromKafkaBatched(Stream):
         # Override the auto-commit config to enforce custom streamz checkpointing
         self.consumer_params['enable.auto.commit'] = 'false'
         if 'auto.offset.reset' not in self.consumer_params.keys():
-            consumer_params['auto.offset.reset'] = 'earliest'
+            consumer_params['auto.offset.reset'] = 'latest'
         self.topic = topic
         self.npartitions = npartitions
         self.poll_interval = convert_interval(poll_interval)
@@ -484,6 +484,11 @@ class FromKafkaBatched(Stream):
             ref = RefCounter(cb=lambda: commit(_part))
             yield self._emit(_part, metadata=[{'ref': ref}])
 
+        if self.npartitions is None:
+            kafka_cluster_metadata = self.consumer.list_topics(self.topic)
+            self.npartitions = len(kafka_cluster_metadata.topics[self.topic].partitions)
+        self.positions = [0] * self.npartitions
+
         tps = []
         for partition in range(self.npartitions):
             tps.append(ck.TopicPartition(self.topic, partition))
@@ -501,10 +506,6 @@ class FromKafkaBatched(Stream):
         try:
             while not self.stopped:
                 out = []
-                # kafka_cluster_metadata = self.consumer.list_topics(self.topic)
-                # self.npartitions = len(kafka_cluster_metadata.topics[self.topic].partitions)
-                # if self.npartitions > len(self.positions):
-                #     self.positions.extend([0] * (self.npartitions - len(self.positions)))
                 for partition in range(self.npartitions):
                     tp = ck.TopicPartition(self.topic, partition, 0)
                     try:
@@ -546,12 +547,6 @@ class FromKafkaBatched(Stream):
             else:
                 self.consumer = ck.Consumer(self.consumer_params)
             self.stopped = False
-
-            if self.npartitions is None:
-                kafka_cluster_metadata = self.consumer.list_topics(self.topic)
-                self.npartitions = len(kafka_cluster_metadata.topics[self.topic].partitions)
-            self.positions = [0] * self.npartitions
-
             tp = ck.TopicPartition(self.topic, 0, 0)
 
             # blocks for consumer thread to come up
@@ -597,6 +592,8 @@ def from_kafka_batched(topic, consumer_params, poll_interval='1s',
     npartitions: int (None)
         | Number of partitions in the topic.
         | If None, streamz will poll Kafka to get the number of partitions.
+        | As of now, streamz does not support changing number of partitions on the fly.
+        | It is recommended to restart the stream after changing the number of partitions.
     start: bool (False)
         Whether to start polling upon instantiation
     max_batch_size: int
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -624,13 +624,12 @@ def from_kafka_batched(topic, consumer_params, poll_interval='1s',
 
         | More information at: https://rapids.ai/start.html
 
-
     Important Kafka Configurations
     ----------
-    If 'auto.offset.reset': 'latest' is set in the consumer configs,
-    the stream starts reading messages from the latest offset. Else,
-    if it's set to 'earliest', it will read from the start offset.
-
+    By default, a stream will start reading from the latest offsets
+    available. Please set 'auto.offset.reset': 'earliest' in the
+    consumer configs, if the stream needs to start processing from
+    the earliest offsets.
 
     Examples
     ----------
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -462,6 +462,8 @@ class FromKafkaBatched(Stream):
             consumer_params['auto.offset.reset'] = 'latest'
         self.topic = topic
         self.npartitions = npartitions
+        if self.npartitions is not None and self.npartitions <= 0:
+            raise ValueError("Number of Kafka topic partitions must be > 0.")
         self.poll_interval = convert_interval(poll_interval)
         self.max_batch_size = max_batch_size
         self.keys = keys
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -453,7 +453,7 @@ class from_kafka(Source):
 class FromKafkaBatched(Stream):
     """Base class for both local and cluster-based batched kafka processing"""
     def __init__(self, topic, consumer_params, poll_interval='1s',
-                 max_batch_size=10000, keys=False,
+                 npartitions=None, max_batch_size=10000, keys=False,
                  engine=None, **kwargs):
         self.consumer_params = consumer_params
         # Override the auto-commit config to enforce custom streamz checkpointing
@@ -461,6 +461,7 @@ class FromKafkaBatched(Stream):
         if 'auto.offset.reset' not in self.consumer_params.keys():
             consumer_params['auto.offset.reset'] = 'earliest'
         self.topic = topic
+        self.npartitions = npartitions
         self.poll_interval = convert_interval(poll_interval)
         self.max_batch_size = max_batch_size
         self.keys = keys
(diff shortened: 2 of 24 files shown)


martindurant:
I think this is fine, except surely there isn't a need to list the partitions on every single loop? I don't know what the overhead of that operation is, but it must add something.

> Please do not merge

Put this into the title or use a draft PR?

> I can add a test if needed

Yes please, both for increase and decrease of the partition count

chinmaychandak:
> I think this is fine, except surely there isn't a need to list the partitions on every single loop? I don't know what the overhead of that operation is, but it must add something.

Yeah, there would be some overhead associated, although there's two things here:

1. If we do not poll every time, and there's a partition count increase, it would cause a delay. How do we set the period at which we poll? 
2. After every poll interval, there is considerable latency reading a batch of messages — I do not think a single extra call per poll interval would add much overhead compared to that, especially since we are getting the cluster metadata for that single topic -- which would typically be smaller than a single Kafka message.

martindurant:
I feel like perhaps it can be optional. Agree that the user shouldn't need to specify the number, but in many cases you at least expect the number to stay static.

chinmaychandak:
> I feel like perhaps it can be optional.

So, basically, just set `npartitions=None` instead of `1` by default. If the user provides a value, we end up using that. If not, then use this logic?

I was also thinking of a more hacky way of implementing this. How about we set a `MAX_ALLOWED_PARTITIONS=128` and then the for loop just iterates from 1 to `MAX_ALLOWED_PARTITIONS`; the partition at which `get_watermark_offsets` throws the `KafkaException: KafkaError{code=_UNKNOWN_PARTITION,val=-190,str="Failed to get watermark offsets: Local: Unknown partition"}`, we break out of the loop and move on the next steps in the stream.

chinmaychandak:
> Yes please, both for increase and decrease of the partition count

I have almost never seen decreasing the partitions of a Kafka topic. It's really hard for Kafka to migrate data and not recommended. You generally only increase. If you need to decrease, it's better to delete the topic and create a new one.

chinmaychandak:
What do you think of the hacky solution @martindurant?



> So, basically, just set `npartitions=None` instead of `1` by default. If the user provides a value, we end up using that. If not, then use this logic? I still feel removing `npartitions` would be better in the long run, though.

Do you feel this is the correct way of making it optional? Although I feel it would be overcomplicating things.

martindurant:
Yes, I think that allowing specifying the number of partitions is OK, and doing a list the first time to get that number if passed None. *In addition*, there can be an option to check the number every N cycles (or every Y seconds).

chinmaychandak:
@martindurant Makes sense. 

I have changed `npartitions` to default to `None`. Also, I have skipped the auto-refresh capability of the number of partitions for now. I will add it in a follow-up PR via an additional parameter.

I have also changed the default `auto.offset.reset` Kafka consumer config to `latest` conforming to the librdkafka configuration [here](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md). 

I needed to fix a small condition in addition to this to make sure checkpointing was working with this `latest` config. Currently, if a stream with `auto.offset.reset: latest` restarts, the offsets are again set to `latest` offsets (high watermark offset) rather than using the checkpoint created by the first run. This needed fixing.

Let me know if this looks good. We are still working on adding the `list_topics()` API to custreamz, hopefully the DRAFT tag can be removed from this PR early next week.

martindurant:
Yes, I think this is along the right lines.

> I have also changed the default auto.offset.reset Kafka consumer config to latest conforming to the librdkafka configuration here.

This should be documented for the source class. I think most other sources (like file) will start from the earliest data.

chinmaychandak:
> This should be documented for the source class. 

Done

chinmaychandak:
@martindurant This is ready for review now, I have removed the DRAFT label. Apologies for the delay, unexpected hiccups slowed this down.

setup.py
```diff
@@ -9,7 +9,7 @@
 
 
 setup(name='streamz',
-      version='0.5.6',
+      version='0.5.5',
```
CJ-Wright: ?

streamz/sources.py
```diff
@@ -510,7 +517,9 @@ def checkpoint_emit(_part):
                     except (RuntimeError, ck.KafkaException):
                         continue
                     if 'auto.offset.reset' in self.consumer_params.keys():
-                        if self.consumer_params['auto.offset.reset'] == 'latest':
+                        if self.consumer_params['auto.offset.reset'] == 'latest' and \
+                                (self.positions == [-1001] * self.npartitions
```
martindurant, line 521: Where does -1001 come from?
In this expression, I would prefer parentheses to make clear which `and`/`or` goes together.

streamz/tests/test_kafka.py
```diff
@@ -223,6 +223,7 @@ def test_kafka_batch():
         stream.upstream.stopped = True
 
 
+@flaky(max_runs=3, min_passes=1)
```
martindurant, line 226: Any idea why? We would rather not have this. How frequently does it fail?

streamz/tests/test_kafka.py
```diff
@@ -291,11 +294,13 @@ def test_kafka_dask_checkpointing_sync_nodes(c, s, w1, w2):
...
             'group.id': 'streamz-test%i' % j1,
-            'enable.auto.commit': False}
+            'enable.auto.commit': False,
+            'auto.offset.reset': 'earliest'}
     j2 = j1 + 1
     ARGS2 = {'bootstrap.servers': 'localhost:9092',
             'group.id': 'streamz-test%i' % j2,
-            'enable.auto.commit': False}
+            'enable.auto.commit': False,
+            'auto.offset.reset': 'earliest'}
```
martindurant, line 303: We could do with some more tests, e.g., with different number of partitions and whether the user passes npartitions or not (and what if the number passed is larger or smaller than that set in kafka); and the offsets earliest/latest. 

This checkpointing test is I suppose a separate thing, but it would be nice to see all the combinations tested somewhere.

streamz/sources.py
chinmaychandak: > Where does -1001 come from?

The default value returned by `consumer.committed()` which is assigned to `self.positions` is `-1001`. 

>  parentheses to make clear which `and`/`or` goes together

There's one already there. It shows `latest` `and` (`-1001` `or` `0`). Did you mean another parenthesis around the entire expression?

streamz/tests/test_kafka.py
chinmaychandak: Nope, no idea at all. This was really surprising for me, too. Unlike other flaky tests, this one passes >90% of the time, but I did see it fail one time, and succeed the next. Hence, I thought adding `flaky` would be best. Should I remove it?

martindurant: Do you believe it was already failing sometimes before, or did something change?

streamz/tests/test_kafka.py
chinmaychandak: > with different number of partitions

Yes, we need this test.

> whether the user passes npartitions or not

We already have this, correct? But I guess none of the tests have the `npartitions` parameter specified  (i.e., all of them use the new default value of `None`), which I think is ok. But yeah, we can add one with `npartitions` maybe specified to something >1. 

> and what if the number passed is larger or smaller than that set in kafka

Not sure I understand. The support for adding dynamic partitions on the fly would be a follow-up PR like we discussed (auto-refresh partitions every `N` cycles based on a user parameter).

 > offsets earliest/latest

These have been covered extensively, although they are spread across different tests. 

Should I add a test with `npartitions` specified and >1 for now? Or can it be in a separate PR with the adding partitions on the fly capability?

streamz/tests/test_kafka.py
chinmaychandak: Maybe it was a one-off, although it is concerning.

chinmaychandak: I don't think anything changed, but I did recall Jarod saying that it failed for him once too, and this was way before the checkpointing stuff.

streamz/tests/test_kafka.py
martindurant: I think certainly we should add npartitions=-1, 0, 1, 2, 4 as a set of tests where the number of real partitions in kafka is 2 (fixed). The first two should fail with ValueError, the middle one should get about half the data, and the last two should both get all the data, but the last one will have partitions that never receive anything. Does that sound right?

streamz/tests/test_kafka.py
martindurant: I would be tempted to leave it as was, and see how often this arises. We can investigate at a later date whats going on - I don't suppose we even know at what stage things are hanging.

streamz/tests/test_kafka.py
chinmaychandak: For npartitions=-1/0, it doesn't enter the polling consumption [loop](https://github.com/chinmaychandak/streamz/blob/ccc7d428ab24939fe07b151a7469a335f853ef9b/streamz/sources.py#L512). 

For npartitions=1, yes, about half the data. 

For npartitions=2/4, yes, they get all the data, but in the last one, partitions 3 & 4 are skipped in every polling consumption loop because the `get_watermarks_offset` throws an [error](https://github.com/chinmaychandak/streamz/blob/ccc7d428ab24939fe07b151a7469a335f853ef9b/streamz/sources.py#L517). 

All of these are handled within `from_kafka_batched` gracefully IMO (we can add a check at the beginning itself that raises a ValueError if npartitions is <=0, if need be), so we might not need tests. However, I do think we need a test with `npartitions` specified explicitly with a value `>1`, and maybe it can be plugged inside the test with the auto-refresh capability.

What do you think?

streamz/tests/test_kafka.py
chinmaychandak: Gotcha, will remove the flaky marker then.

chinmaychandak: Actually, I'm running all tests locally, and this test seems to fail like 1 out of 5 times.

streamz/tests/test_kafka.py
chinmaychandak: Added some tests, do have a look

chinmaychandak: @martindurant, please do let me know if we need to add anything else to this to get it rolled in. If not, there's another PR for auto-refreshing partitions on the way! :)

streamz/tests/test_kafka.py
```diff
@@ -61,6 +50,7 @@ def stop_docker(name='streamz-kafka', cid=None, let_fail=False):
 
 def launch_kafka():
     stop_docker(let_fail=True)
+    os.system("docker pull spotify/kafka")
```
martindurant, line 53: Strictly, docker should do this when you ask to run on the next line, and `os.system` is [discouraged](https://docs.python.org/3.8/library/subprocess.html#replacing-os-system) versus `subprocess.call` here and below.
It would look like (NB: better execute command directly than spawn a shell)
```
subprocess.call(shlex.split("docker pull spotify/kafka"))
```
(if indeed you want to keep the extra line, which is fine)

streamz/tests/test_kafka.py
```diff
@@ -244,6 +234,63 @@ def test_kafka_dask_batch(c, s, w1, w2):
...
+        time.sleep(5)
+
+        for i in range(10):
+            if i % 2 == 0:
+                kafka.produce(TOPIC, b'value-%d' % i, partition=0)
+            else:
+                kafka.produce(TOPIC, b'value-%d' % i, partition=1)
+        kafka.flush()
+
+        try:
```
martindurant, line 265:
```python
with pytest.raises(ValueError):
```

martindurant (approved):
Sorry, just a couple of style requests, then good to go.

streamz/tests/test_kafka.py
chinmaychandak: Doh, don't know how I missed this!

streamz/tests/test_kafka.py
chinmaychandak: Good point, I didn't know about this. Thanks for pointing this out!

martindurant:
Merging now - travis is not reporting the test success for some reason.

Issue #357, opened 2020-08-31
[BUG] Window operations return incorrect output if beginning/intermediate indexes are empty.


chinmaychandak:
The code below:
 
```
from streamz import Stream
from streamz.dataframe import DataFrame
import pandas as pd

# Only pass records with "amount" = 5 into the window, use the "ts" column as datetime index.
def preprocess(pdf):
    mask = pdf["amount"] == 5
    pdf = pdf.loc[mask]
    pdf["ts"] = pdf["ts"].astype("datetime64[ms]")
    pdf.index = pdf["ts"]
    return pdf

source = Stream()

stream = source.map(preprocess)

example = pd.DataFrame({"amount":[], "ts":[]})
sdf = DataFrame(stream=stream, example=example)

sdf.window("1h").amount.sum().stream.sink(print)

source.emit(pd.DataFrame({"amount":[1,2,3], "ts":["2050-01-01 00:00:00", "2050-01-01 01:00:00", "2050-01-01 02:00:00"]}))

source.emit(pd.DataFrame({"amount":[5,5,5], "ts":["2050-01-01 03:00:00", "2050-01-01 04:00:00", "2050-01-01 05:00:00"]}))
```

gives the output:

```
0
15
```

instead of 

```
0
5
```

If I replace the emits to:

```
source.emit(pd.DataFrame({"amount":[5,2,3], "ts":["2050-01-01 00:00:00", "2050-01-01 01:00:00", "2050-01-01 02:00:00"]}))

source.emit(pd.DataFrame({"amount":[5,5,5], "ts":["2050-01-01 03:00:00", "2050-01-01 04:00:00", "2050-01-01 05:00:00"]}))
```

the output is 

```
5
5
``` 
as expected.

I believe the issue is [here](https://github.com/python-streamz/streamz/blob/239b807fd70848f851570463329662ce92c227a4/streamz/dataframe/aggregations.py#L206): if the dataframe that enters the SDF-window op. is empty, then the deque that gets created/updated in the `diff_loc` causes issues.

Any suggestions on what would the best way to fix this?

martindurant:
I could imagine changing
```
dfs.append(new)
```
to
```
if not new.empty:
    dfs.append(new)
```
(or perhaps put the empty ones directly into `old`?)

[...]

PR #360, opened 2020-09-04
Fix windowing over time bugs

chinmaychandak:
This PR fixes two windowing bugs:

1. Incorrect output when intermediate indexes in windowing-over-time operations are empty: #357  

2. Make windowing-over-time output conform with Pandas windowing-over-time. E.g. The code below give output as 10, but the current output is 15.

```
from streamz import Stream
from streamz.dataframe import DataFrame
import pandas as pd

stream = Stream()

example = pd.DataFrame({"amount":[]})
sdf = DataFrame(stream=stream, example=example)

sdf.window("2h").sum().stream.sink(print)

stream.emit(pd.DataFrame({"amount":[5,5,5]}, index=[pd.Timestamp("2050-01-01 00:00:00"),
                                                    pd.Timestamp("2050-01-01 01:00:00"),
                                                    pd.Timestamp("2050-01-01 02:00:00")]))
```


streamz/dataframe/aggregations.py
```diff
@@ -167,18 +167,20 @@ def diff_iloc(dfs, new, window=None):
         List of dataframes to decay
     """
     dfs = deque(dfs)
-    dfs.append(new)
+    if len(new) > 0:
```
martindurant, line 170: ~`not new.empty`~

^ it turns out that using `len` is consistently faster than using `empty`, I can't imagine why.

streamz/dataframe/aggregations.py
```diff
@@ -202,16 +204,18 @@ def diff_loc(dfs, new, window=None):
...
-        old.append(o)  # TODO: avoid copy if fully lost
-        dfs[0] = dfs[0].iloc[len(o):]
-        if not len(dfs[0]):
-            dfs.popleft()
+    if len(dfs) > 0:
+        mx = max(df.index.max() for df in dfs)
+        mn = mx - pd.Timedelta(window) + pd.Timedelta('1ns')
+        while pd.Timestamp(dfs[0].index.min()) < mn:
+            o = dfs[0].loc[:mn]
+            old.append(o)  # TODO: avoid copy if fully lost
```
martindurant, line 215: May as well check for an empty `old` here and fix the TODO?

martindurant: `old = old.append(o) if len(old) else o`. (+ comment to explain)

streamz/dataframe/tests/test_dataframes.py
```diff
@@ -755,12 +755,11 @@ def f(x):
...
 
     first = df.iloc[:diff]
-    lost = first[first.index.min() + value:]
-    first = first.iloc[len(lost):]
+    first = first[first.index.max() - value + pd.Timedelta('1ns'):]
 
     assert_eq(L[0], f(first))
 
-    last = df.loc[index.max() - value + pd.Timedelta('1s'):]
+    last = df.loc[index.max() - value + pd.Timedelta('1ns'):]
```
martindurant, line 762: Should include your specific test case in the suite?

chinmaychandak: Done

streamz/dataframe/aggregations.py
chinmaychandak: Done

chinmaychandak:
@martindurant Does this look good now?

martindurant:
OK, accepting this. I feel like we could still use more testing here, there are probably more edges.

chinmaychandak:
@martindurant, can we cut a release, please? Because these are pretty major windowing-over-time bugs we're seeing in our production pipelines.   

Sorry for the continual tagging in advance!

martindurant:
0.5.6 now on pypi

chinmaychandak:
Just curious, does the regro-cf-autotick-bot take care of the conda release automatically? I see that the previous releases were handled by the conda-forge streamz-feedstock.

martindurant:
Yes: https://github.com/conda-forge/streamz-feedstock/pull/13

PR #363, opened 2020-09-23
Added PeriodicDataFrame

jbednar:
This PR adds a PeriodicDataFrame that is essentially just streamz.dataframe.Random but with the callback parameterized so that users can supply their own function. The resulting object allows essentially arbitrary batched or non-batched callbacks without a user having to understand any of the complicated streamz logic; they just write a simple function and get back a periodically accumulating or updating streaming dataframe.

You can run a reproducible example of roughly this code by downloading the project at [bit.ly/streaming_hvplot](https://bit.ly/streaming_hvplot).  In a few lines that don't involve any tricky concepts (skip to the line starting "import pandas as pd, time, psutil" in tornado_stream.ipynb), users can make an entire streaming viz app with custom calls to system libraries to get real-time data. Following the main streamz API to do this is daunting for most users, but it's trivial with PeriodicDataFrame.

Open questions:
1. Is it cleaner if the callback takes **kwargs, so that we can add additional information there later without breaking existing callbacks?  I don't have any such information in mind now, but one never does...
2. The "freq" parameter is actually an interval (a time duration), not a frequency (cycles per time duration). Not sure if it's worth renaming, since it was already in Random.
3. Random was marked "experimental", but it's being used as an example in several external frameworks, including Panel and hvPlot. I've kept it; it adds only a few lines and no extra complexity.
4. The name could also be TornadoPeriodicDataFrame, because it's specific to the Tornado IOLoop, but I think that would likely just confuse a user that would want this convenient function.

### DIFF
--- a/docs/source/dataframes.rst
+++ b/docs/source/dataframes.rst
@@ -190,7 +190,8 @@ events, and topologies, but what if you simply want to run some Python
 function periodically and collect or plot the results?
 
 streamz provides a high-level convenience class for this purpose, called
-a PeriodicDataFrame. A PeriodicDataFrame uses Tornado's event loop to
+a PeriodicDataFrame. A PeriodicDataFrame uses Python's asyncio event loop
+(used as part of Tornado in Jupyter and other interactive frameworks) to
 call a user-provided function at a regular interval, collecting the results
 and making them available for later processing.
 
--- a/docs/source/dataframes.rst
+++ b/docs/source/dataframes.rst
@@ -221,7 +221,7 @@ results of system calls or other isolated actions, but any number of
 entries can be returned by the dataframe in a single batch. To
 facilitate collecting such batches, the callback is invoked with
 ``last`` (the time of the previous invocation) and ``now`` (the
-time of the current invocation. The callback can then generate or
+time of the current invocation). The callback can then generate or
 query for just the values in that time range.
 
 Arbitrary keyword arguments can be provided to the PeriodicDataFrame
@@ -252,4 +252,3 @@ Once you have a PeriodicDataFrame defined using such callbacks, you
 can then use all the rest of the functionality supported by streamz,
 including aggregations, rolling windows, etc., and streaming
 `visualization. <plotting>`_
-
--- a/docs/source/dataframes.rst
+++ b/docs/source/dataframes.rst
@@ -210,7 +210,7 @@ e.g. every 300 milliseconds:
 
 .. code-block:: python
 
-   df = PeriodicDataFrame(interval='300ms', datafn=random_datapoint)
+   df = PeriodicDataFrame(random_datapoint, interval='300ms')
 
 ``df`` will now be a steady stream of whatever values are returned by
 the `datafn`, which can of course be any Python code as long as it
@@ -220,13 +220,14 @@ Here we returned only a single point, appropriate for streaming the
 results of system calls or other isolated actions, but any number of
 entries can be returned by the dataframe in a single batch. To
 facilitate collecting such batches, the callback is invoked with
-``last`` (the time of the previous invocation) and ``now`` (the
-time of the current invocation). The callback can then generate or
-query for just the values in that time range.
+keyword arguments ``last`` (the time of the previous invocation) and
+``now`` (the time of the current invocation) as Pandas Timestamp
+objects. The callback can then generate or query for just the values
+in that time range.
 
 Arbitrary keyword arguments can be provided to the PeriodicDataFrame
-constructor, which will be passed into the `datafn` when it is
-called so that its behavior can be parameterized.
+constructor, which will be passed into the callback that its behavior
+can be parameterized.
 
 For instance, you can write a callback to return a suitable number of
 datapoints to keep a regularly updating stream, generated randomly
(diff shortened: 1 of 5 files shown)


streamz/dataframe/core.py
```diff
@@ -874,6 +880,22 @@ def _cb(interval, freq, source, continue_):
             last = now
 
 
+class Random(PeriodicDataFrame):
+    """Same as PeriodicDataFrame, but initialized with 3 random streams
+    (Kept for backwards compatibility and examples; otherwise not useful.)
```
martindurant, line 887: "otherwise not useful" - still useful! :)

martindurant (commented):
I think this is all useful and should go in.

Thoughts on the questions:
1. yes, I think the start/end tuple is a little awkward, and kwargs would be better, so that a function can be explicit about which of these it intends to care about. Periodic could even take extra kwargs itself, which get passed to the data generator function in addition to the timings.
2. No strong feeling on the name, but it would be worthwhile to actually parse frequency ("12Hz") if we call it so; a plausible alternative name here would be "cadence". 
3. Agreed, and indeed it is not experimental. I think the event loop/async ideas were new at the time.
4. I'd be tempted to drop explicit remarks about tornado. IOLoop just uses asyncio internally anyway, and supports async/await syntax. In fact, it would make sense, stylistically, for the data generator function here to be async, but I don't think it matters.

I made flake8 fixes for you.

martindurant:
It seems that `random_datapoint` is not covered by tests

streamz/dataframe/core.py
jbednar: Ok, I'll reword that, but note that it strictly adds nothing new to PeriodicDataFrame besides the name Random. :-)  It will also retain the `freq` parameter that I think I'll rename to `cadence` in PeriodicDataFrame.

martindurant: OK

jbednar:
Thanks for the flake fixes! From your replies:

1. Ok, I'll rewrite it using kwargs, and that's a good idea to support extra kwargs for configuring the callback.
2. Formulating and parsing as Hz seems tricky, so I'll rename to `cadence`. Thanks for the suggestion!
3. Ok, it will now be properly documented and no longer marked experimental.
4. PeriodicDataFrame explicitly depends on `tornado.gen` in all cases and on `tornado.ioloop.IOLoop` when not using Dask (for Dask it uses `distributed.client.default_client().loop` instead, which is probably also tornado based?).  I think that using the IOLoop from Tornado as the default event loop is important for keeping the mental model simple for Jupyter users. Maybe the loop itself could be parameterized, but I don't personally know how to use anything _other_ than Tornado.  If you or anyone else could suggest how to parameterize the loop or have non-Tornado examples to follow, that would be great!  For now I think it does need to indicate that the implementation is specific to Tornado.

martindurant:
4. They are the same these days, although I'm sure Tornado can be configured to use non-asyncio loops. Does the jupyter model explicitly talk about Tornado? I'm not sure that it matters.
```python
import asyncio
import tornado.ioloop
loop = tornado.ioloop.IOLoop.current()
assert asyncio.get_event_loop() is loop.asyncio_loop
```

jbednar:
Cool! I'll try replacing the references to Tornado with asyncio; perfect!

jbednar:
@martindurant. as best I can see streamz is only tested against Python 3.7. Does that mean I can ignore Python 2 and 3.6?  

Also, it looks like the name `freq` was inherited from [pd.date_range](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.date_range.html) where it gets used. Pandas has a lot more detail about how to use it than streamz will, so I think it's probably best to keep it with a disclaimer rather than rename it.

martindurant:
2.7 we certainly don't care about any more. I am unsure about 3.6, lots of projects still release/test for it, but it seems to be fading. Probably we should have tests against 3.8 at least. cc @chinmaychandak

jbednar:
Using the same numbering as above, 

1. This PR has now been rewritten using kwargs, both for the callback itself and accepting optional ones in the constructor.  As a bonus, with this approach `freq` no longer needed any special handling in PeriodicDataFrame; it was always just a constant that was passed into the callback, so now it's treated like any other kwarg, provided by Random and consumed by `random_datablock` but never explicitly handled by PeriodicDataFrame. I think this illustrates the utility of the kwarg approach, which easily handles configuration like this and avoids having to explain `freq` at the PeriodicDataFrame level (where it doesn't have any meaning).
2. The `freq` parameter thus stays named `freq`, but now it's really only used for `random_datablock`. `freq` is specific to the case of generating values on a timebase, and isn't broadly applicable to the types of callbacks accepted by PeriodicDataFrame.
3. Random is now properly documented and no longer marked experimental.
4. I haven't yet looked at asyncio (or tests).

docs/source/dataframes.rst
```diff
@@ -178,5 +178,78 @@ and ``DaskStream`` objects.
...
+``df`` will now be a steady stream of whatever values are returned by
+the `datafn`, which can of course be any Python code as long as it
+returns a DataFrame. 
+
+Here we returned only a single point, appropriate for streaming the
+results of system calls or other isolated actions, but any number of
+entries can be returned by the dataframe in a single batch. To
+facilitate collecting such batches, the callback is invoked with
+``last`` (the time of the previous invocation) and ``now`` (the
+time of the current invocation. The callback can then generate or
```
jbednar, line 224:
```suggestion
time of the current invocation). The callback can then generate or
```

jbednar:
Also replaced use of system time() function and the associated 1e9 conversion factors with Pandas Timestamp. As this class is explicitly based on Pandas and requires Pandas in the callbacks, it seems cleaner and more consistent to use Pandas times throughout.  Happy to revert that last commit if that's not appropriate.

docs/source/dataframes.rst
```diff
@@ -178,5 +178,79 @@ and ``DaskStream`` objects.
...
+results of system calls or other isolated actions, but any number of
+entries can be returned by the dataframe in a single batch. To
+facilitate collecting such batches, the callback is invoked with
+keyword arguments ``last`` (the time of the previous invocation) and
+``now`` (the time of the current invocation) as Pandas Timestamp
+objects. The callback can then generate or query for just the values
+in that time range.
+
+Arbitrary keyword arguments can be provided to the PeriodicDataFrame
+constructor, which will be passed into the callback that its behavior
```
jbednar, line 229:
```suggestion
constructor, which will be passed into the callback so that its behavior
```

martindurant:
Thanks @jbednar

PR #365, opened 2020-09-30
Directly alter notebook cell output

### LINKED ISSUE #341: Output cells not updating in jupyter lab
So I just discovered [Jupyterlab cell output issue](https://github.com/python-streamz/streamz/issues/307), and was relieved to see I'm not going insane. I am very reluctantly switching my entire project to a notebook, but hopefully someone who understands what's going on can fix it for jupyter lab. Meanwhile, it would be __really great__ if someone who's responsible for things like this would put a notice in big cheerful letters on the front cover saying __"DO NOT TRY THIS IN JUPYTER LAB!"__ :)

I copied the streamz example from @mrocklin's slides at [streamz-walkthrough-to-dataframes.ipynb](https://gist.github.com/mrocklin/893e693ed19a357ea38962dcf112f616). When I get to cell 11:

`from tornado import gen

async def f():
    while True:
        await gen.sleep(0.100)
        record = create_record()
        await source.emit(record, asynchronous=True)
        
IOLoop.current().add_callback(f)`

the output cells update once and are then silent. If I set the log level to "Info," i can see that the function is running continuously in the background and generating data, but this is not reflected in the notebook cells. I have tried this using tornado and threading on Windows 10 and in Ubuntu under WSL, using streamz_dev.yml in each case. Under Windows 10, the cells will occasionally start updating if I run a simple command like threading.enumerate() multiple times, but if I then run a new cell to access the data, visible updating stops again. I have not been able to duplicate this behavior in WSL; there is simply no visible updating, even though it's evident from the log that the function really is running in the background. I also tried putting a progress bar widget in the same function, and it displays and updates perfectly even though there's still no visible data from the stream.

Has anybody gotten this to work on windows or WSL? Has anybody recently run it successfully on a linux machine? What am I doing wrong here?

martindurant:
Fixes #341

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -350,8 +350,7 @@ class Stream(object):
             if output is None:
                 return
             with output:
-                content, *_ = InteractiveShell.instance(
-                    ).display_formatter.format(val)
+                content, *_ = InteractiveShell.instance().display_formatter.format(val)
                 output.outputs = ({'output_type': 'display_data',
                       'data': content,
                       'metadata': {}},)
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -333,7 +333,7 @@ class Stream(object):
 
     __repr__ = __str__
 
-    def _ipython_display_(self, **kwargs):  # pragma no cover
+    def _ipython_display_(self, **kwargs):  # pragma: no cover
         try:
             from ipywidgets import Output
             from IPython.core.interactiveshell import InteractiveShell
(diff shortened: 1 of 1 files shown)


martindurant:
Note: it would be reasonable to have the output managed by Panel, instead of, or as an alternative to, IPy display. Note that this doesn't actually need ipywidgets, as far as I can tell.

philippjfr:
If we wanted to make the output mechanism configurable you could also use `pn.pane.Streamz()` to get automatically updating output using `Panel`/`Bokeh`.

martindurant:
Thanks for the nudge, @philippjfr . A pane that displays (almost) anything that streamz throws into it sounds perfect.

PR #366, opened 2020-10-16
Add refresh_cycles parameter to from_kafka_batched to accommodate adding Kafka topic partitions on the fly.

chinmaychandak:
Streamz should be able to handle addition of Kafka topic partitions on the fly. The `refresh_cycles` parameter polls Kafka every `N` cycles to get the latest Kafka partition metadata and accommodates any added partitions on the fly. 

[Some more discussion in #358]

### DIFF
--- a/streamz/tests/test_kafka.py
+++ b/streamz/tests/test_kafka.py
@@ -290,7 +290,7 @@ def test_kafka_batch_npartitions():
         stream3.upstream.stopped = True
 
 
-def test_kafka_refresh_cycles():
+def test_kafka_check_npartitions_every():
     j1 = random.randint(0, 10000)
     ARGS = {'bootstrap.servers': 'localhost:9092',
             'group.id': 'streamz-test%i' % j1,
@@ -315,7 +315,7 @@ def test_kafka_refresh_cycles():
 
         stream = Stream.from_kafka_batched(TOPIC, ARGS,
                                            asynchronous=True,
-                                           refresh_cycles=1,
+                                           check_npartitions_every=1,
                                            poll_interval='2s')
         out = stream.gather().sink_to_list()
         stream.start()
--- a/streamz/tests/test_kafka.py
+++ b/streamz/tests/test_kafka.py
@@ -290,7 +290,7 @@ def test_kafka_batch_npartitions():
         stream3.upstream.stopped = True
 
 
-def test_kafka_check_npartitions_every():
+def test_kafka_refresh_partitions():
     j1 = random.randint(0, 10000)
     ARGS = {'bootstrap.servers': 'localhost:9092',
             'group.id': 'streamz-test%i' % j1,
@@ -315,7 +315,7 @@ def test_kafka_check_npartitions_every():
 
         stream = Stream.from_kafka_batched(TOPIC, ARGS,
                                            asynchronous=True,
-                                           check_npartitions_every=1,
+                                           refresh_partitions=True,
                                            poll_interval='2s')
         out = stream.gather().sink_to_list()
         stream.start()
--- a/streamz/tests/test_kafka.py
+++ b/streamz/tests/test_kafka.py
@@ -336,7 +336,8 @@ def test_kafka_refresh_cycles():
         kafka.flush()
         time.sleep(5)
 
-        assert (len(out) == 4 and (len(out[2]) + len(out[3])) == 10)
+        assert (len(out) == 4 and (len(out[2]) + len(out[3])) == 10
+                and out[3][4] == b'value-19')
         stream.upstream.stopped = True
 
 
--- a/streamz/tests/test_kafka.py
+++ b/streamz/tests/test_kafka.py
@@ -87,7 +87,8 @@ def kafka_service():
                 "Kafka not available. "
                 "To launch kafka use `export STREAMZ_LAUNCH_KAFKA=true`")
 
-        producer = ck.Producer({'bootstrap.servers': 'localhost:9092'})
+        producer = ck.Producer({'bootstrap.servers': 'localhost:9092',
+                                'topic.metadata.refresh.interval.ms': '5000'})
         producer.produce('test-start-kafka', b'test')
         out = producer.flush(10)
         if out > 0:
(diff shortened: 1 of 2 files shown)


chinmaychandak:
@martindurant Do review when you get time. 

Would like your opinion on:
1. The name `refresh_cycles` parameter. Maybe `kafka_metadata_refresh_cycles` is a better name, but too long?
2. There are too many `if self.refresh_cycles is not None` clauses, maybe there's a cleaner way to go about this?
2. Is the test case good enough to test the functionality?

martindurant:
I think the design is fundamentally fine, and I am not worried about too many `if is None` statements. It could be more compact, but probably not worth doing this.

```python
if self.refresh_cycles:
    if cycles % self.refresh_cycles == 0:
        ...
    cycles += 1
```
or even
```python
cycles = self.refresh_cycles

if self.refresh_cycles:
    if cycles < 1:
        ...
        cycles = self.refresh_cycles
    cycles -= 1
```

My main quibble is the name "refresh_cycles" in the signature - what it actually does is check the number of partitions every N polls. So why not call it something more specific like "check_npartitions_every". Within the function, `cycles` is fine.

martindurant:
Is there a cost to having more partitions than the topic really has? I wonder whether we want to, perhaps at a future date, allow the number of partitions to decrease too.

streamz/tests/test_kafka.py
```diff
@@ -87,7 +87,8 @@ def kafka_service():
                 "Kafka not available. "
                 "To launch kafka use `export STREAMZ_LAUNCH_KAFKA=true`")
 
-        producer = ck.Producer({'bootstrap.servers': 'localhost:9092'})
+        producer = ck.Producer({'bootstrap.servers': 'localhost:9092',
+                                'topic.metadata.refresh.interval.ms': '5000'})
```
martindurant, line 91: Is this required? It's not in the docstrings that you need to set it.

chinmaychandak: This is required in the producer configs for the producer to start producing to the newly added partitions sooner than the default topic metadata refresh interval (which is 5 minutes). This config is mainly so that the test doesn't need to wait for more than a few seconds for the partitions metadata to refresh.

jsmaupin:
There's no way to make this work so that it happens automatically by the way the library talks with Kafka? It may be confusing to add this extra parameter. It's not something I've seen when working with other Kafka clients.

chinmaychandak:
> Is there a cost to having more partitions than the topic really has? 

No, I'm not aware of any cost associated. In fact, it is recommended to add more partitions to balance any increase in load in the Kafka topic and scaling out improving performance.  

> I wonder whether we want to, perhaps at a future date, allow the number of partitions to decrease too.

I don't think Kafka supports or recommends decreasing number of partitions, because that would lead to data loss. It would be best in this case to just restart the stream.

chinmaychandak:
> There's no way to make this work so that it happens automatically by the way the library talks with Kafka? It may be confusing to add this extra parameter. It's not something I've seen when working with other Kafka clients.

We already talked a little about this in #358. An option is to just refresh the partition count every N cycles by default (so the exact same logic of this PR without exposing it via an additional parameter), without providing an option to the user to configure what N would be. We don't need N to be 1 necessarily, since a concern was that not all users might want this extra call or functionality to be executed in each poll. Also, some users have a really small poll interval, so having this in every poll cycle might not be worth it. But Martin thought it would be best to allow the user to configure this functionality somehow.

martindurant:
@jsmaupin , by "extra parameter" in your comment, were you referring to the kwarg (currently `refresh_cycles`), or the extra config you need to pass to the kafka setup?

chinmaychandak:
> were you referring to the kwarg (currently `refresh_cycles`), or the extra config you need to pass to the kafka setup?

@martindurant, we don't need to pass any extra config in the Kafka setup. I think Jarod was referring to `refresh_cycles` (now `check_npartitions_every` :))

jsmaupin:
I'm referring to the `refresh_cycles` parameter.

martindurant:
@chinmaychandak , are you able to do some kind of benchmark to see the effect of polling for the number of topic partitions on every cycle?

chinmaychandak:
> @chinmaychandak , are you able to do some kind of benchmark to see the effect of polling for the number of topic partitions on every cycle?

Okay, I can try some benchmarking. Do you have something specific in mind?

Although, this is what I've seen with a few things I've tried out:

The size of the metadata returned _for a specific topic_ (this is important, else the metadata for the entire Kafka cluster can be really large) by `list_topics(streamz-topic)` is `64 bytes` (as returned by `sys.getsizeof`). This is smaller than a single typical average Kafka message (anywhere b/w 100 bytes to a few KBs). This also means that the latency of retrieving this `metadata`, once every poll cycle, would be much lesser than that for retrieving a single Kafka message (and <<< than the latency of reading an entire batch of messages from multiple partitions as intended by `from_kafka_batched`). 

I ran a small test, and the time taken for `list_topics(streamz-topic)` is an average of `25ms` (across 100 cycles) for a topic with 8 partitions, and if I increase the number of partitions to 16, the time still remains the same. This is really minimal compared to the time taken to read an entire batch of messages (which can easily be in order of 10s of seconds). 

But, as I said before, if the poll interval is ridiculously small (although I can't think why someone would want it to be less than at least 2-5 seconds, otherwise they're effectively not taking advantage of batching), calling `list_topics` on every cycle can be a problem.

martindurant:
We clearly have *very* different idea of the use case, as we found in our last discussion about this :)

Although the high-throughput, big-batch use case may be very important, in which any latency and overhead from streamz (and dask) are eaten by the downloading of kafka and further processing. That will be the case indeed when you have poll times >1s.

However, streamz is developed to be a real-time processing system. It is perfectly reasonably to have kafka polling many times a second, perhaps with only a few or even no events per batch. Perhaps the most important thing is to have output as quickly after an event appears as possible - but still need batch to deal with spiky traffic.

> size of the metadata returned for a specific topic by list_topics(streamz-topic) is 64 bytes

This is probably unimportant: the data transfer is likely a very small compared to the connection time. It may well be that listing partitions, and polling offsets amounts to double the latency compared to polling offsets alone.

> the time taken for list_topics(streamz-topic) is an average of 25ms

On a local cluster, the typical time to submit, execute and return a value from dask is ~5ms; submit alone is <<1ms and typical task overheads are <<1ms. The per-node latency in streamz is ... tiny (really should benchmark this! perhaps the new metadata stream could include time-of-execution...). So whether or not 25ms seems like a lot depend on the use case. How long does getting offsets take? Is this with kafka on a separate container within the same host?

martindurant:
By the way, since streamz is an async application (on tornado), https://www.confluent.io/blog/kafka-python-asyncio-integration/ might be interesting, as well as maybe aiokafka (which seems unmaintained). I don't know to what extent multiple kafka calls could be made concurrently, but if that works, maybe the various latencies can be amortized.

chinmaychandak:
Okay, I understand the use case you're talking about now. I did feel that calling `list_topics` on every cycle could be problematic when the poll interval is really small.

> On a local cluster, the typical time to submit, execute and return a value from dask is ~5ms; submit alone is <<1ms and typical task overheads are <<1ms.

What kind of executions/processing are we talking about here? 

> Is this with kafka on a separate container within the same host?

`25ms` was with Kafka on AWS, and the consumer on my local machine. With Kafka within the same host, the latency seems to be `<<1ms`, and within the same VPC but from a separate machine, the latency seems to be `~1ms`. 

> How long does getting offsets take?

Getting a single message seems to take `<<<1ms`.

I feel that considering the numbers above ^ it should be safe to call `list_topics` on every cycle, although I'm really okay with keeping the PR as is, if you feel that would be more comfortable with all kinds of use cases.

jsmaupin:
Maybe this was already discussed and I missed it, but I understand that the Kafka consumer generally has a "re-balance trigger" where a trigger is activated when something happens like a new partition is added. There's no way we can tie into this triggering mechanism to schedule a metadata re-download?

martindurant:
> What kind of executions/processing are we talking about here?

The simplest of functions, i.e., this is all overhead/latency. Obviously, real workloads take longer, 
but the blocking part from the client's point of view should not change much.

> Getting a single message seems to take <<<1ms

I don't mean consuming a message, but polling the offsets, which is what the streamz node spends its time on.

Note that the streamz event loop may be watching several inputs, e.g., from sockets or files, so having the kafka callback *fast* is important.

chinmaychandak:
> I don't mean consuming a message, but polling the offsets, which is what the streamz node spends its time on.

I confused this with the `poll` method which consumes the actual messages. 

Okay, so without `list_topics` on every cycle, the avg. time to poll offsets for 100 (or 10000) cycles is around `21ms`. With `list_topics` on every cycle, it is around `23ms`.

chinmaychandak:
> Maybe this was already discussed and I missed it, but I understand that the Kafka consumer generally has a "re-balance trigger" where a trigger is activated when something happens like a new partition is added. There's no way we can tie into this triggering mechanism to schedule a metadata re-download?

I can't seem to find a trigger config anywhere. There is the `topic.metadata.refresh.interval.ms` config which I've used in the test added in this PR and it defaults to `5mins`.

martindurant:
> around 21ms. With list_topics on every cycle, it is around 23ms

Oh! That would suggest not bothering with the parameter, and only providing yes/no to checking. I am surprised. However, I am also surprised that 21ms is as high. Seems I don't understand the technicalities of the TCP connection that ck keeps open. :|

chinmaychandak:
> That would suggest not bothering with the parameter, and only providing yes/no to checking.

So are you suggesting a boolean parameter (`refresh_partitions`? :)) instead defaulting to `True`?

chinmaychandak:
> However, I am also surprised that 21ms is as high.

Yes, I was surprised too. Profiling at a little more granularity, `get_watermark_offsets` is taking about `2.5ms` per call.

martindurant:
Right, so the actual latency for a call to kafka is in the 2ms range - so checking for partitions is still significant in the case of few starting partitions.

> So are you suggesting a boolean parameter (refresh_partitions? :)) instead defaulting to True?

I would have said `False` :). The only reason would be, that this is simpler to understand; but given that the code is already there, this is fine too, but the docstring needs to be precise. I think the current version will do, but may need some clarification later (perhaps in examples).

chinmaychandak:
Done!

chinmaychandak:
@martindurant, is this good to go? Just one more PR in the pipeline after this. :)

chinmaychandak:
@martindurant can you please help cut a release sometime this week? Would love to use the fixes/features of #358 and this PR in our existing workloads. :)

martindurant:
Remind me next week

chinmaychandak:
@martindurant, bump ^

martindurant:
On pypi

PR #367, opened 2020-10-21
Fix checkpointing bug in auto.offset.reset=latest with multiple topic partitions.

chinmaychandak:
Bug was introduced in #358.  

Since `positions` is [modified](https://github.com/python-streamz/streamz/blob/8a4888aa5884912e47adc445019a65ede8344e99/streamz/sources.py#L538) on every iteration, the earlier condition fails to do the intended check when there are multiple partitions.

### DIFF
--- a/streamz/tests/test_kafka.py
+++ b/streamz/tests/test_kafka.py
@@ -290,7 +290,7 @@ def test_kafka_batch_npartitions():
         stream3.upstream.stopped = True
 
 
-def test_kafka_refresh_cycles():
+def test_kafka_check_npartitions_every():
     j1 = random.randint(0, 10000)
     ARGS = {'bootstrap.servers': 'localhost:9092',
             'group.id': 'streamz-test%i' % j1,
@@ -315,7 +315,7 @@ def test_kafka_refresh_cycles():
 
         stream = Stream.from_kafka_batched(TOPIC, ARGS,
                                            asynchronous=True,
-                                           refresh_cycles=1,
+                                           check_npartitions_every=1,
                                            poll_interval='2s')
         out = stream.gather().sink_to_list()
         stream.start()
--- a/streamz/tests/test_kafka.py
+++ b/streamz/tests/test_kafka.py
@@ -290,7 +290,7 @@ def test_kafka_batch_npartitions():
         stream3.upstream.stopped = True
 
 
-def test_kafka_check_npartitions_every():
+def test_kafka_refresh_partitions():
     j1 = random.randint(0, 10000)
     ARGS = {'bootstrap.servers': 'localhost:9092',
             'group.id': 'streamz-test%i' % j1,
@@ -315,7 +315,7 @@ def test_kafka_check_npartitions_every():
 
         stream = Stream.from_kafka_batched(TOPIC, ARGS,
                                            asynchronous=True,
-                                           check_npartitions_every=1,
+                                           refresh_partitions=True,
                                            poll_interval='2s')
         out = stream.gather().sink_to_list()
         stream.start()
--- a/streamz/tests/test_kafka.py
+++ b/streamz/tests/test_kafka.py
@@ -553,3 +553,75 @@ def test_kafka_batch_checkpointing_async_nodes_2():
         assert committed2[1].offset == 2
         assert committed3[0].offset == 1
         assert committed3[1].offset == 1
+
+
+def test_kafka_checkpointing_auto_offset_reset_latest():
+    '''
+    Testing whether checkpointing works as expected with multiple topic partitions and
+    with auto.offset.reset configuration set to latest (also default).
+    '''
+    j = random.randint(0, 10000)
+    ARGS = {'bootstrap.servers': 'localhost:9092',
+            'group.id': 'streamz-test%i' % j,
+            'auto.offset.reset': 'latest'}
+    with kafka_service() as kafka:
+        kafka, TOPIC = kafka
+        TOPIC = "test-checkpointing-offset-reset-latest"
+        subprocess.call(shlex.split("docker exec streamz-kafka "
+                                    "/opt/kafka_2.11-0.10.1.0/bin/kafka-topics.sh "
+                                    "--create --zookeeper localhost:2181 "
+                                    "--replication-factor 1 --partitions 3 "
+                                    "--topic test-checkpointing-offset-reset-latest"))
+        time.sleep(2)
+
+        '''
+        Since the stream has not started yet, these messages are not read because
+        the stream has auto.offset.reset set to latest.
+        '''
+        for i in range(30):
+            kafka.produce(TOPIC, b'value-%d' % i)
+        kafka.flush()
+
+        stream1 = Stream.from_kafka_batched(TOPIC, ARGS, asynchronous=True)
+        out1 = stream1.map(split).gather().sink_to_list()
+        stream1.start()
+        time.sleep(5)
+
+        '''
+        Stream has started, so these are read.
+        '''
+        for i in range(30):
+            kafka.produce(TOPIC, b'value-%d' % i)
+        kafka.flush()
+        time.sleep(5)
+
+        assert (len(out1) == 3 and (len(out1[0]) + len(out1[1]) + len(out1[2])) == 30)
+        '''
+        Stream stops but checkpoint has been created.
+        '''
+        stream1.upstream.stopped = True
+
+        '''
+        When the stream is restarted, these messages are read, because the checkpoint
+        overrrides the auto.offset.reset:latest config this time around as expected.
+        '''
+        for i in range(30):
+            kafka.produce(TOPIC, b'value-%d' % i)
+        kafka.flush()
+
+        stream2 = Stream.from_kafka_batched(TOPIC, ARGS, asynchronous=True)
+        out2 = stream2.map(split).gather().sink_to_list()
+
+        '''
+        Stream restarts here.
+        '''
+        stream2.start()
+        time.sleep(5)
+
+        for i in range(30):
+            kafka.produce(TOPIC, b'value-%d' % i)
+        kafka.flush()
+        time.sleep(5)
+
+        assert (len(out2) == 6 and (len(out2[3]) + len(out2[4]) + len(out2[5])) == 30)
+        stream2.upstream.stopped = True
--- a/streamz/tests/test_kafka.py
+++ b/streamz/tests/test_kafka.py
@@ -255,7 +255,6 @@ def test_kafka_batch_npartitions():
                                     "--create --zookeeper localhost:2181 "
                                     "--replication-factor 1 --partitions 2 "
                                     "--topic test-partitions"))
-        time.sleep(5)
 
         for i in range(10):
             if i % 2 == 0:
@@ -304,7 +303,6 @@ def test_kafka_refresh_partitions():
                                     "--create --zookeeper localhost:2181 "
                                     "--replication-factor 1 --partitions 2 "
                                     "--topic test-refresh-partitions"))
-        time.sleep(2)
 
         for i in range(10):
             if i % 2 == 0:
@@ -327,7 +325,6 @@ def test_kafka_refresh_partitions():
                                     "--alter --zookeeper localhost:2181 "
                                     "--topic test-refresh-partitions --partitions 4"))
         time.sleep(5)
-
         for i in range(10,20):
             if i % 2 == 0:
                 kafka.produce(TOPIC, b'value-%d' % i, partition=2)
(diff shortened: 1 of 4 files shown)


chinmaychandak:
Blocked by this Dask issue it seems: https://github.com/dask/distributed/issues/4177.

martindurant:
Is it possible to construct/edit a test which failed due to this bug, and gets fixed here?

chinmaychandak:
Yes, was planning on adding a test today. Just added it.

On a side note, if I try to run this test as well as some other tests like `test_kafka_refresh_partitions` or `test_kafka_batch_npartitions` **on Dask** via the `@gen_cluster(client=True, timeout=60)` marker, they seem to fail on my local machine consistently (not flaky) because somehow the DaskStream doesn't seem to read any data. But running similar tests on an AWS instance does not encounter any issue, which is weird. 

@martindurant, can you try running any of these tests on Dask on your end and let me know if you are able to see any issues? Maybe there's something in my environment that needs fixing.

streamz/tests/test_kafka.py
```diff
@@ -553,3 +553,75 @@ def test_kafka_batch_checkpointing_async_nodes_2():
...
+        the stream has auto.offset.reset set to latest.
+        '''
+        for i in range(30):
+            kafka.produce(TOPIC, b'value-%d' % i)
+        kafka.flush()
+
+        stream1 = Stream.from_kafka_batched(TOPIC, ARGS, asynchronous=True)
+        out1 = stream1.map(split).gather().sink_to_list()
+        stream1.start()
+        time.sleep(5)
```
martindurant, line 588: It should not be necessary to put all these `sleep` statements in here. If necessary, we could have wait_for loops that poll for successful starting.

martindurant:
> can you try running any of these tests on Dask

what exactly would you like me to run?

streamz/tests/test_kafka.py
chinmaychandak: What predicate do you suggest using in the `wait_for` in this case?

chinmaychandak:
> what exactly would you like me to run?

I modified these two tests to use Dask. Can you try running them? These two test fail for with assertions stating that no data has been read at all.

```
@gen_cluster(client=True, timeout=60)
def test_kafka_batch_npartitions(c, s, w1, w2):
    j1 = random.randint(0, 10000)
    ARGS1 = {'bootstrap.servers': 'localhost:9092',
             'group.id': 'streamz-test%i' % j1,
             'enable.auto.commit': False,
             'auto.offset.reset': 'earliest'}
    j2 = j1 + 1
    ARGS2 = {'bootstrap.servers': 'localhost:9092',
             'group.id': 'streamz-test%i' % j2,
             'enable.auto.commit': False,
             'auto.offset.reset': 'earliest'}
        kafka, TOPIC = kafka

        TOPIC = "test-partitions"
        subprocess.call(shlex.split("docker exec streamz-kafka "
                                    "/opt/kafka_2.11-0.10.1.0/bin/kafka-topics.sh "
                                    "--create --zookeeper localhost:2181 "
                                    "--replication-factor 1 --partitions 2 "
                                    "--topic test-partitions"))
        time.sleep(5)

        for i in range(10):
            if i % 2 == 0:
                kafka.produce(TOPIC, b'value-%d' % i, partition=0)
            else:
                kafka.produce(TOPIC, b'value-%d' % i, partition=1)
        kafka.flush()

        with pytest.raises(ValueError):
            stream1 = Stream.from_kafka_batched(TOPIC, ARGS1,
                                                asynchronous=True,
                                                npartitions=0, dask=True)
            stream1.gather().sink_to_list()
            stream1.start()

        stream2 = Stream.from_kafka_batched(TOPIC, ARGS1,
                                            asynchronous=True,
                                            npartitions=1, dask=True)
        out2 = stream2.gather().sink_to_list()
        stream2.start()
        time.sleep(5)
        assert (len(out2) == 1 and len(out2[0]) == 5)
        stream2.upstream.stopped = True

        stream3 = Stream.from_kafka_batched(TOPIC, ARGS2,
                                            asynchronous=True,
                                            npartitions=4, dask=True)
        out3 = stream3.gather().sink_to_list()
        stream3.start()
        time.sleep(5)
        assert (len(out3) == 2 and (len(out3[0]) + len(out3[1])) == 10)
        stream3.upstream.stopped = True


@gen_cluster(client=True, timeout=60)
def test_kafka_refresh_partitions(c, s, w1, w2):
    j1 = random.randint(0, 10000)
    ARGS = {'bootstrap.servers': 'localhost:9092',
            'group.id': 'streamz-test%i' % j1,
            'enable.auto.commit': False,
            'auto.offset.reset': 'earliest'}
        kafka, TOPIC = kafka
        TOPIC = "test-refresh-partitions"
        subprocess.call(shlex.split("docker exec streamz-kafka "
                                    "/opt/kafka_2.11-0.10.1.0/bin/kafka-topics.sh "
                                    "--create --zookeeper localhost:2181 "
                                    "--replication-factor 1 --partitions 2 "
                                    "--topic test-refresh-partitions"))
        time.sleep(2)

        for i in range(10):
            if i % 2 == 0:
                kafka.produce(TOPIC, b'value-%d' % i, partition=0)
            else:
                kafka.produce(TOPIC, b'value-%d' % i, partition=1)
        kafka.flush()

        stream = Stream.from_kafka_batched(TOPIC, ARGS,
                                           asynchronous=True,
                                           refresh_partitions=True,
                                           poll_interval='2s', dask=True)
        out = stream.gather().sink_to_list()
        stream.start()
        time.sleep(5)
        assert (len(out) == 2 and (len(out[0]) + len(out[1])) == 10)

        subprocess.call(shlex.split("docker exec streamz-kafka "
                                    "/opt/kafka_2.11-0.10.1.0/bin/kafka-topics.sh "
                                    "--alter --zookeeper localhost:2181 "
                                    "--topic test-refresh-partitions --partitions 4"))
        time.sleep(5)

        for i in range(10,20):
            if i % 2 == 0:
                kafka.produce(TOPIC, b'value-%d' % i, partition=2)
            else:
                kafka.produce(TOPIC, b'value-%d' % i, partition=3)
        kafka.flush()
        time.sleep(5)

        assert (len(out) == 4 and (len(out[2]) + len(out[3])) == 10
                and out[3][4] == b'value-19')
        stream.upstream.stopped = True
```

martindurant:
Sleep and async should normally work very badly together, since it will block the event loop; whether or not that crashes dask will depend on a number of things, including the select implementation and various timeouts - maybe that is what you are seeing. There is an async await_for (or some such name) that should be used, to make sure both Dask and our stream is up and ready.

chinmaychandak:
I see, okay, that makes sense. That also might explain why using `yield gen.sleep()` seems to help. Will try out a few more things.

streamz/tests/test_kafka.py
chinmaychandak: Can't think of anything else here other than sleep to ensure that the messages aren't produced to Kafka before the stream successfully starts.

chinmaychandak: What do you think? Is it okay to leave this as is? I see a couple of other tests using sleep too.

martindurant: Is there a way to poll with the producer to see if the broker is ready to accept messages?

martindurant: Or the consumer, I suppose: as soon as list_topics doesn't fail, and includes the topic we're after, we should be ready to start. This kind of logic can live in the kafka fixture too, or a utility function.

Note that the stream's `start` method is blocking, we shouldn't need to wait. It adds the poll_kafka callback, so we could asyncio.sleep(0) to make sure that runs - evrything in it is blocking until the poll is complete, except for possible checkpointing, which will have nothing to do at this point.

chinmaychandak: > Is there a way to poll with the producer to see if the broker is ready to accept messages?

Not that I know of. I'm assuming you're referring to sleep on line 575. Typically, the creation of a topic is blocking, so I'll remove this sleep. 

> Note that the stream's start method is blocking, we shouldn't need to wait. It adds the poll_kafka callback, so we could asyncio.sleep(0) to make sure that runs - evrything in it is blocking until the poll is complete, except for possible checkpointing, which will have nothing to do at this point.

The point is to ensure that the [first polling cycle](https://github.com/python-streamz/streamz/blob/7f92174ae4c21a60105e1b8476ae3936c5f1c6d3/streamz/sources.py#L514), which includes the execution of lines 530 and the branch on line 534, should finish before we produce any new messages to Kafka. Removing sleep is causing the producer to produce before the stream can finish the first poll cycle. Note that this is just so that we can simulate a good scenario to test the functionality where auto.offset.reset is latest. Let me know if this makes sense.

martindurant: so a single asyncio/gen-sleep of zero time should do, as start already queues the polling coroutine, and the first call of it is guaranteed to get that far.

chinmaychandak: Nope, using `gen.sleep(0 or 5)` doesn't seem to work, the producer still produces before the first cycle. I think we need a blocking call to ensure the stream completes the first poll cycle.

chinmaychandak: This maybe due to the `list_topics`, `get_watermark_offsets` being relatively expensive calls.

martindurant: It would be really good to figure out exactly why, and this would go a great distance to fixing the various flaky/unstable kafka pieces. I know it's complicated: streamz's event loop schedules when the coroutine runs, but rdkafka is actually in a different (non-python) thread.

Perhaps the source class could have a `running` instance attribute which is initially `False`, but it gets set the first time that `get_watermark_offsets` succeeds (i.e., after the `except` block) - or let `stopped` have more than `True/False` values.

martindurant: ^ this, so that we have something whose status we can (a)wait on.

chinmaychandak: > this would go a great distance to fixing the various flaky/unstable kafka pieces

I agree.

> Perhaps the source class could have a running instance attribute which is initially False, but it gets set the first time that get_watermark_offsets succeeds (i.e., after the except block) - or let stopped have more than True/False values.

This is definitely a way out, although I'm not sure if we ever need it except for simulating tests with `auto.offset.reset:latest` Kafka config. I could simplify the tests a little, but I was thinking the tests should cover the config exhaustively (at the risk of some sleep statements, for now), especially after I introduced this nasty bug. 

But I agree, it would definitely be worth spending some time figuring our the flaky nature of some of the tests.

martindurant: The trouble is, as we have found, that `sleep` doesn't performs the same on CI as locally, doubtless because of the puny workers that the tests run on. The totality of a `sleep` may sometimes be eaten up by garbage collection or perhaps some other build on the same host. I addition, having a test that takes a minimum of tens of seconds to run is unfortunate.

I do agree with the desire to test all configs.

jsmaupin: Tests should always be predictable so you don't get intermittent failures. You should not use `random` unless it is used to pre-generate test data and using `sleep` is more or less just a patch. The best solution is to figure out how to watch or listen to the event you're waiting for while sleeping, like what the `wait_for` function is used for.

chinmaychandak: @martindurant, would we need to make some changes to the `client.submit` calls in streamz to make sure it works with the Dask [changes](https://github.com/dask/distributed/issues/4177)?

martindurant: Indeed maybe so, but this should be a follow-up. It seems we may need to convert some tuples to lists, but we can wait until things start to fail, I think.

chinmaychandak: I think CI was already failing in the last couple of builds:

https://travis-ci.org/github/python-streamz/streamz/builds/737860381
https://travis-ci.org/github/python-streamz/streamz/builds/738116645

martindurant: I believe the immediate change will be reverted, but let's see what decision is made on that PR. If you have an opinion, can voice it there.

chinmaychandak: Ah, I see that temporary fix just got merged. Hopefully we can trigger a build after some time to see if CI passes here?

chinmaychandak: Maybe the fix isn't in yet? CI is failing with this error: `ImportError: cannot import name 'Layer' from 'dask.highlevelgraph'`.

chinmaychandak: @martindurant, any ideas?

PR #371, opened 2020-11-03
Fix code blocks in quickstart list

remram44:
See problem: https://streamz.readthedocs.io/en/latest/#quickstart

screenshot of current readthedocs

Fixed:

screenshot of fixed version

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -84,14 +84,18 @@ environment but rather for experimentation, learning, or new feature development
 Its most common use would be to interact with the streamz example jupyter notebooks. Lets walk through the steps needed for this.
 
 - Build the Docker container
-.. code-block:: bash
-$ docker/build.sh
-- Run the Docker container
-.. code-block:: bash
-$ docker/run.sh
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -84,14 +84,18 @@ environment but rather for experimentation, learning, or new feature development
 Its most common use would be to interact with the streamz example jupyter notebooks. Lets walk through the steps needed for this.
 
 - Build the Docker container
-.. code-block:: bash
-$ docker/build.sh
+
+  .. code-block:: bash
```
martindurant, line 88: I don't think this directive (and below) should be indented

martindurant (commented):
Thanks; not sure how this got messed up.
Minor comment.

Have you built the HTML locally?

docs/source/index.rst
remram44: It does, that was the problem :) That is how you indicate that the code block is part of the list item.

remram44:
I did build locally, that is where the second screenshot comes from (the first one is from the current readthedocs page).

martindurant:
OK, sure that convinces me :)

Thanks!

PR #376, opened 2020-11-05
Implement timeout and key routing for Stream.partition

### LINKED ISSUE #375: Proposal: add timeout and key options to Stream.partition()
I've been using a custom stream for some time. It works the same way as `partition`, but with two differences:

1. You can specify `timeout` option - the number of seconds after which the stream will force-flush the buffer, even if number of elements in it are below the limit.

    This can be useful when you have some data delivery time requirement, but data flow rate is not constant.

2. You can specify `key` option - a callable that will return a partitioning key from a message, effectively splitting the buffer into several independent buffers.

    This is used when you need your batches to be grouped in some manner. I use this to make sure messages in each batch belong to the same source before passing them on to the sink that writes them to an appropriate database table. It's similar to the PartitionRecord processor in NiFi.

I can make a PR, but the questions are:

- should I? does it make sense to add this to core?
- if yes, should this go into `partition` or some new operator? The standard behavior won't change so it's not breaking, but implementation details will. Someone might rely on `partition._buffer` being a list, not a defaultdict.

roveo:
Resolves #375

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-from collections import deque
+from collections import deque, defaultdict
 from datetime import timedelta
 import functools
 import logging
@@ -951,6 +951,19 @@ class slice(Stream):
 class partition(Stream):
     """ Partition stream into tuples of equal size
 
+    Parameters
+    ----------
+    n: int
+        Maximum partition size
+    timeout: int or float, optional
+        Number of seconds after which a partition will be emitted,
+        even if it's size is less than ``n``. If ``None`` (default),
+        a partition will be emitted only when it's size reaches ``n``.
+    key: hashable or callable, optional
+        Emit items with the same key together as a separate partition.
+        If ``key`` is callable, partition will be identified by ``key(x)``,
+        otherwise by ``key[x]``. Defaults to ``None``.
+
     Examples
     --------
     >>> source = Stream()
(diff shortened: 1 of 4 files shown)


roveo:
Looks like flake8 check failed, but not on the files that I changed. Should I fix these?

roveo:
The callback flushing mechanism is flawed, data accumulates in pending callbacks, and this breaks backpressure. Will have to implement some kind of locking mechanism to prevent this.

roveo:
Nope, it isn't, I was just building my test pipeline in an incorrect way. Decribed in #378

martindurant:
> Looks like flake8 check failed, but not on the files that I changed. Should I fix these?

Please do fix - it may be due to a new version of flake8

streamz/core.py
```diff
@@ -951,6 +951,19 @@ def _check_end(self):
...
+    n: int
+        Maximum partition size
+    timeout: int or float, optional
+        Number of seconds after which a partition will be emitted,
+        even if it's size is less than ``n``. If ``None`` (default),
+        a partition will be emitted only when it's size reaches ``n``.
+    key: hashable or callable, optional
+        Emit items with the same key together as a separate partition.
+        If ``key`` is callable, partition will be identified by ``key(x)``,
+        otherwise by ``key[x]``. Defaults to ``None``.
```
chinmaychandak, line 965: Shouldn't this be `x[key]`?

streamz/core.py
```diff
@@ -960,30 +973,67 @@ class partition(Stream):
...
+        metadata_result, self._metadata_buffer[key] = self._metadata_buffer[key], []
+        yield self._emit(tuple(result), list(metadata_result))
+        self._release_refs(metadata_result)
 
+    @gen.coroutine
     def update(self, x, who=None, metadata=None):
         self._retain_refs(metadata)
-        self._buffer.append(x)
+        key = self._get_key(x)
+        buffer = self._buffer[key]
```
chinmaychandak, line 1021: Do we need to create new variables `buffer` and `metadata_buffer` here?

streamz/tests/test_core.py
```diff
@@ -164,6 +164,65 @@ def test_partition():
...
+    source = Stream()
+    L = source.partition(2, key=lambda x: x % 2).sink_to_list()
+
+    for i in range(10):
+        source.emit(i)
+
+    assert L == [(0, 2), (1, 3), (4, 6), (5, 7)]
+
+
+def test_partition_size_one():
```
chinmaychandak, line 217: What is the goal of this test?

chinmaychandak (changes requested):

streamz/core.py
chinmaychandak: cc: @jsmaupin for review because this PR is changing how metadata is being handled in `partition`.

streamz/tests/test_core.py
roveo: Here:

https://github.com/roveo/streamz/blob/3747d8b37c909ff53b16f9a3723140698a0dd8a8/streamz/core.py#L1028-L1036

I didn't have `self.n > 1` condition at first and `self._callbacks[key].cancel()` raised KeyError when `n=1`. So it's kind of an edge case test. If you think it's unnecessary, I'll remove it.

streamz/core.py
roveo: Yes, you're right. Will fix

streamz/core.py
roveo: > Do we need to create new variables buffer and metadata_buffer here?

We really don't have to, but since they're referenced multiple times, I figured it'd me more readable than `self._metadata(_buffer)[key]`.

streamz/tests/test_core.py
chinmaychandak: I'm thinking if we should add a check in `partition` (and update docstring) forcing `n>1`. If `n=1`, the user would be better off avoiding the use of `partition` altogether. What do you think @martindurant?

martindurant: For the sake of test and debug, I think it's OK to allow a redundant partition with n=1.

martindurant:
Feel free to push an empty commit here to restart the CI

PR #377, opened 2020-11-06
allow passing additional options to graphviz, fix #332

### LINKED ISSUE #332: How to pass additional style and color options to graphviz?
My application has many streams of different sources, how can I pass some additional style of color params to graphviz? FOr example I would like streams that start via xls files to be red, streams that start via xml files to be yellow and streams that take input from websockets to be blue. It would also be nice to pass color and style options to intermediate nodes like map, combine latest, etc. Is this possible?

badiku:
allow passing additional options to graphviz, fix #332

### DIFF
--- a/streamz/dataframe/tests/test_dataframes.py
+++ b/streamz/dataframe/tests/test_dataframes.py
@@ -1008,9 +1008,9 @@ def test_windowed_groupby_aggs_with_start_state(stream):
     out_df1 = pd.DataFrame({'name':['Alice', 'Bob', 'Linda', 'Tom'], 'amount':[50.0, 550.0, 100.0, 150.0]})
     assert_eq(output1[-1][1].reset_index(), out_df1)
 
-    
+
 def test_dir(stream):
     example = pd.DataFrame({'name': [], 'amount': []})
     sdf = DataFrame(stream, example=example)
     assert 'name' in dir(sdf)
-    assert 'amount' in dir(sdf)    
+    assert 'amount' in dir(sdf)
--- a/streamz/graph.py
+++ b/streamz/graph.py
@@ -131,7 +131,7 @@ def readable_graph(graph):
 def to_graphviz(graph, **graph_attr):
     import graphviz
 
-    gvz = graphviz.Digraph(graph_attr=graph_attr)
+    gvz = graphviz.Digraph(**graph_attr)
     for node, attrs in graph.nodes.items():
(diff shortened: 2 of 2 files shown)


martindurant:
Unfortunately, this breaks current usage (see the test failure), where the passed kwarg was expected to be already graph_attr. To allow both old and new usage, you could see if any of the top-level kwargs are in the passed dict, 
{'name',
 'comment',
 'filename',
 'directory',
 'format',
 'engine',
 'encoding',
 'graph_attr',
 'node_attr',
 'edge_attr',
 'body',
 'strict'}
and if not, pass as `graphviz.Digraph(**{'graph_attr': graph_attr})`. I would use set intersection to see if there are any top-level keys.

badiku:
> streamz/tests/test_kafka.py::test_kafka_checkpointing_auto_offset_reset_latest FAILED [ 99%]

test_kafka_checkpointing_auto_offset_reset_latest  faild? don't know why.

```python
def launch_kafka():
   ...
   wait_for(predicate, 10, period=0.1)
   ...

streamz/tests/test_kafka.py:592: 
wait_for(lambda: len(out1) == 3 and (len(out1[0]) + len(out1[1]) + len(out1[2])) == 30, 10, 0.1)

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
predicate = <function test_kafka_checkpointing_auto_offset_reset_latest.<locals>.<lambda> at 0x7f4fb8b33670>
timeout = 10, fail_func = 0.1, period = 0.001
    def wait_for(predicate, timeout, fail_func=None, period=0.001):
        """Wait for predicate to turn true, or fail this test"""
        # from distributed.utils_test
        deadline = time() + timeout
        while not predicate():
            sleep(period)
            if time() > deadline:  # pragma: no cover
                if fail_func is not None:
>                   fail_func()
E                   TypeError: 'float' object is not callable
streamz/utils_test.py:115: TypeError
```

so some code at streamz/tests/test_kafka.py:592 send wrong arguments to `wait_for` function,
fail_func got float 0.1 but should be callable?
 
but I did not touch this kafka part.

martindurant:
Try to just run it again (push an empty commit if you have to). The kafka tests have sometimes become flaky, but we don't know why.

PR #379, opened 2020-11-09
Proposal: A simple plugin system

roveo:
Right now, if I need to add my own custom stream nodes, I have to do this:

```py
import mypackage.streamz_extras  # noqa: F401
```

It would be nice to have a way to distribute additional functionality as separate packages that can just be installed via pip. This can be done with entry_points, similar to the way it works in airflow. This is a super bare-bones implementation of this mechanism. Check out https://github.com/roveo/streamz_example_plugin for an example of a plugin.

Problems:
- this should be tested, but I have no idea how, short of shipping the code for example plugin with tests
- plugged-in classes should be checked for validity in some way. I can add a simple check e.g. `isinstance(plugin, Stream)`, but there is probably something else I haven't thought of

### DIFF
--- a/streamz/__init__.py
+++ b/streamz/__init__.py
@@ -3,6 +3,10 @@ from __future__ import absolute_import, division, print_function
 from .core import *
 from .graph import *
 from .sources import *
+from .plugins import load_plugins
+
+load_plugins()
+
 try:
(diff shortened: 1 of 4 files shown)


roveo:
I messed up the branch with irrelevant commits, will create another PR 😞

PR #380, opened 2020-11-09
A simple plugin system

roveo:
Right now, if I need to add my own custom stream nodes, I have to do this:

```py
import mypackage.streamz_extras  # noqa: F401
```

It would be nice to have a way to distribute additional functionality as separate packages that can just be installed via pip. This can be done with entry_points, similar to the way it works in airflow. This is a super bare-bones implementation of this mechanism. Check out https://github.com/roveo/streamz_example_plugin for an example of a plugin.

Problems:
- this should be tested, but I have no idea how, short of shipping the code for example plugin with tests
- plugged-in classes should be checked for validity in some way. I can add a simple check e.g. `isinstance(plugin, Stream)`, but there is probably something else I haven't thought of

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -260,7 +260,7 @@ class Stream(object):
             self.upstreams.remove(upstream)
 
     @classmethod
-    def register_api(cls, modifier=identity):
+    def register_api(cls, modifier=identity, attribute_name=None):
         """ Add callable to Stream API
 
         This allows you to register a new method onto this class.  You can use
@@ -290,9 +290,20 @@ class Stream(object):
             @functools.wraps(func)
             def wrapped(*args, **kwargs):
                 return func(*args, **kwargs)
-            setattr(cls, func.__name__, modifier(wrapped))
+            name = attribute_name if attribute_name else func.__name__
+            setattr(cls, name, modifier(wrapped))
             return func
         return _
+
+    @classmethod
+    def register_plugin_entry_point(cls, entry_point, modifier=identity):
+        def stub(*args, **kwargs):
+            attribute = entry_point.load()
+            cls.register_api(
+                modifier=modifier, attribute_name=entry_point.name
+            )(attribute)
+            return attribute(*args, **kwargs)
+        cls.register_api(modifier=modifier, attribute_name=entry_point.name)(stub)
 
     def start(self):
         """ Start any upstream sources """
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -312,7 +312,7 @@ class Stream(object):
                 raise TypeError(
                     f"Error loading {entry_point.name} "
                     f"from module {entry_point.module_name}: "
-                    f"{entry_point.cls.__name__} must be a subclass of Stream"
+                    f"{node.__class__.__name__} must be a subclass of Stream"
                 )
             cls.register_api(
                 modifier=modifier, attribute_name=entry_point.name
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -307,6 +307,7 @@ class Stream(object):
     @classmethod
     def register_plugin_entry_point(cls, entry_point, modifier=identity):
         def stub(*args, **kwargs):
+            """ Entrypoints-based streamz plugin. Will be loaded on first call. """
             node = entry_point.load()
             if not issubclass(node, Stream):
                 raise TypeError(
@@ -314,9 +315,10 @@ class Stream(object):
                     f"from module {entry_point.module_name}: "
                     f"{node.__class__.__name__} must be a subclass of Stream"
                 )
-            cls.register_api(
-                modifier=modifier, attribute_name=entry_point.name
-            )(node)
+            if getattr(cls, entry_point.name).__name__ == "stub":
+                cls.register_api(
+                    modifier=modifier, attribute_name=entry_point.name
+                )(node)
             return node(*args, **kwargs)
         cls.register_api(modifier=modifier, attribute_name=entry_point.name)(stub)
 
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -273,7 +273,7 @@ class Stream(object):
             >>> Stream().foo(...)  # this works now
 
         It attaches the callable as a normal attribute to the class object.  In
-        doing so it respsects inheritance (all subclasses of Stream will also
+        doing so it respects inheritance (all subclasses of Stream will also
         get the foo attribute).
 
         By default callables are assumed to be instance methods.  If you like
@@ -285,6 +285,15 @@ class Stream(object):
             ...     ...
 
             >>> Stream.foo(...)  # Foo operates as a static method
+
+        You can also provide an optional ``attribute_name`` argument to control
+        the name of the attribute your callable will be attached as.
+
+            >>> @Stream.register_api(attribute_name="bar")
+            ... class foo(Stream):
+            ...     ...
+
+            >> Stream().bar(...)  # foo was actually attached as bar
         """
         def _(func):
             @functools.wraps(func)
@@ -298,11 +307,17 @@ class Stream(object):
     @classmethod
     def register_plugin_entry_point(cls, entry_point, modifier=identity):
         def stub(*args, **kwargs):
-            attribute = entry_point.load()
+            node = entry_point.load()
+            if not issubclass(node, Stream):
+                raise TypeError(
+                    f"Error loading {entry_point.name} "
+                    f"from module {entry_point.module_name}: "
+                    f"{entry_point.cls.__name__} must be a subclass of Stream"
+                )
             cls.register_api(
                 modifier=modifier, attribute_name=entry_point.name
-            )(attribute)
-            return attribute(*args, **kwargs)
+            )(node)
+            return node(*args, **kwargs)
         cls.register_api(modifier=modifier, attribute_name=entry_point.name)(stub)
 
     def start(self):
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -306,6 +306,12 @@ class Stream(object):
 
     @classmethod
     def register_plugin_entry_point(cls, entry_point, modifier=identity):
+        if hasattr(cls, entry_point.name):
+            raise ValueError(
+                f"Can't add {entry_point.name} from {entry_point.module_name} "
+                f"to {cls.__name__}: duplicate method name."
+            )
+
         def stub(*args, **kwargs):
             """ Entrypoints-based streamz plugin. Will be loaded on first call. """
             node = entry_point.load()
--- a/streamz/tests/test_plugins.py
+++ b/streamz/tests/test_plugins.py
@@ -53,3 +53,17 @@ def test_register_plugin_entry_point_raises():
 
     with pytest.raises(TypeError):
         Stream.test()
+
+
+def test_register_plugin_entry_point_already_registered():
+    @Stream.register_api()
+    class test(Stream):
+        pass
+
+    entry_point = MockEntryPoint("test_double", test, "test_module")
+
+    Stream.register_plugin_entry_point(entry_point)
+
+    assert Stream.test_double.__name__ == "stub"
+    Stream.test_double()
+    assert Stream.test_double.__name__ == "test"
--- a/streamz/tests/test_plugins.py
+++ b/streamz/tests/test_plugins.py
@@ -1,0 +1,38 @@
+from streamz.sources import Source
+from streamz import Stream
+
+
+class MockEntryPoint:
+
+    def __init__(self, name, cls):
+        self.name = name
+        self.cls = cls
+
+    def load(self):
+        return self.cls
+
+
+def test_register_plugin_entry_point():
+    class test(Stream):
+        pass
+
+    entry_point = MockEntryPoint("test_node", test)
+    Stream.register_plugin_entry_point(entry_point)
+
+    assert Stream.test_node.__name__ == "stub"
+
+    Stream().test_node()
+
+    assert Stream.test_node.__name__ == "test"
+
+
+def test_register_plugin_entry_point_modifier():
+    class test(Source):
+        pass
+
+    entry_point = MockEntryPoint("from_test", test)
+    Stream.register_plugin_entry_point(entry_point, staticmethod)
+
+    Stream.from_test()
+
+    assert Stream.from_test.__self__ is Stream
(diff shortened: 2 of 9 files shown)


martindurant:
Thank you. I totally agree that using entry_points is a great alternative or addition to the registration system.
@CJ-Wright @jsmaupin @chinmaychandak - anyone have time and interest to give this a once-over?

jsmaupin:
I think this is a great idea.  I was hoping that we could figure out how do this in the future for sources and sinks where you could connect Streamz to external services using packages install-able via pip or conda because, these connections often have a lot of dependencies that you don't necessarily want in your core package.  I hadn't considered it for functionality that operates in the middle of the pipeline. I think this is the step that gets us in that direction.

CJ-Wright:
I think this is great! How should we document all these plugins so people can discover them?

martindurant:
If we go with entry_points as a way of declaring plugins, then we *do not* document them, at least not exhaustively. Intake faces this: we have a page of known plugins, which people can submit PRs for changing, but there's nothing to stop new plugins appearing which we don't know about. It would be possible to have a configuration that only loads a select number of plugins, or instead to not load the plugins except upon explicit user invocation.

chinmaychandak:
I think is really nice! 

I think for validating the plugins, we would definitely need `isinstance(plugin, Stream)` (similar to Airflow which tests against one of its subclasses called `PluginManager` which I don't think is necessary at this point).  

Also, for testing an example plugin, having a separate repository/page called `streamz-plugins` could be an option? I think that's what Airflow does, too. The `from_iterable` node you wrote can be a good plugin for testing. 

As for tests for future plugins, I think if we consider adding a plugin to `setup.py` in streamz, then we would need to verify the plugin behavior separately beforehand; if, however, we allow the plugin to be loaded lazily (only when the user explicitly invokes them, either through CLI or otherwise), then I think we don't need to worry about having tests in streamz itself (the plugin repo can have those tests, I mean). I'm just speaking my mind here, I may be missing something!

roveo:
> Also, for testing an example plugin, having a separate repository/page called streamz-plugins could be an option?
I think this is the way to go. Then I'll add tests, so that they are not bound to the repo in my account.

> As for tests for future plugins, I think if we consider adding a plugin to setup.py in streamz, then we would need to verify the plugin behavior separately beforehand
So, plugins in streamz-plugins repo that are tested and "verified" can be provided via `extras_require` in setup.py, but nothing is stopping anyone from creating their own plugin package and installing it with pip.

chinmaychandak:
This seems reasonable to me, although I would defer to Martin and CJ to make the final call whether or not this is the best way to proceed.

chinmaychandak:
I see Travis builds timing out. Is this due to it being overloaded? Should we try something like `install: travis_wait 30 mvn install` temporarily?

martindurant:
We need to transfer our tests away from travis...

martindurant:
Yes, we should do whatever it takes to get travis to pass for now...

I am only now looking at the actual code. I wonder, is there any appetite for making this system lazy? As it is, streamz will import all packages that claim to have relevant plugins - but the node classes still need to be annotated with decorators as before.

I could imagine entry points spelled like:
```
streamz.plugins: [
  "transform=streamz_addon.module.TransformNode"
]
```
which adds "transform" to the dir() of Stream, but only imports and registers when the `.transform` is first used (i.e., via a `__getattr__`).

Furthermore, we may want to be more specific with our entrypoint group naming: 
- streamz.node
- streamz.source
- streamz.sink
- ...

What do you think?

btw: does this require the package `entrypoints` as a dependency?

CJ-Wright:
I like having them lazy, since that allows us to not need the plugins to be installed until we actually need them

CJ-Wright:
@martindurant would it be worthwhile to setup the CI using conda-forge's CI provisioning system? That way if we need to move to something we just rerender?

martindurant:
Sorry @CJ-Wright , you're the expert there, I really don't know what's involved.

CJ-Wright:
I'll put up a PR if we're interested, but the rough outline is here: https://conda-forge.org/docs/user/ci-skeleton.html

martindurant:
Does it work for github actions CI?

CJ-Wright:
Not yet, since CF doesn't use GH actions as a CI (yet). The hope of this approach would be that we don't need to care about what particular CI we are using, since we could rerender and the CIs could move.

We could add GH actions to CF and then it would support that CI.

martindurant:
In that case I am hesitant, since the dask repos are all going to github. I don't have a circle or azure account myself - is it all free like Travis used to be?

CJ-Wright:
Sorry, can we move the CI conversation to an issue? I don't want to derail things here. https://github.com/python-streamz/streamz/issues/382

roveo:
> As it is, streamz will import all packages that claim to have relevant plugins - but the node classes still need to be annotated with decorators as before.

Do you mean that with lazy loading they won't have to be annotated?

> adds "transform" to the dir() of Stream, but only imports and registers when the .transform is first used (i.e., via a __getattr__)

This can be done with a @property that looks it up in a class attribute dict (e.g. Stream._plugins) and loads the entrypoint if it's not already loaded.

> Furthermore, we may want to be more specific with our entrypoint group naming

Do these types of nodes require different treatment when being loaded?

> btw: does this require the package entrypoints as a dependency?

I don't think so, this is built into setuptools. `entrypoints` looks like a different package (with the last 2 releases in 2017 and 2019).

martindurant:
> Do you mean that with lazy loading they won't have to be annotated?

Indeed, this would be an alternative to the annotation; although doing both should be allowed for backward compatibility. 

> This can be done with a @property that looks it up in a class attribute dict (e.g. Stream._plugins) and loads the entrypoint if it's not already loaded.

Something like that, yes. There are many ways to achieve it. You would call register_api on the class once it's imported, so it gets added to the namespace and you don't need to import it again.

> Do these types of nodes require different treatment when being loaded?

*maybe*. This gives us more flexibility for the future. We could argue that, for the time being, all possible plugins as "nodes" (i.e., they appear as attributes of Stream). My motivation here is, that in Intake we started off having things called "plugins", but then made other parts of the library pluggable and had to, painfully, rename the original plugins to "drivers".

> I don't think so, this is built into setuptools.

Better check that setuptools has this functionality for all python versions of interest. Also, setuptools had better be included in the deps (yes, I know everyone already has it).

roveo:
> You would call register_api on the class once it's imported

Then register_api would have to allow us to specify which attribute the class will be registered at. Right now it's `setattr(cls, func.__name__, modifier(wrapped))`, so it's always the class name.

> Better check that setuptools has this functionality for all python versions of interest. Also, setuptools had better be included in the deps (yes, I know everyone already has it).

OK, will do.

martindurant:
Correct. I imagine we'll have redundancy for now "name=package.module.name", but there might be a case for when we want the class name and method name to be different, which is not yet possible.

Actually, there is a difference between normal nodes and sources: the latter are registered as `staticmethod`.

chinmaychandak:
@roveo, could you try something like `install: travis_wait 30 mvn install` in `travis.yml`, until we can switch from Travis to GitHub Actions / CF? I can't guarantee it will work (because the issue is due to Travis ending free builds), but I think it's worth a shot.

roveo:
> Better check that setuptools has this functionality for all python versions of interest.
Do we have a list of python versions that should be supported? Setuptools supported entry_points since `0.6a1` (year 2009), so I don't think this should be a problem.

roveo:
Added Plugins page to docs (I'm not a native speaker, so edits and additions are very welcome).

Note that there's a table commented out in __Known plugins__ section. I think at this point we should think about which existing functionality should go into extras (the leading candidate being kafka).

streamz/core.py
```diff
@@ -285,15 +285,41 @@ def register_api(cls, modifier=identity):
...
                 return func(*args, **kwargs)
-            setattr(cls, func.__name__, modifier(wrapped))
+            name = attribute_name if attribute_name else func.__name__
+            setattr(cls, name, modifier(wrapped))
             return func
         return _
 
+    @classmethod
+    def register_plugin_entry_point(cls, entry_point, modifier=identity):
+        def stub(*args, **kwargs):
```
martindurant, line 309: Could use a docstring, since the real method won't appear until after it's been called at least once. 
"entrypoints-based streamz plugin, loaded on first call"

streamz/tests/test_plugins.py
```diff
@@ -0,0 +1,69 @@
...
+        fn.__name__ = 'modified_name'
+        return staticmethod(fn)
+
+    entry_point = MockEntryPoint("from_test", test_source)
+    Stream.register_plugin_entry_point(entry_point, modifier)
+
+    Stream.from_test()
+
+    assert Stream.from_test.__name__ == "modified_name"
+
```
martindurant, line 44:
```
assert inspect.isfunction(Stream().from_test)
```
(would be a method if not static)
Note that `Stream` is instantiated.

streamz/tests/test_plugins.py
```diff
@@ -0,0 +1,69 @@
...
+def test_register_plugin_entry_point_already_registered():
+    @Stream.register_api()
+    class test(Stream):
+        pass
+
+    entry_point = MockEntryPoint("test_double", test, "test_module")
+
+    Stream.register_plugin_entry_point(entry_point)
+
+    assert Stream.test_double.__name__ == "stub"
```
martindurant, line 67: Actually, I'm not sure whether the method should have been overwritten - presumably `Stream.test_double` already existed before the call to `register_plugin_entry_point`.

In normal usage, the stubs always come first, since they happen at import time, and streamz should always be imported before derived classes.

martindurant (commented):
I really like this PR

streamz/tests/test_plugins.py
roveo: Nice! thanks, I spent some time figuring out how to do this better, haven't thought of `inspect`.

streamz/tests/test_plugins.py
roveo: Yes, I guess this test actually doesn't test anything. I wanted to test that running `Stream.register_api` twice on the same class doesn't break anything, in case the implementation of `register_api` changes. We can just check for this in `register_plugin_entry_point`.

martindurant: The test is OK, but not comprehensive. The cases are:
- register_api, then register entrypoint
- register entrypoint then register_api
- either of the to methods called twice

In every case, calling the method should result in the same test class.

We *could* check for whether the name is being overwritten, but I don't think that's essential.

roveo: Btw, what should we do if a plugin wants to override a built-in method? On the one hand, this would allow people to create useful extensions of built-in functionality (my version of `partition` could be provided as a plugin long before it's merged into core), on the other — it could break things and lead to confusion.

CJ-Wright: Personally I would have the system balk at a built-in override. My hope would be that names are cheap, so adding more text onto a would be `partition` describing how it is difficult would add less burden than issues around which `partition` was used. This could get particularly thorny as envs get updated, code that previously worked and relied on the built-in are now producing errors in subtle ways that could be difficult to find.

roveo: > In every case, calling the method should result in the same test class.

The class is stored in `Stream.method.__wrapped__`, but I don't see how anything else but the test class can end up there in any of these cases.

martindurant: You mean we should have a `clobber=False` argument for the register methods? 
I wonder in that case, if there is a sane way to tell when an entrypoint and register_api refer to the same thing, which would not be a problem even without clobber.

roveo: > I wonder in that case, if there is a sane way to tell when an entrypoint and register_api refer to the same thing, which would not be a problem even without clobber.

We can just check that `hasattr(Stream, entry_point.name)`. But would have to happen during plugin load (so that we're not overwriting built-in methods) and so importing streamz with a bad plugin would result in an error.

martindurant: but this will fail when someone adds the entrypoint to a class that already has register_api, no?

Note @CJ-Wright , that the current implementation does allow you to use register_api to overwrite methods at will. Of course this is python, so we can't ever disable someone who wants to do that anyway.

CJ-Wright: That's fair. To some extent this boils down to locality for me. If the author of the code is the user of the env then if they override things then they are in a better place to clean up. When installing 3rd party things that state may not hold as well.

roveo: > If the author of the code is the user of the env then if they override things then they are in a better place to clean up. When installing 3rd party things that state may not hold as well.

I agree. With `register_api` things are explicit and visible. A plugin is a black box.

@martindurant please look at the last commit to see what I mean

streamz/core.py
```diff
@@ -307,16 +307,18 @@ def wrapped(*args, **kwargs):
...
             if not issubclass(node, Stream):
                 raise TypeError(
                     f"Error loading {entry_point.name} "
                     f"from module {entry_point.module_name}: "
                     f"{node.__class__.__name__} must be a subclass of Stream"
                 )
-            cls.register_api(
-                modifier=modifier, attribute_name=entry_point.name
-            )(node)
+            if getattr(cls, entry_point.name).__name__ == "stub":
```
martindurant, line 324: I don't *think* this can be False, but it doesn't hurt to check.

martindurant (commented):
I'm happy with this. @CJ-Wright ?

streamz/core.py
roveo: It is false when the plugin class returned from `entry_point.load()` is decorated with `@Stream.register_api`. Then it is registered right away when loaded, and so at that moment stub is overwritten with the actual class.

CJ-Wright (approved):
LGTM! One thing to consider is scraping the conda-forge packages/data for packages who declare the streamz plugin, if we are interested in building a plugin registry.

martindurant:
Right, but there aren't any yet :)
I actually would expect any plugin developers to contact us, if they wanted to be listed on the streamz documentation somewhere.

martindurant:
@chinmaychandak : that kafka test is failing too often! (and giving an incorrect message, due to using positional rather than keyword arguments to wait)

CJ-Wright:
It was more of a general idea, since other plugin mediated systems could use that approach as well.

Issue #378, opened 2020-11-07
Using connect on an empty Stream() breaks backpressure


roveo:
Here's an example:

```py
from time import sleep
from streamz import Stream


class Counter:
    def __init__(self):
        self.n = 0

    def __call__(self, x):
        self.n += 1
        return x

    def __str__(self):
        return f"Counter({self.n})"


cin = Counter()
cout = Counter()

src = Stream.from_textfile("1000_lines.txt")
out = Stream()
out.map(cin).rate_limit(1).sink(cout)
src.connect(out)

src.start()
sleep(1)
print(cin, cout)
assert cin.n < 1000  # AssertionError
```

I expected this to work the same way as `src.map(...).rate_limit(...).sink(...)`, but `yield _emit()` in the source doesn't block. If you go into `rate_limit.update`, you'll see that it gets called multiple times, so data actually accumulates in the form of running blocked coroutines.

You can do `out = Stream().map(cin)` and then in will work fine, but if you want to additionally emit something from `out` _before_ the first map (so that it is counted by `cin`), this isn't a solution.

martindurant:
Sorry, are you saying that the existence of the line
```
src.connect(out)
```
makes the difference in the behaviour of the variable `cin`, on the other branch of the stream?

I do note that `Counter` ought to be a subclass of `Stream`. I imagine that matters.

roveo:
> I do note that Counter ought to be a subclass of Stream. I imagine that matters.

Why? It's just a callable that's passed to map.

> Sorry, are you saying that the existence of the line `src.connect(out)` makes the difference in the behaviour of the variable cin, on the other branch of the stream?

No, that wasn't what I was trying to demonstrate. The counters are there just to show that backpressure doesn't work as expected, but the example would still work the same without them.

If you write it like this:

```py
src = Stream.from_textfile("1000_lines.txt")
out = Stream()
out.map(cin).rate_limit(1).sink(cout)
src.connect(out)
```

and then start the pipeline, the asynchronous source (from_file) will immediately emit 1000 messages, while I expect them to be emitted one by one every second due to the existence of backpressure.

The counters in my example are to make it clear that 1000 messages passed through `.map(cin)` (cin stands for "count in"), but only one message went to `.sink(cout)` (count out). So these messages in this case are somewhere between the 1st and the 2nd `.map` nodes, namely they are accumulated as running coroutines `rate_limit.update`.

Here's the same thing in the form of a test:

```py
def test_connect_empty_stream():
[...]

PR #381, opened 2020-11-09
Add backpressure test

martindurant:
Ref #378 

@roveo - how is this example different from yours? This passes.

### DIFF
--- a/streamz/tests/test_sources.py
+++ b/streamz/tests/test_sources.py
@@ -1,6 +1,6 @@
 from flaky import flaky
 import pytest
-from streamz import Source
+from streamz import Source, Stream
 from streamz.utils_test import wait_for, await_for, gen_test
 import socket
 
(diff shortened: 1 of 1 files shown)


roveo:
- My guess is, the part of the pipeline that applies backpressure to the source (rate_limit) is built by method chaining. There's no `Source()` node in the middle, so everything works OK, and this affects `sout` too.
- If you change the `Stream.update` method

https://github.com/python-streamz/streamz/blob/e9a25452ca13af5ca0edf3346f8c67353d291501/streamz/core.py#L450-L453

to be a coroutine:

```py
    @gen.coroutine
        yield self._emit(x, metadata=metadata)
```

then my test example passes too.

roveo:
OK, so here's a log of calls with `time.time()` from my example:

```
1604961709.142924 from_list.emit 0
1604961709.1429608 Stream.update 0
1604961709.1429892 rate_limit.update 0
1604961709.143005 sink 0
1604961709.1431391 from_list.emit 1
1604961709.1431592 Stream.update 1
1604961709.143174 rate_limit.update 1
1604961709.1433449 from_list.emit 2
1604961709.1433642 Stream.update 2
1604961709.143379 rate_limit.update 2
1604961709.143423 from_list.emit 3
1604961709.143439 Stream.update 3
1604961709.143453 rate_limit.update 3
1604961709.1434932 from_list.emit 4
1604961709.143507 Stream.update 4
1604961709.14352 rate_limit.update 4

AssertionError  # from the test (source list is empty). At this point we have 4 instances of rate_limit.update running.

1604961710.145355 sink 1
1604961711.1437702 sink 2
1604961712.144693 sink 3
1604961713.147051 sink 4
```

And here's the same thing, but with `Stream.update` a coroutine:

```
1604961822.514569 from_list.emit 0
1604961822.514626 Stream.update 0
1604961822.514658 rate_limit.update 0
1604961822.514681 sink 0
1604961822.515048 from_list.emit 1
1604961822.515075 Stream.update 1
1604961822.515093 rate_limit.update 1
1604961823.5194051 sink 1
1604961823.519525 from_list.emit 2
1604961823.519547 Stream.update 2
1604961823.519562 rate_limit.update 2
1604961824.5165849 sink 2
1604961824.516799 from_list.emit 3
1604961824.516845 Stream.update 3
1604961824.5168781 rate_limit.update 3
1604961825.518817 sink 3
1604961825.5191329 from_list.emit 4
1604961825.519197 Stream.update 4
1604961825.5192409 rate_limit.update 4
1604961826.519273 sink 4
```

As you can see, in the first instance `rate_limit.update` gets called 5 times immediately, and then these running coroutines return each second.

martindurant:
I assume that the change would break things elsewhere? Or do you have the whole test suite passing too?

roveo:
> do you have the whole test suite passing
Almost everything passes locally, but it gets stuck at 98% somewhere in Kafka tests. Core, dataframes etc. are ok.

martindurant:
Sorry, this one got lost
(thanks for pushing on this and other progress in this repo!)

For some reason, this tests didn't run (ref #387 ), can you please kick it?

roveo:
I don't think I can, this is your PR :)

martindurant:
oops :) I suppose we should have a new PR with this test, or something like it, plus the coroutine change, so that we are talking about the same thing.

roveo:
Created #390

PR #386, opened 2020-11-13
Refactor sinks

### LINKED ISSUE #384: Creating custom sink nodes
I tried to create a sink for my example plugin (to manually test #380) and encountered some issues.

It's not entirely clear from the docs how to do this. There are classes for `Stream` and `Source` that one can inherit from, but there is no such class for sinks. At the same time, sinks clearly demonstrate some special behavior. They have to add themselves to `_global_sinks`, without which they don't work (btw, why?). Since we're talking about plugins, I think this detail should be hidden from whoever wants to create one.

There is only one reference implementation of a sink (_the_ `sink`) in the codebase. There's also `sink_to_list`, which is just a method of `Stream`, and `sink_to_file`, which is not really a sink — it's a function and it's not in `Streams`'s `dir()`.

You can really get by with just the built-in `sink` by passing a callable object to it, but @martindurant mentioned sinks in the plugin discussion, so there should be at least some situation where you'd want to create another.

roveo:
Resolves #384 

- added `Sink` class
- moved related code to `sinks.py`
- moved related tests to `test_sinks.py`
- didn't remove `sink_to_file` (which is not a sink, but still someone might be using it)
- added `sink_to_textfile` (for symmetry) since there is `from_textfile` in core
- `sink_to_list` is a special case since it returns a list. This can be done with a class, but involves some black magic with `__new__`, don't think it's worth it.

### DIFF
--- a/streamz/__init__.py
+++ b/streamz/__init__.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 from .core import *
 from .graph import *
 from .sources import *
+from .sinks import *
 try:
     from .dask import DaskStream, scatter
 except ImportError:
--- a/streamz/sinks.py
+++ b/streamz/sinks.py
@@ -55,15 +55,16 @@ class sink(Sink):
 
 @Stream.register_api()
 class sink_to_textfile(Sink):
-    """ Write elements to a plain text file, one element per line. Type of elements
-        must be ``str``.
+    """ Write elements to a plain text file, one element per line.
+
+        Type of elements must be ``str``.
 
         Arguments
         ---------
         file: str or file-like
             File to write the elements to. ``str`` is treated as a file name to open.
-            If file-like, descriptor must be open in text mode.
-            Note that the file descriptor will be closed when sink is destroyed.
+            If file-like, descriptor must be open in text mode. Note that the file
+            descriptor will be closed when this sink is destroyed.
         end: str, optional
             This value will be written to the file after each element.
             Defaults to newline character.
@@ -74,7 +75,7 @@ class sink_to_textfile(Sink):
         Examples
         --------
         >>> source = Stream()
-        >>> source.map(str).sink_to_file("test.txt")
+        >>> source.map(str).sink_to_textfile("test.txt")
         >>> source.emit(0)
         >>> source.emit(1)
         >>> print(open("test.txt", "r").read())
--- a/streamz/sinks.py
+++ b/streamz/sinks.py
@@ -1,0 +1,95 @@
+from tornado import gen
+
+from streamz import Stream
+
+_global_sinks = set()
+
+
+class Sink(Stream):
+
+    _graphviz_shape = 'trapezium'
+
+    def __init__(self, upstream, **kwargs):
+        super().__init__(upstream, **kwargs)
+        _global_sinks.add(self)
+
+
+@Stream.register_api()
+class sink(Sink):
+    """ Apply a function on every element
+
+    Examples
+    --------
+    >>> source = Stream()
+    >>> L = list()
+    >>> source.sink(L.append)
+    >>> source.sink(print)
+    >>> source.sink(print)
+    >>> source.emit(123)
+    123
+    123
+    >>> L
+    [123]
+
+    See Also
+    --------
+    map
+    Stream.sink_to_list
+    """
+
+    def __init__(self, upstream, func, *args, **kwargs):
+        self.func = func
+        # take the stream specific kwargs out
+        stream_name = kwargs.pop("stream_name", None)
+        self.kwargs = kwargs
+        self.args = args
+        super().__init__(upstream, stream_name=stream_name)
+
+    def update(self, x, who=None, metadata=None):
+        result = self.func(x, *self.args, **self.kwargs)
+        if gen.isawaitable(result):
+            return result
+        else:
+            return []
+
+
+@Stream.register_api()
+class sink_to_textfile(Sink):
+    """ Write elements to a plain text file, one element per line. Type of elements
+        must be ``str``.
+
+        Arguments
+        ---------
+        file: str or file-like
+            File to write the elements to. ``str`` is treated as a file name to open.
+            If file-like, descriptor must be open in text mode.
+            Note that the file descriptor will be closed when sink is destroyed.
+        end: str, optional
+            This value will be written to the file after each element.
+            Defaults to newline character.
+        mode: str, optional
+            If file is ``str``, file will be opened in this mode. Defaults to ``"a"``
+            (append mode).
+
+        Examples
+        --------
+        >>> source = Stream()
+        >>> source.map(str).sink_to_file("test.txt")
+        >>> source.emit(0)
+        >>> source.emit(1)
+        >>> print(open("test.txt", "r").read())
+        0
+        1
+    """
+    def __init__(self, upstream, file, end="\n", mode="a", **kwargs):
+        self._fp = open(file, mode=mode, buffering=1) if isinstance(file, str) else file
+        self._end = end
+        super().__init__(upstream, ensure_io_loop=True, **kwargs)
+
+    def __del__(self):
+        self._fp.close()
+
+    @gen.coroutine
+    def update(self, x, who=None, metadata=None):
+        yield self.loop.run_in_executor(None, self._fp.write, x)
+        yield self.loop.run_in_executor(None, self._fp.write, self._end)
--- a/streamz/sinks.py
+++ b/streamz/sinks.py
@@ -67,6 +67,10 @@ class sink(Sink):
         else:
             return []
 
+    def destroy(self):
+        super().destroy()
+        _global_sinks.remove(self)
+
 
 @Stream.register_api()
 class sink_to_textfile(Sink):
--- a/streamz/sinks.py
+++ b/streamz/sinks.py
@@ -105,11 +105,10 @@ class sink_to_textfile(Sink):
         self._end = end
         self._fp = open(file, mode=mode) if isinstance(file, str) else file
         weakref.finalize(self, self._fp.close)
-        super().__init__(upstream, ensure_io_loop=True, **kwargs)
+        super().__init__(upstream, **kwargs)
 
     def __del__(self):
         self._fp.close()
 
-    @gen.coroutine
     def update(self, x, who=None, metadata=None):
-        yield self.loop.run_in_executor(None, self._fp.write, x + self._end)
+        self._fp.write(x + self._end)
--- a/streamz/sinks.py
+++ b/streamz/sinks.py
@@ -1,7 +1,11 @@
+import inspect
+import weakref
+
 from tornado import gen
 
 from streamz import Stream
 
+# sinks add themselves here to avoid being garbage-collected
 _global_sinks = set()
 
 
@@ -17,6 +21,16 @@ class Sink(Stream):
 @Stream.register_api()
 class sink(Sink):
     """ Apply a function on every element
+
+    Parameters
+    ----------
+    func: callable
+        A function that will be applied on every element.
+    args:
+        Positional arguments that will be passed to ``func`` after the incoming element.
+    kwargs:
+        Stream-specific arguments will be passed to ``Stream.__init__``, the rest of
+        them will be passed to ``func``.
 
     Examples
     --------
@@ -40,10 +54,11 @@ class sink(Sink):
     def __init__(self, upstream, func, *args, **kwargs):
         self.func = func
         # take the stream specific kwargs out
-        stream_name = kwargs.pop("stream_name", None)
-        self.kwargs = kwargs
+        sig = set(inspect.signature(Stream).parameters)
+        stream_kwargs = {k: v for (k, v) in kwargs.items() if k in sig}
+        self.kwargs = {k: v for (k, v) in kwargs.items() if k not in sig}
         self.args = args
-        super().__init__(upstream, stream_name=stream_name)
+        super().__init__(upstream, **stream_kwargs)
 
     def update(self, x, who=None, metadata=None):
         result = self.func(x, *self.args, **self.kwargs)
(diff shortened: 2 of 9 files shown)


streamz/sinks.py
```diff
@@ -0,0 +1,96 @@
...
+        >>> print(open("test.txt", "r").read())
+        0
+        1
+    """
+    def __init__(self, upstream, file, end="\n", mode="a", **kwargs):
+        self._fp = open(file, mode=mode, buffering=1) if isinstance(file, str) else file
+        self._end = end
+        super().__init__(upstream, ensure_io_loop=True, **kwargs)
+
+    def __del__(self):
```
martindurant, line 90: I am preferring using `weakref.finalize` for this kind of thing, because `__del__` often gets called during interpreter shutdown, when the variable has already been cleaned up, resulting in "ignored exceptions" that are harmless, but look bad.
```
weakref.finalize(self, self._fp.close)
```
in init, ensures that the file keeps a reference until then.

streamz/sinks.py
```diff
@@ -0,0 +1,96 @@
...
+        >>> source = Stream()
+        >>> source.map(str).sink_to_textfile("test.txt")
+        >>> source.emit(0)
+        >>> source.emit(1)
+        >>> print(open("test.txt", "r").read())
+        0
+        1
+    """
+    def __init__(self, upstream, file, end="\n", mode="a", **kwargs):
+        self._fp = open(file, mode=mode, buffering=1) if isinstance(file, str) else file
```
martindurant, line 86: Integrate with `fsspec.open`?
Why set the buffering?

streamz/sinks.py
```diff
@@ -0,0 +1,96 @@
...
+        self._end = end
+        super().__init__(upstream, ensure_io_loop=True, **kwargs)
+
+    def __del__(self):
+        self._fp.close()
+
+    @gen.coroutine
+    def update(self, x, who=None, metadata=None):
+        yield self.loop.run_in_executor(None, self._fp.write, x)
```
martindurant, line 95: why run_in_executor?
I would combine the two calls and write `x + self._end`

streamz/tests/test_sinks.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_sink_to_textfile_named():
+    source = Stream()
+    with tmpfile() as filename:
+        source.map(str).sink_to_textfile(filename)
+        source.emit(0)
+        source.emit(1)
+
+        sleep(0.01)
```
martindurant, line 45: wait_for

streamz/tests/test_sinks.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_sink_to_textfile_closes():
+    source = Stream()
+    with tmpfile() as filename:
+        sink = source.sink_to_textfile(filename)
+        fp = sink._fp
+        _global_sinks.remove(sink)
+        del sink
+
+        with pytest.raises(ValueError):  # I/O operation on closed file
```
martindurant, line 58: You can specify the expected error with `match=`

streamz/tests/test_sinks.py
```diff
@@ -0,0 +1,59 @@
...
+
+        assert open(filename, "r").read() == "0\n1\n"
+
+
+def test_sink_to_textfile_closes():
+    source = Stream()
+    with tmpfile() as filename:
+        sink = source.sink_to_textfile(filename)
+        fp = sink._fp
+        _global_sinks.remove(sink)
```
martindurant, line 55: Is there no public way to do this? It should be possible to destruct sinks (and their whole graph).

streamz/sinks.py
```diff
@@ -0,0 +1,96 @@
...
+from streamz import Stream
+
+_global_sinks = set()
+
+
+class Sink(Stream):
+
+    _graphviz_shape = 'trapezium'
+
+    def __init__(self, upstream, **kwargs):
```
CJ-Wright, line 12: Does this need special removal methods?

streamz/sinks.py
roveo: > Why set the buffering?

Used line buffering in development to see what's going on without delay, forgot to remove. Also some tests won't work as is without it, but I guess I'll just use `wait_for`.

> Integrate with `fsspec.open`?

This is really a question about what filesystem functionality should be provided in core. If we want to integrate this sink with fsspec, then from_textfile should be updated too (or some other nodes added, like `from_fs`/`sink_to_fs`), which is beyond sink refactoring. And then, we should think about how users might have a way to install fsspec extras (s3, hdfs etc.) and what other fs-related nodes might be useful, like:

- listening for new files in a folder/glob and "tailing" them
- writing to same-sized file partitions (like `split` command)
- writing to files partitioned by time (e.g. rotating every 5 minutes)

What I'm getting at is that it looks like a plugin to me. We could have `streamz[fs]` for nodes and `fsspec` built-ins and then `streamz[s3,hdfs]` for extras.

streamz/sinks.py
roveo: > why run_in_executor?

Probably makes little sense with local filesystem and it's tiny delays, but I wanted it to be non-blocking (like other i/o nodes), and non-blocking local i/o can only be done with threads.

streamz/tests/test_sinks.py
roveo: OK

streamz/__init__.py
```diff
@@ -3,6 +3,7 @@
 from .core import *
 from .graph import *
 from .sources import *
+from .sinks import *
```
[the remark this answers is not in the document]
roveo: I agree, but it comes down to backwards-compatibility. Since this has already been done like this (`*`), someone might import something internal directly from `streamz`, like `_global_sinks` for some custom implementation of a sink. We might say that it's OK that people will have to tweak code a little bit or call it 2.0, but I don't want to change it without your/other maintainers' "OK".

streamz/tests/test_sinks.py
roveo: Will do.

streamz/tests/test_sinks.py
roveo: I thought about this. It's tricky to figure out what exactly the user wants to destroy, especially with complex graphs. You don't want to keep a reference to the sink in a variable, so this should probably be a method of `Stream` that can be called on any node and then propagated through the graph. But if there are many branches, you don't know where it should be propagated. Maybe something like `.destroy_downstream()`?

streamz/sinks.py
roveo: Answered (with a question, really) here: https://github.com/python-streamz/streamz/pull/386#discussion_r523273681

streamz/tests/test_sinks.py
CJ-Wright: This exists, I think in https://github.com/python-streamz/streamz/blob/master/streamz/core.py#L250 and https://github.com/python-streamz/streamz/blob/master/streamz/core.py#L506

CJ-Wright: I should note that the existing `destroy` should use the `_remove_upstream` methods since that handles things properly.

roveo: In terms of semantics, `.destroy` sounds like something is permanently deleted, while in reality it works more like `disconnect_from_upstreams`. If there's a sink at the end, nothing is destroyed. Maybe it should call `.destroy` on all downstreams, and then for sinks we should override `destroy` to also delete itself from `_global_sinks`?

streamz/sinks.py
```diff
@@ -0,0 +1,96 @@
...
+    See Also
+    --------
+    map
+    Stream.sink_to_list
+    """
+
+    def __init__(self, upstream, func, *args, **kwargs):
+        self.func = func
+        # take the stream specific kwargs out
+        stream_name = kwargs.pop("stream_name", None)
```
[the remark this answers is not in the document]
roveo: I don't know how we can both allow mixing stream-specifiec and func arguments in `**kwargs` _and_ avoid explicitly listing them.

streamz/__init__.py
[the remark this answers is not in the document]
roveo: There's another problem with this: docs are written on the assumption that everything is importable directly from `streamz`, so I'm leaving this as is for now.

streamz/sinks.py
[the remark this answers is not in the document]
roveo: Please see here: https://github.com/python-streamz/streamz/pull/386/commits/5c1b3a6bdc556293157f542c9fad821d27545e3b#diff-eb662403bf433f3884021435c3f6cd010b387b4174a01367ae415b3e593cde5eR56-R60

Still looks ugly to me, but it does what you asked for :)

streamz/tests/test_sinks.py
roveo: Removed buffering and used `.flush()` in tests

streamz/sinks.py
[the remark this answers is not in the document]
martindurant: What do you think, @CJ-Wright , is this better or worse?

streamz/__init__.py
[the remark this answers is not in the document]
martindurant: Fair. We may come back to this.

streamz/sinks.py
martindurant: I thin it certainly just adds overhead here - I think the file write blocks the GIL while sending data to the OS buffer anyway. There is [aiofiles](https://github.com/Tinche/aiofiles), but I'd rather not complicate things.

streamz/tests/test_sinks.py
martindurant: An explicit destroy on a sink should remove it from the global sinks I think, yes. That can then be what gets called here in the test.

streamz/sinks.py
roveo: `aiofiles` still uses a threadpool behind the scenes, so there's no point in using it anyway. OK, will simplify and just use plain write.

streamz/tests/test_sinks.py
roveo: > I should note that the existing destroy should use the _remove_upstream methods since that handles things properly.

I tried doing this and `test_destroy` in core fails, because `_remove_upstream` replaces an upstream with `[None]` if there's only one upstream:

https://github.com/python-streamz/streamz/blob/a6e91114d6d3a9e2965856ffaaf3e56b6e9d3dfa/streamz/core.py#L257-L258

Why is a single upstream a special case? I'd expect just `self.upstreams.remove(upstream)` here.

CJ-Wright: I think it was so you had the option to add upstreams in case you needed to, see the associate addition method above

roveo: But the condition in `_add_upstream` will never be satisfied:

https://github.com/python-streamz/streamz/blob/a6e91114d6d3a9e2965856ffaaf3e56b6e9d3dfa/streamz/core.py#L238-L244

if you called `_remove_upstream` before, `self.upstreams` is `[[None]]`, not `[None]`. And why not just use `append` for all cases?

CJ-Wright: Also `[None]` is the state of a stream with no upstream or upstreams. https://github.com/python-streamz/streamz/blob/a6e91114d6d3a9e2965856ffaaf3e56b6e9d3dfa/streamz/core.py#L166

CJ-Wright: The `[[None]]` may be a mistake.

roveo: > Also [None] is the state of a stream with no upstream or upstreams.

Why not just an empty list? It seems there's a lot of code just to support this `else` branch in `Stream.__init__`:

https://github.com/python-streamz/streamz/blob/a6e91114d6d3a9e2965856ffaaf3e56b6e9d3dfa/streamz/core.py#L168-L171

> The [[None]] may be a mistake.

I'll try to make in `[None]` and rerun tests.

roveo:
I took the liberty of making the default `Stream.downstreams` an empty list instead of `[None]` and then just using append/remove to manipulate, core tests are passing.

martindurant:
The line that's annoying codecov is [this](https://github.com/python-streamz/streamz/blob/master/streamz/graph.py#L73), which should now be removed, if we don't have `None`s.

roveo:
I also have a question. What is the point of returning `[]` here?

https://github.com/roveo/streamz/blob/269d3ab397103436c1599b013a67aa83324e28c0/streamz/sinks.py#L63-L68

I tried commenting it out and running tests, everything seems to work the same.

CJ-Wright:
Hmm I'm not certain. Although at this point I'd trust the tests, if we break something then we should have a test for that behavior

martindurant:
> What is the point of returning [] here?

Looks like we're keeping it for now :) Feel free to make a new PR with the change.
This is being merged now!

roveo:
Great, thanks!

martindurant:
Just noticed `sink_to_file` in `streamz.sources` - obviously in the wrong place, and sorry to say you had to repeat some of this. The former seems to be called just once in test_core, and could be deleted in favour of yours.

roveo:
> Just noticed sink_to_file in streamz.sources

Yeah, I mentioned it in PR description. Wasn't confident I should be deleting it, especially because it's not really a source/stream node, but more of a utility function. If you're sure it won't be a breaking change, I'll remove.

martindurant:
Sorry I missed that.
Maybe just move it to sinks for now?

On November 27, 2020 6:40:30 PM EST, Mikhail Akimov <notifications@github.com> wrote:
>> Just noticed sink_to_file in streamz.sources
>
>Yeah, I mentioned it in PR description. Wasn't confident I should be
>deleting it, especially because it's not really a source/stream node,
>but more of a utility function. If you're sure it won't be a breaking
>change, I'll remove.
>
>-- 
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/python-streamz/streamz/pull/386#issuecomment-735011951

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #387, opened 2020-11-13
CI stuff

CJ-Wright:


martindurant:
Thanks @CJ-Wright ! What is the consequence here, after merge there should be a new CI running? Or does someone need to register with a provider?

CJ-Wright:
I'm doing the registering

CJ-Wright:
cycle for ci?

CJ-Wright:
I think I have registered the azure provider. I'm not certain if it would show up on the PR so I'm going to restore the travis file and we can try this out. Worst case we can back out the commits. What do you think @martindurant

CJ-Wright:
we might need to skip python 3.6 in the recipe

chinmaychandak:
I'm curious: Are there any advantages in using Conda Forge's CI instead of GitHub Actions, @CJ-Wright? Only reason I'm asking is because I see Dask moving to GH Actions.

martindurant:
@CJ-Wright - do you want to continue with this?

You could essentially copy this for Intake, which should run easier (no docker, plenty of server processes).

martindurant:
Note: intake and fsspec have moved to gh-actions. 
@CJ-Wright , what's the status here?

martindurant:
Edit: fastparquet also on gh-actions. This is the only active repo holdout!

CJ-Wright:
I'm sorry that this is holding things up. I may not be able to work on it in the near future, please feel free to push, close, or do anything else that you see fit.

PR #390, opened 2020-11-17
Test for Stream() backpressure and fix

### LINKED ISSUE #378: Using connect on an empty Stream() breaks backpressure

roveo:
Resolves #378 

Turns out, coroutine isn't required to fix this, just adding `return` is enough (like in `.map`).

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -483,7 +483,7 @@ class Stream(object):
             sync(self.loop, _)
 
     def update(self, x, who=None, metadata=None):
-        self._emit(x, metadata=metadata)
+        return self._emit(x, metadata=metadata)
 
     def gather(self):
         """ This is a no-op for core streamz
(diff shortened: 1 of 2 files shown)


martindurant:
Great

PR #393, opened 2020-11-26
from_iterable source

roveo:
A source that emits items from any iterable.

Why this should be in core:
- useful for tests (don't need to write the for loop with emits)
- this is the most generic form of a source, in theory you can implement any source as an unbounded generator

Also previously I missed that the `.destroy()` method in sinks was defined on `sink`, not the base `Sink` class, so there's a fix for that too.

### DIFF
--- a/streamz/sinks.py
+++ b/streamz/sinks.py
@@ -17,6 +17,10 @@ class Sink(Stream):
         super().__init__(upstream, **kwargs)
         _global_sinks.add(self)
 
+    def destroy(self):
+        super().destroy()
+        _global_sinks.remove(self)
+
 
 @Stream.register_api()
 class sink(Sink):
@@ -66,10 +70,6 @@ class sink(Sink):
             return result
         else:
             return []
-
-    def destroy(self):
-        super().destroy()
-        _global_sinks.remove(self)
 
 
 @Stream.register_api()
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -731,3 +731,36 @@ def get_message_batch_cudf(kafka_params, topic, partition, keys, low, high, time
     finally:
         consumer.close()
     return gdf
+
+
+@Stream.register_api(staticmethod)
+class from_iterable(Source):
+    """ Emits items from an iterable.
+
+    Parameters
+    ----------
+    iterable: iterable
+        An iterable to emit messages from.
+
+    Examples
+    --------
+
+    >>> source = Stream.from_iterable(range(3))
+    >>> L = source.sink_to_list()
+    >>> source.start()
+    >>> L
+    [0, 1, 2]
+    """
+
+    def __init__(self, iterable, **kwargs):
+        super().__init__(ensure_io_loop=True, **kwargs)
+        self._iterable = iterable
+
+    def start(self):
+        self.stopped = False
+        self.loop.add_callback(self._run)
+
+    @gen.coroutine
+    def _run(self):
+        for x in self._iterable:
+            yield self._emit(x)
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -763,4 +763,7 @@ class from_iterable(Source):
     @gen.coroutine
     def _run(self):
         for x in self._iterable:
+            if self.stopped:
+                break
             yield self._emit(x)
+        self.stopped = True
--- a/streamz/tests/test_sinks.py
+++ b/streamz/tests/test_sinks.py
@@ -65,6 +65,9 @@ def test_sink_destroy():
     sink = Sink(source)
     ref = weakref.ref(sink)
     sink.destroy()
+
+    assert sink not in _global_sinks
+
     del sink
 
     assert ref() is None
--- a/streamz/tests/test_sinks.py
+++ b/streamz/tests/test_sinks.py
@@ -2,7 +2,7 @@ import weakref
 
 import pytest
 from streamz import Stream
-from streamz.sinks import _global_sinks
+from streamz.sinks import _global_sinks, Sink
 from streamz.utils_test import tmpfile
 
 
@@ -62,7 +62,7 @@ def test_sink_to_textfile_closes():
 
 def test_sink_destroy():
     source = Stream()
-    sink = source.sink(lambda x: None)
+    sink = Sink(source)
     ref = weakref.ref(sink)
     sink.destroy()
     del sink
--- a/streamz/tests/test_sources.py
+++ b/streamz/tests/test_sources.py
@@ -95,3 +95,10 @@ def test_process():
     s.start()
     yield await_for(lambda: out == [b'0\n', b'1\n', b'2\n', b'3\n'], timeout=5)
     s.stop()
+
+
+def test_from_iterable():
+    source = Source.from_iterable(range(3))
+    L = source.sink_to_list()
+    source.start()
+    wait_for(lambda: L == [0, 1, 2], 0.1)
--- a/streamz/tests/test_sources.py
+++ b/streamz/tests/test_sources.py
@@ -102,3 +102,28 @@ def test_from_iterable():
     L = source.sink_to_list()
     source.start()
     wait_for(lambda: L == [0, 1, 2], 0.1)
+
+
+def test_from_iterable_backpressure():
+    it = iter(range(5))
+    source = Source.from_iterable(it)
+    L = source.rate_limit(0.01).sink_to_list()
+    source.start()
+
+    wait_for(lambda: L == [0], 1)
+    assert next(it) == 2  # 1 is in blocked _emit
+
+
+def test_from_iterable_stop():
+    from _pytest.outcomes import Failed
+
+    source = Source.from_iterable(range(5))
+    L = source.rate_limit(0.01).sink_to_list()
+    source.start()
+
+    wait_for(lambda: L == [0], 1)
+    source.stop()
+
+    assert source.stopped
+    with pytest.raises(Failed):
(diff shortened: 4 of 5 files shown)


martindurant:
Yes, I think this is a useful addition. Should there be a time-based version, or is that superfluous, since one could rate-limit the iterator anyway?

roveo:
I think rate_limit does the job, so there's no point in adding it to the source.

martindurant:
So what happens if the user also installs streamz_example_plugin, which from_iterable do you get?

roveo:
They'll get a warning that there's name collision with core, and core will take precedence.

martindurant:
OK :)

streamz/sources.py
```diff
@@ -731,3 +731,36 @@ def get_message_batch_cudf(kafka_params, topic, partition, keys, low, high, time
...
+
+    >>> source = Stream.from_iterable(range(3))
+    >>> L = source.sink_to_list()
+    >>> source.start()
+    >>> L
+    [0, 1, 2]
+    """
+
+    def __init__(self, iterable, **kwargs):
+        super().__init__(ensure_io_loop=True, **kwargs)
```
martindurant, line 756: There is actually nothing async about this source - do we need to bother with a loop? If yes, then are there any sources or other nodes which *don't* need a loop?

streamz/sources.py
```diff
@@ -731,3 +731,36 @@ def get_message_batch_cudf(kafka_params, topic, partition, keys, low, high, time
...
+        self._iterable = iterable
+
+    def start(self):
+        self.stopped = False
+        self.loop.add_callback(self._run)
+
+    @gen.coroutine
+    def _run(self):
+        for x in self._iterable:
+            yield self._emit(x)
```
martindurant, line 766: I suppose `self.stopped = True` when we run out of items; in other sources, `self.stopped` is also used as a condition to quit the loop.

streamz/tests/test_sources.py
```diff
@@ -95,3 +95,10 @@ def test_process():
     s.start()
     yield await_for(lambda: out == [b'0\n', b'1\n', b'2\n', b'3\n'], timeout=5)
     s.stop()
+
+
+def test_from_iterable():
+    source = Source.from_iterable(range(3))
+    L = source.sink_to_list()
+    source.start()
+    wait_for(lambda: L == [0, 1, 2], 0.1)
```
martindurant, line 104: Is there a lag, or is this immediate?

streamz/sinks.py
```diff
@@ -17,6 +17,10 @@ def __init__(self, upstream, **kwargs):
         super().__init__(upstream, **kwargs)
         _global_sinks.add(self)
 
+    def destroy(self):
```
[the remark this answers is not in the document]
roveo: Yes, the test was incorrect too (well, correct for the incorrect implementation). Fill fix.

streamz/sources.py
roveo: > There is actually nothing async about this source - do we need to bother with a loop?

I think we do: the rest of the loop is still asynchronous, and we need to run `_run` as a callback on an IOLoop, so it should be ensured. This source is not asynchronous, but the rest of the pipeline is, so we need to await on whatever's next. Sources are always at the start of a pipeline, so they _are_ the place to make sure that a loop exists.

I tried to think how this could be implemented without an IOLoop and without `gen.coroutine` and I couldn't figure out how.

streamz/sources.py
roveo: Yes, I should add a breaking condition and also add `self.stopped = True` at the end in case somebody wants to check if iteration is done.

streamz/tests/test_sources.py
roveo: There's a lag. I tried doing just `assert L == [0, 1, 2]` and it fails: I think the callback is not run immediately at `.start()` and the interpreter runs the assertion first.

streamz/sources.py
roveo: Btw, there are also async generators, and I don't think they are supported in tornado. We might want to support them here, but then `_run` won't be in line with the rest of the codebase, because it would be a built-in `async def` coroutine.

martindurant: I see that all the rest of the sources are all also async, which I hadn't realised. In a separate PR, we should refactor all the `start()` methods, which seem to be mostly identical.

>  because it would be a built-in async def coroutine

as a separate issue, I don't know that tornado is warranted in this code anymore, everything could be done with asyncio.

streamz/tests/test_sources.py
martindurant: Right - it was not immediately obvious to me that there *should* be a lag, which is a consequence of emitting asynchronously. Would you say that this basically comes from wanting to respect backpressure? 

I would imagine, if this class were not async, that backpressure would happen in another node downstream that *is*, such as `buffer`, which would take responsibility for creating the loop, if it doesn't exist. Then, calls to `emit` on the source would block until all items are processed.

streamz/sources.py
```diff
@@ -731,3 +731,39 @@ def get_message_batch_cudf(kafka_params, topic, partition, keys, low, high, time
...
+
+    def start(self):
+        self.stopped = False
+        self.loop.add_callback(self._run)
+
+    @gen.coroutine
+    def _run(self):
+        for x in self._iterable:
+            if self.stopped:
+                break
```
martindurant, line 767: Could use a test by calling `.stop()` (defined in `Source`)

streamz/tests/test_sinks.py
```diff
@@ -62,7 +62,7 @@ def test_sink_to_textfile_closes():
 
 def test_sink_destroy():
     source = Stream()
-    sink = source.sink(lambda x: None)
+    sink = Sink(source)
     ref = weakref.ref(sink)
     sink.destroy()
     del sink
```
martindurant, line 68: `assert sink not in sinks._global_sinks`

streamz/sources.py
roveo: > as a separate issue, I don't know that tornado is warranted in this code anymore, everything could be done with asyncio.

If you're comfortable mixing both in the codebase, I can rewrite this. But there's a lot still going on with tornado ioloop in stream internals, I don't think it can be just asyncio code in nodes without changing how the whole thing works, `self.loop` is still tornado.

streamz/tests/test_sinks.py
roveo: OK, will add. Didn't add this because I figured we only cared about the end result: that the sink is garbage collected.

streamz/tests/test_sources.py
roveo: > Would you say that this basically comes from wanting to respect backpressure?

Yes, but also see below.

> I would imagine, if this class were not async, that backpressure would happen in another node downstream that is

There's a couple of things.

- `Source` interface should be uniform. You have to call `.start()`, the call doesn't block and exits immediately. All sources are running in a concurrent fashion.
- The source is what's "running" the pipeline, so there has to be some code that runs some steps, emits and yields control so that other nodes can do their thing too. This is what async coroutines and plain generators can do, but in case of generators, something has to run the generator. That's what the event loop is for, so we come back to `gen.coroutine`s.
- I'm not sure how we could achieve this without it being async. We could run `loop.run_in_executor` in `.start()`, or run it without the loop in a thread, but it's really the same thing, just uglier (and breaks backpressure, I checked). If `.start()` blocks, this breaks the uniformity and is even more unexpected.

So I don't know what it _means_ for a source to not be async. If you have some specific solution in mind, please point me in the right direction.

roveo: Added a test for backpressure, in case you (or someone) wants to experiment with this not being async.

martindurant:
I'm convinced..

@CJ-Wright , what do you think of *requiring* sources to be async, so we can consolidate behaviour and code? Not in this PR, of course.

CJ-Wright:
I think that is reasonable. We might need to have some docs around how to write the async source, or have some nice tooling so that you only need to write the `poll` method or something (something that handles simple cases where the thing providing the data doesn't ship its own loop). We also will need to handle 3rd party async loops since we may not be able to assume that everything is tornado friendly.

martindurant:
Thanks, @CJ-Wright . Let's fold that into the whole "we need more documentation/example", and "could use refactoring" topics :). 

I wonder if you think that Tornado is justified at all? Probably the only things that directly depend on it are the TCP/HTTP sources (the former of which asyncio could do directly)  - but it would change quite a lot of things that use the properties yield, not to mention downstream code that depends on the current implementation. As things stand, async/await is possible but not required and not typical in the codebase.

In any case, I am merging this PR, thanks to @roveo for sticking with it!
I would suggest that an immediate refactor of Source to provide the loop, start=True init kwarg, stop() method and the main callback in `run` would be a very good change to simplify our code.

CJ-Wright:
Hmm, my understanding was the tornado integration was mostly focused on our dask integration. If that is not the case then we could consider moving away from it, or providing some way to tell us what kind of loop it runs on so we can do the translation to tornado loop under the hood.

martindurant:
> tornado integration was mostly focused on our dask integration

That would be the origin, but the dask client returns awaitable futures, so it ought to work. However, the dask client does have its own internal tornado loop (because it's actually a TCP app), so it may be that *if* you are using dask you must use that loop
```
(DaskStream.__init__)
        if 'loop' not in kwargs:
            kwargs['loop'] = default_client().loop
```
but tornado loops all contain asyncio loops maybe others?

PR #395, opened 2020-12-03
refactor and make less tornado

martindurant:
Following #393 , this moves some common code into Source, and takes the chance to make some of the sources async/await rather than tornado (the `loop` is still tornado running asyncio for now...). Kafka sources not touched, since they are more complex.

### DIFF
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -3,6 +3,7 @@ from glob import glob
 import os
 import time
 from tornado import gen
+import weakref
 
 from .core import Stream, convert_interval, RefCounter
 
@@ -446,6 +447,7 @@ class from_kafka(Source):
             self.stopped = False
             self.consumer = ck.Consumer(self.cpars)
             self.consumer.subscribe(self.topics)
+            weakref.finalize(self, self.consumer.close)
             tp = ck.TopicPartition(self.topics[0], 0, 0)
 
             # blocks for consumer thread to come up
@@ -496,7 +498,7 @@ class FromKafkaBatched(Source):
 
         @gen.coroutine
         def checkpoint_emit(_part):
-            ref = RefCounter(cb=lambda: commit(_part))
+            ref = RefCounter(cb=lambda: commit(_part), loop=self.loop)
             yield self._emit(_part, metadata=[{'ref': ref}])
 
         if self.npartitions is None:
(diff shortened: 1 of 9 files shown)


martindurant:
Should it always be `await asyncio.gather(*self._emit)`, and if yes, should this be folded into emit itself? _emit returns a list, and that list will be all coroutines if the pipeline has a loop - which all Sources now enforce.

martindurant:
cc @roveo

streamz/sources.py
```diff
@@ -35,15 +36,28 @@ def write(text):
...
 
+    def start(self):
+        """start polling"""
+        self.stopped = False
+        self.started = True
+        self.loop.add_callback(self.run)
+
+    async def run(self):
+        while not self.stopped:
+            await self._run()
```
roveo, line 59: Should there be a `_run` method that raises `NotImplemented`?

roveo:
Nice! This won't even break any existing custom sources, because most of them implement their own `.start()` method.

> Should it always be await `asyncio.gather(*self._emit)`

This looks like an idiom that should be hidden from whoever is using this API. Maybe `Source` could implement custom `.emit()` that's always async (and doesn't require `asynchronous=True`).

Something like

```py
async def emit(self, x, metadata=None):
    await asyncio.gather(*self._emit(x, metadata=metadata))
```

and then one will just

`await self.emit(x)` in `_run` implementation.

But then other node types will still have to use `asynchronous=True`, which is not uniform.  It would be nice to have one emit for synchronous/tornado coro emitting and one for async.

streamz/sources.py
martindurant: That would make sense

streamz/sources.py
```diff
@@ -17,6 +17,7 @@ def _():
 
     pc = tornado.ioloop.PeriodicCallback(_, callback_time, **kwargs)
     pc.start()
+    pc.start()
```
roveo, line 20: Why is this called the second time?

Also, as far as I can see, PeriodicCallback is only used in a single test (testing PeriodicCallback I think). Might as well remove it altogether.

martindurant: You would replace it with from_iter/buffer if you wanted this kind of source?

roveo: Ah, it's a source. I thought it was some utility for something that was removed. I didn't even recognize it as such without `@Stream.register_api(staticmethod)`. Then I think that it's not necessarily redundant, but I'd expect to see it mounted as `Stream` method, like other sources, e.g. `Stream.from_periodic_callback`.

martindurant: I think it might be useful, as the simplest of sources along with from_iter. Its style could be made to fit the rest of the module:

```python
@Stream.register_api(staticmethod)
class from_periodic(Source):
    
    def __init__(self, callback, poll_interval=0.1, **kwargs)
        self.cb = callback
        self.poll = poll_interval
        super().__init__(**kwargs)

    async def _run(self):
        await asyncio.gather(self._emit(self.cb()))
        await asyncio.sleep(self.poll)
```

It is sink_to_file that I would remove, since we now have to_textfile in sinks.

roveo: I agree, looks good.

As a side note, I think we need some clobbering protection for nodes (ref. #318). For example, if at some point we're going to have a node called `poll` in core, this will break. I'm personally using `_` at the beginning for all stream attrs, but it's really easy to forget/miss.

martindurant:
@chinmaychandak , please see the small kafka changes here, required to get things passing again.

chinmaychandak:
LGTM!

PR #400, opened 2020-12-04
Test showing RefCounter is broken on master

wwoods:
Since I put a regression test together to double check this was the issue, I'm submitting this as a not-ready-to-be-merged PR rather than an issue.

Either way, this code is fine on 0.6.1, but not master. Being able to use `RefCounter` manually to check when a certain piece of work is finished in the graph is important for my use case.

Edit: Disregard for the moment; there is an issue in master but not on 0.6.1, but at the moment this particular test is the schism between asynchronous and synchronous streamz.

### DIFF
--- a/streamz/tests/test_dask.py
+++ b/streamz/tests/test_dask.py
@@ -7,7 +7,7 @@ pytest.importorskip('dask.distributed')
 from tornado import gen
 
 from streamz.dask import scatter
-from streamz import Stream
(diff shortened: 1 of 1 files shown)


martindurant:
Is anyone else getting the feeling that all Streams should be async, in the current thread or separate event loop thread, and never plain python blocking calls?

wwoods:
Closed in favor of #401.

PR #401, opened 2020-12-04
Test showing synchronous partition(timeout) + dask scatter is broken

wwoods:
The output for this failing test shows, `ValueError: Two different event loops active`. Basically, the new `partition(timeout=...)` functionality doesn't play well with synchronous dask streams.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,3 +39,8 @@ max-line-length = 120
 
 [bdist_wheel]
 universal=1
+
+[tool:pytest]
+markers:
+  network: Test requires an internet connection
+  slow: Skipped unless --runslow passed
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -21,6 +21,11 @@ try:
 except ImportError:
     PollIOLoop = None  # dropped in tornado 6.0
 
+try:
+    from distributed.client import default_client as _dask_default_client
+except ImportError:
+    _dask_default_client = None
+
 from collections.abc import Iterable
 
 from .compatibility import get_thread_identity
@@ -41,6 +46,15 @@ _io_loops = []
 def get_io_loop(asynchronous=None):
     if asynchronous:
         return IOLoop.current()
+
+    if _dask_default_client is not None:
+        try:
+            client = _dask_default_client()
+        except ValueError:
+            # No dask client found; continue
+            pass
+        else:
+            return client.loop
 
     if not _io_loops:
         loop = IOLoop()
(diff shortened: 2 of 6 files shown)


martindurant:
I think you either have to create the dask client first and pass its `.loop` to the source
```
source = Stream(asynchronous=True, loop=c.loop)
```

DaskStream checks for a client and uses its loop, but I think it happens too late here, the source will already have created a loop. 
```
    def __init__(self, *args, **kwargs):
        if 'loop' not in kwargs:
            kwargs['loop'] = default_client().loop
        super(DaskStream, self).__init__(*args, **kwargs)
```
Perhaps the test for default_client() should be done in core.get_io_loop?

wwoods:
I see. Avoiding manual loop passing wherever possible would certainly simplify usage, and would better follow the Dask convention for client.

If the `distributed.client` import were wrapped s.t. it tolerates an `ImportError`, and that's OK to include in `core.py`, then trying `default_client` first in `core.get_io_loop` might streamline the user experience.

To this end, I've pushed a commit which I think addresses the issue, but clearly something isn't quite right since the tests don't pass. Anyway, putting this down for now.

wwoods:
Replying to @martindurant 's comment on #400 , since this is the open PR with a better description of the issue.

Perhaps the biggest oddity about removing synchronous calls in Streamz would be that Streamz itself doesn't really support asynchronous operations. The async computation looks like it's only used to prop up other functionality. For example, if I wanted an asynchronous computation in `map`, it doesn't look like Streamz supports this at the moment.

martindurant:
There are a few places where we check a function to see if it's awaitable (i.e., a coroutine object, after being called), and indeed it would make sense to use this sort of logic wherever a callable is required by the API. Again, we should probably have a firm idea of exactly what we support.

martindurant:
So, I think we do still need to force the dask stream to fetch its IO loop - otherwise it doesn't run at all. I wonder what would happen if it was instantiated *before* the dask client, would we get two even loops?

martindurant:
If you don't mind, I'll still push some small changes here to squash most of the warnings we see

wwoods:
Yeah push whatever you want here -- I actually didn't realize GitHub allowed commits on PRs outside of the repository providing the base PR branch.

The dask client is running its IO loop in a separate thread, it just breaks in the tests for some reason. If you try the code I submitted with a custom script, it works fine. E.g., this does what you'd expect:

```python
            s = streamz.Stream()
            L = s.partition(2, timeout=1.).scatter().map(lambda x: [xx+1 for xx in x]).buffer(8).gather().flatten().sink_to_list()
            s.emit(1)
            s.emit(2)
            s.emit(3)
            time.sleep(3)
            print(L)
```

wwoods:
Awesome, thanks for finishing this up! Much appreciated.

PR #404, opened 2021-01-08
Pieces that occurred while working on intake-streamz

martindurant:
https://github.com/intake/intake-streamz coming soon

Changes:
- a `current_value` attribute that holds the most recently emitted thing from a stream (and is *not* part of reference counting, but very useful for introspection)
- make API registering a mixin and apply to `DataFrame`, with only one method right now: `from_periodic`
- add `start` and `stop` methods to `DataFrame`, and make these actions work wherever in a pipeline they are called

### DIFF
--- a/streamz/collection.py
+++ b/streamz/collection.py
@@ -232,6 +232,12 @@ class Streaming(OperatorMixin, core.APIRegisterMixin):
     def current_value(self):
         return self.stream.current_value
 
+    def start(self):
+        self.stream.start()
+
+    def stop(self):
+        self.stream.stop()
+
     def _ipython_display_(self, **kwargs):
         return self.stream.latest().rate_limit(
             0.5).gather()._ipython_display_(**kwargs)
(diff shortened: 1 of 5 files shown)


martindurant:
Note that is we have `@DataFrame.register_api`, then we can also have this type of entry point

martindurant:
@roveo , what do you think of from_* methods on dataframes, and an entrypoints category for them? At the moment there is only one, `from_periodic` (plus maybe a counterpart to `Random`, which is just a special case).

martindurant:
@chinmaychandak @CJ-Wright , I thought you might find the changes here generally interesting.

chinmaychandak:
Excellent! Thanks for adding this :)

roveo:
> what do you think of from_* methods on dataframes, and an entrypoints category for them?

I only work with core day-to-day, so what's currently supported by plugins is in no way exhaustive. If there's a need for DataFrame sources in plugins (and nodes, and possibly sinks?), then sure, let's support them.

PR #411, opened 2021-03-19
Display in notebook without ipywidgets

### LINKED ISSUE #410: Informative repr for streamz when ipywidgets is not installed
Currently if the user does not have `ipywidgets` installed then running something like the following in a notebook produces no output

```python
import streamz.dataframe
df = streamz.dataframe.Random()
df.tail()
```

It might be nicer to include a message encouraging users to install ipywidgets in this case.

martindurant:
Fixes #410

### DIFF
--- a/streamz/collection.py
+++ b/streamz/collection.py
@@ -239,8 +239,17 @@ class Streaming(OperatorMixin, core.APIRegisterMixin):
         self.stream.stop()
 
     def _ipython_display_(self, **kwargs):
-        return self.stream.latest().rate_limit(
-            0.5).gather()._ipython_display_(**kwargs)
+        try:
+            from ipywidgets import Output
+            return self.stream.latest().rate_limit(
+                0.5).gather()._ipython_display_(**kwargs)
+        except ImportError:
+            # since this function is only called by jupyter, this import must succeed
+            from IPython.display import display, HTML
+            if hasattr(self, '_repr_html_'):
+                return display(HTML(self._repr_html_()))
+            else:
+                return display(self.__repr__())
 
     def emit(self, x):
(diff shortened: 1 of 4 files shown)


martindurant:
@chinmaychandak - are you aware of anything changing with kafka ? We seem to have a consistent failure here.

chinmaychandak:
This seems to be some kind of timing issue: https://github.com/python-streamz/streamz/pull/411/checks?check_run_id=2151820531#step:6:1617.  

Somehow streamz ended up reading multiple messages in a batch as opposed to just one; this should never have happened because there's a `sleep(2)` in place after flushing every new message to Kafka which is more than the poll interval b/w batches. We could add `max_batch_size=1`, or maybe try increasing sleep time.

I'm wondering why it worked with Py3.7, should we just rerun? Hesitant to add `flaky` here.

martindurant:
I did rerun a couple of times with the same result. I can't think why the two versions would be different either.

martindurant:
> this should never have happened because there's a sleep(2) in place after flushing every new message to Kafka 

Could it be that kafka (the broker or the driver, who knows) is just that slow in this CI setup?

chinmaychandak:
> > this should never have happened because there's a sleep(2) in place after flushing every new message to Kafka
> 
> Could it be that kafka (the broker or the driver, who knows) is just that slow in this CI setup?

Definitely possible! In this case, we should just increase sleep time. Sorry, I'm in IST currently, so replied will be timed accordingly.

martindurant:
@chinmaychandak if you have the time, would like to see that test fixed. I don't love the sleep() option, but if it's the only way...

Issue #220, opened 2019-01-21
corr and ewm


fl4p:
Would be nice to have more Pandas rolling window functions in streamz.

I implemented rolling correlation and exponential weighted window. The EWM uses a cut-off threshold for finite window sizes.

Is there interest in a PR of either?

mrocklin:
In principle yes, PRs for that functionality sounds great.

As always there are other concerns about code complexity, cohesion with the
current codebase and such, but assuming that things fit in well those
features would be great to have.

jdye64:
Late to the game here but corr would be exceptionally awesome for me in my case!

pkuphenix:
> Would be nice to have more Pandas rolling window functions in streamz.
> 
> I implemented rolling correlation and exponential weighted window. The EWM uses a cut-off threshold for finite window sizes.
> 
> Is there interest in a PR of either?

Will you share the implementation?

PR #412, opened 2021-03-22
Add Expanding and EWM.mean

mroeschke:
xref #220

Added `expanding` and `ewm` ops. For `ewm`, I used pandas' ewm implementation.

### DIFF
--- a/streamz/dataframe/core.py
+++ b/streamz/dataframe/core.py
@@ -170,6 +170,12 @@ class Frame(BaseFrame):
         """
         return Window(self, n=n, value=value, with_state=with_state, start=start)
 
+    def expanding(self, with_state=False, start=None):
+        return Expanding(self, n=1, with_state=with_state, start=start)
+
+    def ewm(self, com=None, span=None, halflife=None, alpha=None, with_state=False, start=None):
+        return EWM(self, n=1, com=com, span=span, halflife=halflife, alpha=alpha, with_state=with_state, start=start)
+
     def _cumulative_aggregation(self, op):
         return self.accumulate_partitions(_cumulative_accumulator,
                                           returns_state=True,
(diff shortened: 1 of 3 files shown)


jreback (changes requested):
@mroeschke if u can add some tests esp to assert the NotImplemented

streamz/dataframe/core.py
```diff
@@ -622,6 +640,122 @@ def groupby(self, other):
...
+            if halflife <= 0:
+                raise ValueError("halflife must satisfy: halflife > 0")
+            decay = 1 - np.exp(np.log(0.5) / halflife)
+            com = 1 / decay - 1
+        elif alpha is not None:
+            if alpha <= 0 or alpha > 1:
+                raise ValueError("alpha must satisfy: 0 < alpha <= 1")
+            com = (1 - alpha) / alpha
+        else:
+            raise ValueError("Must pass one of com, span, halflife, or alpha")
```
jreback, line 718: are these all hit in tests?

jreback (changes requested):

ramicaza:
I started working on implementing EWMA today then halfway through I realized "hey I should probably check if there were any PRs in the last month since I last checked".  Then I found this haha. 

Anyways, I noticed there was no mention of "adjust" in your changes. My implementation currently computes a running adjusted EWM which is equivalent to the default pandas ewm: `pd.DataFrame.ewm(adjust=True)`.  I was wondering if your implementation is the adjusted or non-adjusted ewma.  I think it's pretty important to offer both of these in the streamz ewm.

streamz/dataframe/core.py
mroeschke: Added tests for these.

mroeschke:
Agreed, and it would be great if you could follow up with those features @ramicaza! I don't think I'll be able to tackle that if/when this PR is merged in.

mroeschke:
@martindurant ready for a look when you get the chance (the failing test seems unrelated `test_kafka_batch_checkpointing_async_nodes_1`)

martindurant:
Agreed, same as in https://github.com/python-streamz/streamz/pull/411

PR #423, opened 2021-06-15
Fix  `apply` import from `dask`

galipremsagar:
In a recent dask PR: https://github.com/dask/dask/pull/7801, `apply` was removed from `dask.compatibility`. This PR fixes the issue here and is backward compatible.


galipremsagar:
Could someone from the maintainers review this PR?

martindurant:
Thanks for fixing

PR #433, opened 2021-10-01
Generic exception handling

freol35241:
Work in progress: A (naive?) stab at generic exception handling within streamz.

This PR attaches an exception handler to each Stream object that acts as an optional start of a new pipeline branch.
```python
from streamz import Stream

def problem(item):
    raise ValueError

s = Stream()
mapper = s.map(problem)

# Default
s.emit(123) # <-- raises and InvalidDataError

# explicit silencing of exceptions
mapper.on_exception().silent = True
s.emit(123) # No exception 

# Do something with the exception and the data that caused it
def problem_solver(item):
    data_that_caused_the_problem, resulting_exception = item
    # Do something
mapper.on_exception().map(problem_solver)
s.emit(123) # The exception and the faulty data is pushed downstream in the on_exception branch
```

This PR requires some linting, a few tests fail (but nothing major it seems) and updated docs. Consider it a starting point for discussion, mainly in relation to #86. Would appreciate some feedback on this before I continue.

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -255,7 +255,7 @@ class Stream(APIRegisterMixin):
             self.upstreams = []
 
         # Lazily loaded exception handler to avoid recursion
-        self._on_exception =  None
+        self._on_exception = None
 
         self._set_asynchronous(asynchronous)
         self._set_loop(loop)
@@ -678,16 +678,21 @@ class Stream(APIRegisterMixin):
                 m['ref'].release(n)
 
     def on_exception(self):
-        """Returns the exception handler associated with this stream
+        """ Returns the exception handler associated with this stream. The exception handler is either lazily loaded
+        at this point or (if alredy loaded) just returned.
         """
         self._on_exception = self._on_exception or _on_exception()
         return self._on_exception
 
 
 class InvalidDataError(Exception):
-    pass
+    """Generic error that is raised when data passed into a node causes an exception
+    """
+
 
 class _on_exception(Stream):
+    """ Internal exception-handler for Stream-nodes.
+    """
 
     def __init__(self, *args, **kwargs):
         self.silent = False
@@ -695,12 +700,13 @@ class _on_exception(Stream):
 
     def update(self, x, who=None, metadata=None):
         cause, exc = x
-        
+
         if self.silent or len(self.downstreams) > 0:
             return self._emit(x, metadata=metadata)
         else:
             logger.exception(exc)
             raise InvalidDataError(cause) from exc
+
 
 @Stream.register_api()
 class map(Stream):
@@ -737,7 +743,7 @@ class map(Stream):
 
     def update(self, x, who=None, metadata=None):
         result = self.func(x, *self.args, **self.kwargs)
-        self._emit(result, metadata=metadata)
+        return self._emit(result, metadata=metadata)
 
 
 @Stream.register_api()
--- a/streamz/tests/test_core.py
+++ b/streamz/tests/test_core.py
@@ -16,6 +16,7 @@ from tornado.ioloop import IOLoop
 import streamz as sz
 
 from streamz import RefCounter
+from streamz.core import InvalidDataError
 from streamz.sources import sink_to_file
 from streamz.utils_test import (inc, double, gen_test, tmpfile, captured_logger,   # noqa: F401
         clean, await_for, metadata, wait_for)  # noqa: F401
@@ -933,7 +934,7 @@ def test_pluck():
     assert L == [2]
     a.emit([4, 5, 6, 7, 8, 9])
     assert L == [2, 5]
-    with pytest.raises(IndexError):
+    with pytest.raises(InvalidDataError):
         a.emit([1])
 
 
@@ -945,7 +946,7 @@ def test_pluck_list():
     assert L == [(1, 3)]
     a.emit([4, 5, 6, 7, 8, 9])
     assert L == [(1, 3), (4, 6)]
-    with pytest.raises(IndexError):
+    with pytest.raises(InvalidDataError):
         a.emit([1])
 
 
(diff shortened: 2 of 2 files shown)


martindurant:
I like this idea! I can't promise exactly when I can have a look, but soon. 
First a couple of questions:
- what happens if an exception is raised not by the next node, but something else downstream?
- does this respect async backpressure? (I would expect the exception handler to not normally be async, but it could be)

freol35241:
Hmm, good questions:
* Since there is one exception handler associated with each node, each exception handler only deals with exceptions caused by "its own" node, nothing further downstream. On the other hand, exceptions caused by "its own" node can be both from a user-supplied function (for example as input to `map` and internal errors in `streamz`. Looking to my own use case this is exactly what I want, all exceptions caused by a node can be handled/not handled easily by the end user without interrupting a running pipeline.
* To be honest, I am not knowledgeable enough about python async programming to give you an answer to this. On the other hand, the implementation is not really making any major changes to the dataflow (still just using `update` of a downstream node) so I wouldn't think it should cause any problems.

All existing tests now pass and the linting complaints have been taken care of.

Remaining things:
* add tests for the new functionality
* docs

But I wont continue further before you have had a chance to take a look.

martindurant:
I have had a chance to glance at the code, and it had no occurred to me that the exception handling needs to go into *all* stream nodes, and handled upstream of the actual exception, in order to make things possible. I can see why you can't put exception handling into the node that is actually generating the exception, but might I suggest a less general change.

You could make a node for this specific functionality, something like 
```python
class ExceptionHandler(Stream):
    def __init__(self, upstream, on_error, catch=Exception, **kwargs): ...

        try:
            self.emit(x, metadata=metadata)
        except self.catch as e:
            self.on_error.update((x, e), self, metadata=metadata)
```
(here, `on_error` is a streamz node, but it could be just a function maybe)
Unfortunately, the exception branch would not show up in the graph visualisation. Somewhere, probably in the checkpointing conversation, we already talked about the possibility of labelling off-track nodes (i.e., something not following upstream/downstream), but I don't think there was any progress on that.

Finally, `InvalidDataError` seems to me more like a testing entity, not something to put into the main core module.

cc @jsmaupin

freol35241:
Thanks for taking a look 👍 

Yes, my opinion is that a fully generic and integrated mechanism for exception handling probably needs to be present in all nodes. With that said, your suggestion is a less intrusive (but makes "building" the pipeline a bit more verbose and perhaps slightly less intuitive (maybe a highly personal preference?)) so if you feel my suggestion is over the top feel free to close this PR.

martindurant:
I think I would prefer the minimal change - would you like to code it up along the lines of my suggestion? I think your example should be easy enough to make.

roveo:
My thoughts:

- In the general case, handlers should only apply to the stuff that happens in a paricular node's logic. This is currently tough to do, because calling the downstreams is done inside `_emit`, which is normally called inside `.update`, so there's no separation between result computation and passing the result downstream/calling the downstreams. There are multiple solutions to this (like having ._emit return a callable), but none I can think of are backwards-compatible. To achieve this, `update` should terminate before downstreams are called.

- I think API-wise, which error type is handled should be more explicit, e.g.

```
dicts = src.map(json.loads)
dicts.on_exception(ParsingError).sink_to_file(...)
```

instead of checking for error type in the handler function and re-raising if it's not `ParsingError`.

- Handlers should be part of the stream graph and be a 1st-class citizen. Let's say I'm getting JSON strings from a source and want to do something if there's an error parsing it:

```py
dicts = src.map(json.loads)
invalid_batches = dicts.on_exception(ParsingError).partition(100, timeout=30)
invalid_batches.sink_to_s3(...)
invalid_batches.map(make_slack_message).sink_to_slack(...)
```

So I want to wait 30 seconds if there are any more bad records, write them to S3 for manual inspection and notify myself. This looks like a normal stream, just a different branch.

- There's retrying always somewhere close to exception handling. We should think about how this would play with a 3rd-party library like `tenacity` or maybe integrate it into streamz. Ideally, I want to do something like:

```py
sink = result.sink_to_db(...)
sink.on_exception(InvalidDataError).sink_to_file("rejecteddata.log")
sink.on_exception(ConnectionError).retry(wait_seconds=10, max_retries=5)
```

- An idea that I'm not sure is good/in scope, but still want to share. Is there a use case for wrapping whole sections of a pipeline in a handler?

```py
node.try().something().something_else().except(Exception).map(handle)
```

freol35241:
Stumbled upon this again after not having used streamz for a while. For what its worth, I currently use the following for exception handling, it gives me the flexibility I need but I wouldn't be surprised if it has some nasty side effects that I have yet to discover...

```python
from types import MethodType
from streamz import Stream, Sink

@Stream.register_api()
class on_exception(Sink):
    def __init__(self, upstream: Stream, exception=Exception, **kwargs):

        original_upstream_update_method = upstream.update

        def _(upstream_self, x, who=None, metadata=None):
            try:
                return original_upstream_update_method(x, who, metadata)
            except exception as exc:
                # Pass down the branch started with this stream instead
                self._emit((x, exc), metadata)

        # Bind to upstream
        upstream.update = MethodType(_, upstream)

        pass  # NO-OP
```

martindurant:
Hah, monkey-patch :)

I still think there is merit in making this kind of operation available for everyone. I'm afraid I still like my solution, which can be fleshed out a bit:
```python
class ExceptionHandler(Stream):
    def __init__(self, upstream, on_error: Stream,
                 catch: Tuple[Exception] = (Exception,), retries=0, retry_wait=1, **kwargs):
        self.on_error = on_error
        self.catch = catch
        self.retries = retries
        self.retry_wait = retry_wait

        retries = self.retries
        while True:
            try:
                return super().emit(x, metadata=metadata)
            except self.catch as e:
                if retries > 1:
                    time.sleep(self.retry_wait)  # or async?
                    retries -= 1
                    continue
                return self.on_error.update((x, e), self, metadata=metadata)
 ```

Issue #432, opened 2021-09-22
Integrate with River


martindurant:
https://riverml.xyz/latest/ provides online, per-event machine learning. Whilst we have tried to approach them to accept streamz as a known type of input, it makes sense to just do it here.

The idea would be to have RiverLearn streamz node (either a sink, or a normal node returning a metric) and a RiverPredict node, the pair sharing the same River model or pipeline. The River calls are blocking, but fast - so this can be part of a fast async events system. There are other specific types in River that might emit on a condition, e.g., the drift detector's `.change_detected`, so we might have a few classes to make, but all of them pretty simple.

PR #434, opened 2021-10-04
Add River nodes and examples

martindurant:


martindurant:
Fixes #432 (eventually)

@philippjfr - the notebook version of this works ok, but is limited by the jupyter comm to about 20Hz. Running the script version fails quickly with:
```
RuntimeError: _pending_writes should be non-None when we have a document lock, and we should have the lock when the document changes
```

It would be cool to figure a viz pipeline that only updates every so often, perhaps with datashading. 

NB:  but the cluster centres are not actually reasonable. When the pipeline and viz works OK, I'd like to get in touch with someone at River to explain the correct way to prime the model.

martindurant:
https://user-images.githubusercontent.com/6042212/136070852-8128d479-d696-4a12-b93d-6c93791fbd96.mp4

PR #437, opened 2021-10-08
websocket integration

### LINKED ISSUE #420: MQTT as a first-class Source?
MQTT is a widely used means for data transport especially in the sensors/IoT domain.  It would be incredibly helpful if it could be added as a Source along with all the others that currently appear in streamz.

I have tried adding this patch manually to `sources.py` and it seems to work. I appreciate that more is required to make it a really useful function, but this "proof of concept" seems usable.

Would the developers be supportive of adding this type of function, or is it considered "out of scope" of the intended use of the library?

**PATCH**

In the header section of `sources.py` add:
```
import paho.mqtt.client as mqtt
```
(obviously `paho-mqtt` needs to be installed as a dependency.)

At the end add:
```
@Stream.register_api(staticmethod)
class from_mqtt(Source):
    """
    Creates events by subscribing to an MQTT data stream

    Parameters
    ----------
    host : str
        The host to listen on
    topic : str or list
        The topic, or list of topics, to listen on
    port : int
        The port to listen on (default 1883)
    start : bool
        Whether to immediately initiate the source. You probably want to
        set up downstream nodes first.
    server_kwargs : dict or None
        If given, set of further parameters to pass on to the MQTT broker

    Examples
    --------
    >>> source = Source.from_mqtt('broker.hivemq.com', 'sensor/temperature')  # doctest: +SKIP

    """

    def __init__(self, host, topic, port=None, server_kwargs=None, **kwargs):
        self.server_kwargs = server_kwargs or {}
        self.host = host
        self.topic = topic
        self.port = int(port) if port else 1883
        self.client = None
        super().__init__(**kwargs)

    def run(self):

        def on_connect(client, userdata, flags, rc):
            client.subscribe(self.topic)

        def on_message(client, userdata, msg):
            self._emit(msg.payload.decode())

        self.client = mqtt.Client()
        self.client.on_connect = on_connect
        self.client.on_message = on_message
        self.client.connect(
            self.host,
            self.port,
            self.server_kwargs.get('keep_alive', 60)
        )
        self.client.loop_forever()

    def stop(self):
        if self.client and not self.stopped:
            self.client.loop_stop()
            self.client.disconnect()
            self.stopped = True
```

**TEST**

1. First start publishing to the publicly accessible MQTT broker:
```
#!/usr/bin/env python3
import time
import random
import paho.mqtt.client as mqtt

client = mqtt.Client()
client.connect("broker.hivemq.com", 1883, 60)

while True:
    temp = random.random() * 20.0 + 5.0
    print(temp)
    client.publish("streamz/sensor/temperature", temp)  # Celcius
    time.sleep(5)

client.disconnect()
```
2. Then connect up a stream to process the data:
```
import time
from streamz import Stream

source = Stream.from_mqtt('broker.hivemq.com', 'streamz/sensor/temperature')
source.sink(print)
source.start()

while True:
    time.sleep(0.1)
```

### LINKED ISSUE #422: Sink to a Web Socket server?
I am trying to implement a web socket sink.

I an running https://hub.docker.com/r/jmalloc/echo-server as a simple server (the "host" and "port" in the code below) to test delivery of messages.

Below is the simple (and probably naive) code I have used as a test:

```
import os
import time
from streamz import Stream
# pip install websocket-client
import websocket
from websocket import create_connection


def increment(x):
    return str(x * x)


def wss(x, url, websocket):
    while not websocket.connected:
        print('reconnecting')
        websocket.connect(url)
    try:
        print('SND:', x)
        r = web_socket.send(x)
        print('RCV', r)
    except WebSocketTimeoutException as err:
        print(err)
    except Exception as err:
        print(err)
        

host = os.getenv("WEBSOCKET_HOST", "123.45.67.89")
port = os.getenv("WEBSOCKET_PORT", "8345")

web_socket = create_connection(
    "ws://{}:{}/".format(host, port),
    timeout=30)
source = Stream()
source.map(increment).sink(wss)

for j in range(1, 4):
    print('\nstarting loop:', j)
    for i in range(1, 10):
        source.emit(i * j)
    time.sleep(60)

web_socket.close()
```

This seems to work, but if I change this setup to use an asynchronous source instead, e.g. data from a message broker which arrives intermittently, then very soon (after a dozen messages) the WebSocketTimeoutException is raised with each subsequent message being sent.

Is there a documented and straightforward way to reliably sink messages into a web socket?

If not, are there guidelines to follow - especially as we'd like to extend the range of sinks; and it seems possible there is a more generic approach that will apply when connecting to similar dynamic end-points?

martindurant:
Fixes #422
Fixes #420

### DIFF
--- a/ci/environment-py38.yml
+++ b/ci/environment-py38.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - python=3.7
+  - python=3.8
   - pytest
   - flake8
   - black
@@ -32,3 +32,5 @@ dependencies:
   - flaky
   - pytest-cov
   - coveralls
+  - paho-mqtt
+  - websockets
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -47,6 +47,7 @@ Stream
 .. automethod:: Stream.sink_to_list
 .. automethod:: Stream.sink_to_textfile
 .. automethod:: Stream.to_websocket
+.. automethod:: Stream.to_mqtt
 .. automethod:: Stream.update
 .. automethod:: Stream.visualize
 
@@ -58,6 +59,7 @@ Sources
    filenames
    from_kafka
    from_kafka_batched
+   from_mqtt
    from_process
    from_websocket
    from_textfile
--- a/streamz/sinks.py
+++ b/streamz/sinks.py
@@ -134,17 +134,57 @@ class to_websocket(Sink):
         self.uri = uri
         self.ws_kw = ws_kwargs
         self.ws = None
-        super().__init__(upstream, **kwargs)
+        super().__init__(upstream, ensure_io_loop=True, **kwargs)
 
-    @gen.coroutine
-    def update(self, x, who=None, metadata=None):
+    async def update(self, x, who=None, metadata=None):
         import websockets
         if self.ws is None:
-            self.ws = yield websockets.connect(self.uri, **(self.ws_kw or {}))
-        yield self.ws.send(x)
+            self.ws = await websockets.connect(self.uri, **(self.ws_kw or {}))
+        await self.ws.send(x)
 
     def destroy(self):
         super().destroy()
         if self.ws is not None:
             sync(self.loop, self.ws.protocol.close)
         self.ws = None
+
+
+@Stream.register_api()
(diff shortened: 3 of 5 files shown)


martindurant:
(needs docs)

PR #438, opened 2021-10-30
Fixes issue 431 for kafka oauth source

prodo56:
Confluent kafka fails to connect to kafka broker with oauth authentication when consumer.poll() is not invoked before fetching the watermark offsets for the topic. Added support to connect to kafka brokers with oauth authentication by invoking consumer.poll()

### DIFF
--- a/streamz/sources.py
+++ b/streamz/sources.py
@@ -452,7 +452,8 @@ class from_kafka(Source):
             weakref.finalize(self, lambda consumer=self.consumer: _close_consumer(consumer))
             tp = ck.TopicPartition(self.topics[0], 0, 0)
 
-            # blocks for consumer thread to come up
+             # blocks for consumer thread to come up and invoke poll to establish connection with broker to fetch oauth token for kafka
+            self.consumer.poll()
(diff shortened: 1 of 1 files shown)


martindurant:
Calling `poll()` seems to be causing the tests to hang. Any thoughts?

pthalasta:
@martindurant i see there is a difference in confluent-kafka version from my local system compared to the one in the repo. Could you tell me how to execute the test suite on local or where can i find the guide doc for executing the tests? I can run some tests on my local and make necessary changes.

Confluent kafka that i have on my system is 1.5.0

martindurant:
To run locally, you need one of the environments in ci/ , docker, and environment variable STREAMZ_LAUNCH_KAFKA=true

prodo56:
Thanks! let me make some changes and test it locally.

prodo56:
@martindurant i've fixed the version conflicts, please review

martindurant:
OK, that does the trick - this is going in, thank you.
I didn't understand why it worked, though!

PR #446, opened 2021-12-22
RiverML: update the plots less frequently

maximlt:
Instead of using `hvplot` I've directly used `holoviews` and `DynamicMap`s and `Pipe`s. I've partitioned the observations and clusters and concatenated them to send them only every X events. Running the script with `viz=True` I get a refresh rate of about 50 Hz for a cadence of 100 Hz, while I get an output at about 70 Hz without any viz (same cadence).

### DIFF
--- a/examples/river_kmeans.py
+++ b/examples/river_kmeans.py
@@ -1,4 +1,6 @@
+import functools
 import random
+import time
 
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks very much, good to see! We should make this example more prominent, right?

maximlt:
Yes why not! It's an interesting and powerful application of streamz+holoviews. The question is, where? @jbednar any recommendation? This is what the example looks like (running the `river_kmeans.py` script with `viz=True`):

riverml_streamz_example

jbednar:
Maybe link from examples.pyviz?

maximlt:
Maybe this should just be a proper example on examples.pyviz.org

@martindurant would you be interested in writing a notebook that introduces streamz and its support of RiverML, and the example (online k-means clustering)? If so, I could take care of polishing the plot and publishing the example on examples.pyviz.org (creating the anaconda-project and deploying it on AE5).

PR #458, opened 2022-11-22
handle username and pw for mqtt source

phbj:
Addition to mqtt source for handling username and password inputs


martindurant:
Thanks!

PR #463, opened 2022-12-21
Fix #462 flattening of iterators without len

### LINKED ISSUE #462: flatten doesn't work with iterables without defined length
The documentation of `flatten` states, that it works with any iterable. A python iterable in general isn't required to have a predefined length. So a call of `len(iterable)` should be handled with the necessary care. Unfortunately, this doesn't happen in `flatten` and fails with a `TypeError`.

Reproducer:

```python
from streamz import Stream
stream = Stream()
lst = stream.flatten().sink_to_list()
gen = (r for r in range(10))
stream.emit(gen)
```

```
File ~/.local/lib/python3.10/site-packages/streamz/core.py:1632, in flatten.update(self, x, who, metadata)
   1630 L = []
   1631 for i, item in enumerate(x):
-> 1632     if i == len(x) - 1:
   1633         y = self._emit(item, metadata=metadata)
   1634     else:

TypeError: object of type 'generator' has no len()
```
The loop should probably just emit the previous item, until it runs in a `StopIteration` and then emit the last item together with the metadata. Or maybe even just sending the metadata together with the first item?

ml31415:
Fixes https://github.com/python-streamz/streamz/issues/462

### DIFF
--- a/streamz/core.py
+++ b/streamz/core.py
@@ -1,6 +1,7 @@
 import asyncio
 from collections import deque, defaultdict
 from datetime import timedelta
+from itertools import chain
 import functools
 import logging
 import six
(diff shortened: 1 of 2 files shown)


streamz/core.py
```diff
@@ -1637,15 +1638,20 @@ class flatten(Stream):
...
+        item = next(items)
+        for item_next in items:
+            y = self._emit(item)
+            item = item_next
             if type(y) is list:
                 L.extend(y)
             else:
                 L.append(y)
+        y = self._emit(item, metadata=metadata)
+        if type(y) is list:
```
martindurant, line 1651: `isinstance(y, list)` ? Would any iterable work here?

ml31415: I took that from the previous implementation and wondered about the same question. As I wasn't sure if there is some special treatment for lists hidden somewhere else in the codebase, I decided to keep it as is.

martindurant: OK, well if it was like that before...

martindurant:
Thanks for putting this together

PR #485, opened 2025-12-12
Flake8 fix

dwskoog:
* Fixed Flake8 error in streamz/plugins.py
* Fixed conftest.py to use current pytest_ignore_collect params
  * Pytest 7 added the new param [name/type](https://docs.pytest.org/en/latest/reference/reference.html#pytest.hookspec.pytest_ignore_collect) and it was removed in 9. Pytest 9 gets pulled for the 3.10+ test environments.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -10,6 +10,6 @@ def pytest_runtest_setup(item):
         pytest.skip("need --runslow option to run")
 
 
-def pytest_ignore_collect(path, config):
-    if 'run_test.py' in str(path):
+def pytest_ignore_collect(collection_path, config):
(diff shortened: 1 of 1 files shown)


dwskoog:
> I wonder, do you use streamz at all? I'd love to hear of some interesting use cases.

At my company, we use it at two different points in our data processing to create a disjoint subcover of a noisy stream of telemetry readings and then blend in metadata from users and external authoritative resources to give users an understanding of what (and what kinds of) events are happening in their space.

We, unfortunately, were not able to use it for a larger data processing project because the project looked dead and the 0.6.4 version conflicts with Python 3.12+ semantics.

PR #483, opened 2025-12-04
Add support for Python 3.14

amotl:
## About
A few adjustments to make the majority of the test cases succeed on Python 3.14 on my machine.

```
1031 passed, 438 skipped, 8 xfailed, 96 xpassed, 151 warnings in 51.54s
```

## References
- https://github.com/python-streamz/streamz/issues/482

PR #486, opened 2025-12-16
Add Python 3.13 to CI

dwskoog:
* add conda environment file for 3.13
* add py313 to test.strategy.matrix

I noticed that there are several DeprecationWarnings around event loop uses. I can include changes for resolving those as well. It seems like they might be real problems in the 3.14+ world as there are major changes around asyncio there:
* 3.14 turns off the creation an event loop in `get_event_loop` and strongly encourages just using `get_running_loop`
* 3.14 deprecates event loop policies entirely which are scheduled for removal in 3.16
* Tornado 6.5 IOLoop is basically just a wrapper for asyncio according to their docs and 6.2+ changes several behaviors of the IOLoop

Most of the changes were just in the test cases, but the API change in Tornado impacts streamz/core.py as that is eagerly trying to create an event loop in an incorrect place.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        CONDA_ENV: [py39, py310, py311, py312]
+        CONDA_ENV: [py39, py310, py311, py312, py313]
     # env:
       # STREAMZ_LAUNCH_KAFKA: true
 
--- a/ci/environment-py313.yml
+++ b/ci/environment-py313.yml
@@ -1,0 +1,27 @@
+name: test_env
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - python=3.13
+  - pytest
+  - flake8
+  - black
+  - isort
+  - tornado
+  - toolz
+  - librdkafka
+  - dask
+  - distributed
+  - pandas
+  - python-confluent-kafka
+  - codecov
+  - coverage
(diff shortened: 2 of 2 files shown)


amotl:
> 3.14 deprecates event loop policies entirely which are scheduled for removal in 3.16

This patch might already include relevant adjustments?

- GH-483

dwskoog:
> > 3.14 deprecates event loop policies entirely which are scheduled for removal in 3.16
> 
> This patch might already include relevant adjustments?
> 
> * [Add support for Python 3.14 #483](https://github.com/python-streamz/streamz/pull/483)

The child watcher issue that you ran into is a good example. Those were deprecated in 3.12 and replaced with OS-specific "optimal" implementations and the ability to control a watcher was deprecated in favor of baking that in to the loop factory itself. More than that however, the behavior of `get_event_loop` changed as well. In 3.12 they deprecated the "create new loop if none" behavior and in 3.14 they made it an error. I crawled through the 3.12 and 3.13 builds to find the deprecation warnings and reconciled the Python 3.12+ changes with Tornado 6.2+ changes. I have a couple commits on another branch that resolves those. [One of them](https://github.com/dwskoog/streamz/commit/5e9434b0052d3118b7e2f2fe0ae568ab1a62eb1e#diff-86e0552de77bcec94f13f1be8531f620d1378e2f73842e4776d49640adb88c2fR54-R55) looks like it is a real change that the main code base needs.

PR #487, opened 2025-12-17
Documentation: Add page about enumerating use cases

amotl:
@martindurant: At https://github.com/python-streamz/streamz/pull/485#issuecomment-3657451732, you asked:

> I'd love to hear of some interesting use cases.

I think it would be nice to also have use cases on the documentation, so I am trying to make a start hereby in the hope you will like it.

@dwskoog: I've added your response from https://github.com/python-streamz/streamz/pull/485#issuecomment-3662789573 1:1 to this patch. Let me know if you would like to see some adjustments before possibly being published, depending on Martin's verdict.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -123,3 +123,4 @@ data streaming systems like `Apache Flink <https://flink.apache.org/>`_,
    async.rst
    plotting.rst
    plugins.rst
+   use-cases.rst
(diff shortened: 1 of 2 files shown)


amotl:
I can have a look. Can you enable the pull request builder in the RTD console, so we can also have previews via CI, and at the same time see if relevant patches are valid on behalf of the PR itself?

martindurant:
https://streamz.readthedocs.io/en/latest/use-cases.html

PR #488, opened 2025-12-17
Add missing RTD configuration file, which is obligatory today

amotl:
## Problem
It looks like the RTD [configuration file](https://docs.readthedocs.com/platform/stable/config-file/v2.html) was missing. It was optional before, but is obligatory today.

## References
- https://github.com/python-streamz/streamz/pull/487#issuecomment-3667244693

PR #490, opened 2025-12-17
RTD: Install package. Synchronize dependency list.

amotl:
## Problem
The RTD build currently fails, because it needs the package itself to be installed, in order to build API docs.

## References
- https://github.com/python-streamz/streamz/pull/488#issuecomment-3667320110

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -1,9 +1,5 @@
 numpydoc
 sphinx
 sphinx_rtd_theme
-tornado
-toolz
-zict
 pandas
-dask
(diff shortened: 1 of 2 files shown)


docs/requirements-docs.txt
```diff
@@ -1,9 +1,5 @@
 numpydoc
 sphinx
 sphinx_rtd_theme
-tornado
-toolz
-zict
```
amotl: The removed packages are dependencies of `streamz`. If all works well, they will be installed by the package install now defined within the RTD configuration.

